PR #29, opened 2025-11-26
chore: Add Taskfile

robertgzr:
Refs: [TOOL-182](https://linear.app/unikraft/issue/TOOL-182/share-github-actions-workflow-across-go-based-projects)

---

- **fix: Use Msg for dynamic message**
- **chore(build): Use shared Taskfile**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 dist/
+.task/
 .goreleaser.yaml
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,0 +1,21 @@
+# https://taskfile.dev
+version: 3
+
+includes:
+  go:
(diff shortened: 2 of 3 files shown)


robertgzr:
for CI, the shared workflow doesn't handle multiple platforms, so that would have to be addressed before integrating it here.

robertgzr:
looks like the `check/pr` here is broken as well🥲 doesn't accept `chore`

nderjung (approved):
The `chore` topic was nominally added to the internal actions in [#6](https://github.com/unikraft-cloud/internal-actions/pull/6).  This PR's pipeline is failing on this (even after re-runs) and won't in the future so I will merge anyway. Thanks!

PR #30, opened 2025-12-01
test: add golden tests

jedevc:
As discussed at the offsite, this can be used as a way to test + visualize the output of various commands.

The general approach is:
- Capture the output of the specified commands
- Cleanup the output by removing dynamic data (timestamps, platform-specific info, etc)
- Format the output into reports, and write the data to the file

Running `go test ./cmd/unikraft` tests that the output hasn't changed (and provides a diff if they have), while `go test ./cmd/unikraft -update` updates the files.

This provides a couple benefits:
- Improved review flow: reviewers can check that the output/ui is as expected without needing to review *all* underlying logic.
- Git history: the history of any ui can be followed, and fits into standard git blame tooling.


cmd/unikraft/testdata/TestGolden/help
```diff
@@ -0,0 +1,22 @@
...
+	  -c, --config
+	          Set the configuration file. ($UNIKRAFT_CONFIG).
+	      --log-level
+	          Set the logging level. ($UNIKRAFT_LOG_LEVEL)
+	          [default: info, choice: trace,debug,info,warn,error,fatal]
+	      --log-type
+	          Set the log type. ($UNIKRAFT_LOG_TYPE)
+	          [default: text, choice: text,json]
+	  -v, --version
+	          Print version information..
```
jedevc: This is weird, fix incoming in https://github.com/unikraft-cloud/x/pull/66.

cmd/unikraft/main_test.go
```diff
@@ -0,0 +1,177 @@
...
+	ctx := t.Context()
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			require.NotEmpty(t, tc.commands, "no commands specified")
+
+			output := strings.Builder{}
+
+			for i, command := range tc.commands {
+				require.NotEmpty(t, command, "no command specified")
```
robertgzr, line 52: what do you think of using `gotest.tools/v3/assert` over `require`?

cmd/unikraft/main_test.go
```diff
@@ -0,0 +1,177 @@
...
+	repl    string
+}
+
+// cleaners are patterns applied to command output to normalize variable data
+// so we get consistent golden files.
+var cleaners = []cleaner{
+	{
+		// times like "12:34PM" change between runs
+		pattern: regexp.MustCompile(`\b\d\d?:\d\d[AP]M `),
+		repl:    "XX:XX ",
```
robertgzr: maybe make that `HH:MM` to make it (even more) obvious in the golden files that this is a timestamp

robertgzr (approved):

cmd/unikraft/main_test.go
jedevc: Have updated it everywhere, but left one `require` in one place where continuing would result in reading into an empty array.

cmd/unikraft/main_test.go
jedevc: Done :tada:

cmd/unikraft/testdata/TestGolden/help
```diff
@@ -0,0 +1,22 @@
+$ unikraft --help
+
+stdout:
+	Usage:
+	  unikraft [flags]
+
+	Version:
+	  dev (linux/amd64)
```
nderjung: What happens when I try this on my darwin/arm64?

jedevc: Ah indeed. I'll add it to the output cleanup regexes.

(the sad disadvantage of this entire approach)

nderjung: Would be nice to have fancy in-line things like:
```
Version
  dev ($$\w+\/\w+$$)
```

jedevc: Mm. But then you miss being able to autogenerate them right? It's quite nice to just be able to pass `-update` to get them all up to date.

nderjung: Oh that's nice.  That said, if the tool had this it would be nice to skip ones that contain these escape expressions because they are manually manipulated unless a `-force` flag was also provided.  But I'm just making things up now :) 

For the majority of cases, though, I imagine these are architecture and platform agnostic. I suspect only a few places where this might matter: here of course; the scenario where `unikraft build` with default arguments auto-selects the host platform's architecture (which would only occur on a local build AND local execute of unikraft core, not the remote which would always be x86_64); and, similarly where `unikraft run` is used.

It could be that the simple mitigation for this is to ensure that tests are executed against the linux/amd64 goos/goarch always. Since we have device emulation this should not be a problem and since it will usually be executed in a CI where this is also the case it should not be a problem either -- it just needs to be enforced so stupid macOS users like me can run `make test` and computer go green :) If we take this approach, no nasty regex is necessary and no manually fashioned screens need to be made, making this process much more maintainable. Wdyt?

jedevc: Yeah I think it'll be possible to keep this list pretty short *hopefully*.

IMO, it's not too much work to just add another regexp here and get it working - I can get some test scripts and CI setup as well (but doing the emulation dance is a bit trickier than having a couple more lines here). I think I can auto-craft a regexp for all the goos/goarch combos as well that might make it better - will take a look asap tomorrow.

There's other cases as well where we need the regex, like for timestamps in logs, etc.

Obviously at some point if the regexp list becomes far too large, we can do another approach (e.g. maybe we could easily have the cleanup regexes as a header to the golden files).

But also who knows how this will scale, or work at all, just wanna have something so there's some sort of a baseline to track.

nderjung: Ok fair enough. Cheers! I'll wait for your ping then on this PR. Thanks again!

jedevc: Updated to simplify the regex construction for this case and for go version detection :tada:

PR #33, opened 2025-12-11
feat: Add authentication

jedevc:
This is the start of a piece-by-piece import of the CLI handed off by @nderjung.

There's a *lot* of different threads to untangle, so I'm untangling them and splitting them into separate PRs as I understand them.

This PR extracts the `login` command as well as the profile logic. There are two main changes from @nderjung's version:
- A `logout` command, which removes the current profile.
- The `login` command will consume `UKC_TOKEN` if set (currently it's not validated, but this should be possible by asking the control plane).

Finally, a super basic version of the `instances` command is added so that the auth has something to be used on. This will be filled out.


nderjung (approved):
Thanks!

PR #39, opened 2025-12-29
feat: Add images command

jedevc:
Allow `unikraft images`.

This also includes a breaking change - to do metro-specific requets, instead of saying `<metro>/<name-or-uuid>`, you now have to say `<metro>#<name-or-uuid>`. This is because image names can be full docker images :shrug: This wasn't a super used feature, it's mostly only used so you can be super specific anyways, or do cool things like `unikraft instance delete $(unikraft instance ls --filter "<my-filter>")`, and have it get all the fully resolved keys.

At some point we should just print `index.<metro-endpoint>/<image>`, and have that operate as the full key. But this just doesn't work right now - there's still so much confusion over where an image will end up.

### DIFF
--- a/internal/cmd/images.go
+++ b/internal/cmd/images.go
@@ -1,0 +1,259 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/distribution/reference"
+	"github.com/opencontainers/go-digest"
+	"unikraft.com/cloud/sdk/platform"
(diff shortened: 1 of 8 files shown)


internal/cmd/images.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (Image) Get(ctx context.Context, keys []string) ([]resource.Resource, error) {
+	cl, err := multimetro.NewClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	resources, err := multimetro.DoKeys(ctx, cl, multimetro.ParseKeys(keys), func(ctx context.Context, mc *multimetro.MetroClient, keys multimetro.Keys) ([]resource.Resource, []multimetro.Key, error) {
+		log.G(ctx).Trace().Msg("getting images")
+		resp, err := mc.GetImages(ctx, platform.TagOrDigest{}, "")
```
nderjung, line 111: Looking forward to an update API so we don't have to list to get.

nderjung (approved):
Thanks!

PR #43, opened 2025-12-30
feat: Add resource lists

jedevc:
For example:
image

image

TODO:
- [x] XXX
- [x] Display lists in `ls` tables
- [x] More lists
- [x] Changed Patches to pointer, need to check that everything still works


nderjung (approved):
Looks good, thanks! The last commit seems as though it's part of a different story and would have been better in a separate PR but not going to fuss over it and will merge as-is. Cheers!

PR #46, opened 2026-01-05
test: Add better golden tests and run them in CI

jedevc:
This PR adds support for *experimental* sandboxing. Resources created in the
sandbox are tracked and isolated from other resources (to provide our testing framework a reliable clean environment). Sandboxes are persisted to disk as JSON files, which track resource types and keys that belong to the sandbox.

Additionally, we allow easily cleaning up resources that belong to the sandbox. Tests *should* ideally be well behaved and clean up after themselves, but they might not (due to an error, or some other weird thing) - we shouldn't leave dangling resources in this case.

Finally, we start running tests in CI.

### DIFF
--- a/cmd/unikraft/main_test.go
+++ b/cmd/unikraft/main_test.go
@@ -3,16 +3,22 @@
 // Licensed under the BSD-3-Clause License (the "License").
 // You may not use this file except in compliance with the License.
 
+//go:build integration
+
 package main
 
 import (
 	"bytes"
+	"context"
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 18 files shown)


cmd/unikraft/main_test.go
```diff
@@ -191,17 +286,69 @@ func wordCleanerf(word string, args ...any) *regexp.Regexp {
...
+				Endpoint: "http://api.ukp-stable.apw.unikraft.internal/",
+				Country:  "xx",
+			},
+			{
+				Name:     "ukp-staging",
+				Endpoint: "http://api.ukp-staging.apw.unikraft.internal/",
+				Country:  "xx",
+			},
+		},
+	}
```
jedevc: @ajpetersons @nderjung does this seem reasonable to you as a place to run these tests from CI?

I would need an account + token setup as a secret in CI, but that seems reasonable to me? The tests do try and tear-down neatly after themselves, and create their own isolation sandboxes, so only one account is required.

nderjung: I'm wondering if it's better to "secretly" inject this information (secret env var). Also allows us to swap out the backend transparently and re-run without requiring code changes.

.github/workflows/run-tests.yaml
```diff
@@ -0,0 +1,56 @@
+name: run-tests
```
nderjung: Prefer singular verb "`test`" to be inline with other workflows (`build`, `tag`, `release`, etc.)

cmd/unikraft/main_test.go
ajpetersons: from my POV running tests against `ukp-stable` and `ukp-staging` accomplishes exactly what those deployments are meant for.

one caveat I would consider is that those deployments are likely to have very different versions of software installed. meaning that the same request might produce different output (even in terms of syntax) on each of the metros, making the test very unpredictable, assuming that this uses both metros at the same time. 

another thing is that `ukp-staging` is expected to be volatile and could break cli tests without any change to the cli itself, though that's something we might be more willing to accept and expect to also be able to test against latest changes in the platform itself

nderjung (approved):
Thanks!

PR #47, opened 2026-01-05
ci: Introduce more linters

jedevc:
And fix the issues they reveal. The new linters are:
- `staticcheck`, finds a whole bunch of issues
- `errcheck`, ensures we actually explicitly handle/discard errors
- `govet`, same as `go vet`
- `ineffassign`, finds assignments we don't actually use
- `unused`, prevents unused code
- `godox`, prevents `XXX` comments from passing (which is a personal little habit to highlight things that need doing before merging).

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -4,8 +4,14 @@ run:
 linters:
   default: none
   enable:
-    - misspell
+    - staticcheck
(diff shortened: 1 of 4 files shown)


nderjung (approved):
Thanks!

PR #48, opened 2026-01-05
fix(mirror): Fix mirroring times

jedevc:
The `time.Time` structs weren't being correctly mirrored, resulting in bad values. This fixes the issue and adds a test.


nderjung (approved):
Thanks!

PR #51, opened 2026-01-06
feat: Add instance start/stop/restart

jedevc:
Depends on https://github.com/unikraft-cloud/proto/pull/185 and corresponding changes in https://github.com/unikraft-cloud/go-sdk.

This does need tests, but first we will need `instance create` or `run` or similar.


.crush.json
```diff
@@ -0,0 +1,11 @@
...
+    "$schema": "https://charm.land/crush.json",
+    "lsp": {
+        "go": {
+            "command": "gopls",
+            "env": {
+                "GOTOOLCHAIN": "go1.25.5"
+            }
+        }
+    }
+}
```
nderjung: Committed by mistake?

nderjung: (fine to have in the repo, but think it's a separate PR - maybe with others and for later?)

nderjung (approved):
Thanks!

PR #52, opened 2026-01-07
feat!: Replace --format with --output

jedevc:
I added this to originally be similar to the docker command. But it's a bit weird, and I don't love it. `jq` is well loved for doing scripting anyways, so we should just support a `--json` output for fields instead (not to be confused with `--raw` which is the raw API data).

### DIFF
--- a/internal/resource/cmd/cmd.go
+++ b/internal/resource/cmd/cmd.go
@@ -99,9 +99,7 @@ type FormatOpts struct {
 	Field  []string `short:"f" help:"Specify which fields to include in the output."`
 	Filter []string `help:"Filter output based on a field value (e.g. --filter status==active)." sep:"none"`
 
-	Quiet  bool   `short:"q" help:"Only display resource keys."`
-	Format string `help:"Format the output using a Go template."`
-	Raw    bool   `help:"Output raw JSON API response."`
+	Output Printer `short:"o" help:"Output format. One of: kv, table, json, yaml, raw, quiet, template."`
 }
 
 type ResourceListCmd[R resource.GettableResource] struct {
@@ -136,22 +134,13 @@ func (cmd *ResourceListCmd[R]) Run(ctx context.Context, cfg *config.Config, sand
 		if err != nil {
 			return err
 		}
-
 		resources, err = filterResources(resources, filter)
 		if err != nil {
 			return err
 		}
-
-		switch {
-		case cmd.Quiet:
-			return printQuiet(out, resources...)
-		case cmd.Raw:
-			return printRaw(out, resources...)
-		case cmd.Format != "":
-			return printTemplate(out, cmd.Format, resources...)
-		default:
-			return printTable[R](out, cmd.Field, resources...)
-		}
+		return cmd.FormatOpts.Output.
+			WithDefault(PrinterTypeTable).
+			Print(out, cmd.Field, empty, resources...)
 	}
 
 	if cmd.Watch > 0 {
@@ -186,22 +175,13 @@ func (cmd *ResourceInspectCmd[R]) Run(ctx context.Context, cfg *config.Config, s
 		if err != nil {
 			return err
 		}
-
 		resources, err = filterResources(resources, filter)
 		if err != nil {
 			return err
 		}
-
-		switch {
-		case cmd.Quiet:
-			return printQuiet(out, resources...)
-		case cmd.Raw:
-			return printRaw(out, resources...)
-		case cmd.Format != "":
-			return printTemplate(out, cmd.Format, resources...)
-		default:
-			return printInspect(out, cmd.Field, resources...)
-		}
+		return cmd.FormatOpts.Output.
+			WithDefault(PrinterTypeKeyValue).
+			Print(out, cmd.Field, empty, resources...)
 	}
 
 	if cmd.Watch > 0 {
@@ -424,5 +404,5 @@ func (cmd *ResourceCreateCmd[R]) Run(ctx context.Context, cfg *config.Config, sa
 	if err != nil {
 		return err
 	}
-	return printInspect(cfg.Stdout, nil, res)
+	return printKVFormatted(cfg.Stdout, nil, res)
 }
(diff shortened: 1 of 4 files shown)


nderjung:
I am wondering about the naming here.  We use `--raw` and I wonder if that should be called `--json` instead since the outputs for both are JSON.  This new `--json` flag at a high-level just post-processes this "raw" data anyway and then re-structures it and actually removes some information.  At least before, the `--format` flag allowed to you to stuff without having to install `jq`.

jedevc:
Mmm, agreed naming is tricky. I don't really want to call the platform response directly `--json`, since I would expect that it outputs *exactly* what I see as a JSON object instead. However, note that currently that the API info doesn't include metro information, or any other of the other processing we do (especially for images).

Regardless of what we do for `--format`, I think keeping the platform API output as `--raw` or maybe even `--api` would be best for clarity.

> At least before, the `--format` flag allowed to you to stuff without having to install jq.

Good point - and it was *technically* possible to get it as JSON before as well: using `--format "{{ toJson . }}"` (super elegant obviously). But I personally don't find writing go templates a super easy way to format custom output - I don't really know any tools that do this aside from docker.

nderjung:
> Mmm, agreed naming is tricky. I don't really want to call the platform response directly `--json`, since I would expect that it outputs _exactly_ what I see as a JSON object instead. However, note that currently that the API info doesn't include metro information, or any other of the other processing we do (especially for images).

That's i think a very important point; it misses the metro information.  Not sure if our discussion on trying to get that data into the API natively affects this?

> Regardless of what we do for `--format`, I think keeping the platform API output as `--raw` or maybe even `--api` would be best for clarity.

That makes sense.  Either-or work for me.

> > At least before, the `--format` flag allowed to you to stuff without having to install jq.
> 
> Good point - and it was _technically_ possible to get it as JSON before as well: using `--format "{{ toJson . }}"` (super elegant obviously). But I personally don't find writing go templates a super easy way to format custom output - I don't really know any tools that do this aside from docker.

I agree it's not the best but I think powerful enough as-is to leave in as-is even if most of the time people use `jq`, for the use case where you may not have it.  I think writing go templates in the CLI does suck but if you have a file it's much easier - i wonder if allowing to read in from stdin may make sense if you have a more complicated template with if-statements and then you pipe it in; but that's also a bit of an esoteric use case also. :D

jedevc:
Okay, to move this forward, how about:
- *keeping* `--format` for cases where users don't have `jq` installed and really like writing go templates. But maybe rename it to `--template` to be clear, and disambiguate it from the "f"y options: `--field` and `-f`/`--filter`
- *adding* `--json` as proposed here
- *keeping* `--raw` to simply show the raw platform response

nderjung:
I fear overloading the term with 'template':
- `unikraft instance create --template`
- `unikraft instance template verb --template`

Or does this only apply to `list` and `get` ?

jedevc:
Argh, you're not wrooong. Ideally I want the formatting opts to apply everywhere, not just to `get` and `list` though that's where they're supported now.

So yeah, maybe not. I just don't love the current name of `--format`, I don't want to inherit that name from the docker CLI. Honestly, there's not a lot of CLIs with this kind of functionality - I do actually think that having a good JSON output format will be good enough for most use cases. If users ask for something like go templates in the CLI again, we can definitely add it back in (once we've got an idea of how everything fits together better).

jedevc:
Alternatively. What about something like `-o, --output` like `kubectl`?

It allows:
- `-o json`
- `-o yaml`
- `-o template=<template goes here>`, `-o templatefile=<path to template goes here>`
- `-o jsonpath=...`
- And more.

We don't need to support *all* of them, but it's a neat kind of way of selecting them. We could also allow `-o table` to format as a table and `-o kv` to format as key-values (so you could get `inspect` or `edit` to format as a table if you really wanted to).

Note, we would still support `--raw` I think, since this is different from output formatting - you're actually outputting something different, but just styling it differently. And we'd still support `-f, --field`, because you can choose what fields are selected for all of the above (e.g. you could choose to do `-o json -fuuid,runtime.args` to get all the uuids and the args they have as a JSON object).

nderjung:
Agreeing with all of this :)

jedevc:
> Agreeing with all of this :)

Wonderful! I've updated to do this then :tada:

:warning: will have test failures when merged with https://github.com/unikraft/cli/pull/62 (since those tests use `--format` and need to be updated)

internal/resource/cmd/cmd.go
```diff
@@ -50,9 +50,7 @@ type FormatOpts struct {
 	Field  []string `short:"f" help:"Specify which fields to include in the output."`
 	Filter []string `help:"Filter output based on a field value (e.g. --filter status==active)." sep:"none"`
 
-	Quiet  bool   `short:"q" help:"Only display resource keys."`
-	Format string `help:"Format the output using a Go template."`
-	Raw    bool   `help:"Output raw JSON API response."`
+	Output Printer `short:"o" help:"Output format. One of: kv, table, json, yaml, raw, quiet, template."`
```
nderjung, line 102: You can use the kong `enum` tag to enforce the value:
```suggestion
	Output Printer `short:"o" help:"Output format." enum:"kv,table,json,yaml, raw,quiet,template" default:"table"`
```

nderjung (commented):
One suggestion otherwise works like a charm!

internal/resource/cmd/cmd.go
jedevc: This doesn't quite work for `template`, since it's `template=...`, right?

nderjung: I see. Yes this will need [a custom mapper decoder](https://github.com/alecthomas/kong/blob/98b7aff57abc5f59dfe8fad5ba057bca4920d875/README.md?plain=1#L736-L753). I'll open up as a separate low-prio issue (as it would be nice to keep flag enums consistent across the codebase). Merging as-is.

nderjung (approved):
Thanks!

PR #57, opened 2026-01-08
feat: Editing list fields

jedevc:
This is a big yak shave for `unikraft run`:

- Started with `unikraft instance create`
- Realized we needed to be able to attach to a service group
- Currently cannot create a service group, because we can't patch the list of `Services`, so it can't actually expose anything.
- So scaled back to working on `unikraft service edit` (and `unikraft service create` later)

This all kind of culimates in: [asciicast](https://asciinema.org/a/MDmU8dsa0YOvVpe2)

There's a *lot* in here. But the main bit is that we now support editing list fields, as well as working with fields that contain both values + children (e.g. the new `Service` type that can be referred to in shorthand `<src>:<dest>/<handler>`).

### DIFF
--- a/internal/cmd/services.go
+++ b/internal/cmd/services.go
@@ -60,9 +60,9 @@ type ServiceGroup struct {
 }
 
 type Service struct {
-	Source      uint32                     `mirror:"port" json:"source" field:",long"`
-	Destination uint32                     `mirror:"destination_port" json:"destination" field:",long"`
-	Handlers    []platform.ServiceHandlers `mirror:"handlers" json:"handlers" field:",long"`
+	Source      uint32                     `mirror:"port" json:"source" field:",short"`
+	Destination uint32                     `mirror:"destination_port" json:"destination" field:",short"`
+	Handlers    []platform.ServiceHandlers `mirror:"handlers" json:"handlers" field:",short"`
 }
 
 func (s *Service) String() string {
--- a/internal/cmd/services.go
+++ b/internal/cmd/services.go
@@ -23,8 +23,10 @@ import (
 )
 
 type ServicesCmd struct {
-	resource.ResourceCmd[ServiceGroup]         `set:"name=service" set:"names=services"`
-	resource.EditableResourceCmd[ServiceGroup] `set:"name=service" set:"names=services"`
+	resource.ResourceCmd[ServiceGroup]          `set:"name=service" set:"names=services"`
+	resource.DeletableResourceCmd[ServiceGroup] `set:"name=service" set:"names=services"`
+	resource.EditableResourceCmd[ServiceGroup]  `set:"name=service" set:"names=services"`
+	resource.CreatableResourceCmd[ServiceGroup] `set:"name=service" set:"names=services"`
 }
 
 type ServiceGroup struct {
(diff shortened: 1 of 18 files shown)


internal/multimetro/key.go
```diff
@@ -20,7 +20,7 @@ type Key struct {
 }
 
 func ParseKey(s string) Key {
-	if metro, key, ok := strings.Cut(s, "/"); ok {
+	if metro, key, ok := strings.Cut(s, "#"); ok {
```
nderjung: Can we make this a `const`:
```go
// MetroKeySeparator defines the symbol used to ...
const MetroKeySeparator = "#"
```

internal/cmd/images.go
```diff
@@ -0,0 +1,259 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
```
nderjung: It's now 2026 :)

internal/multimetro/key.go
jedevc: Updated in https://github.com/unikraft/cli/pull/39, which is what this PR is based on.

internal/cmd/images.go
jedevc: Updated in https://github.com/unikraft/cli/pull/39.

internal/resource/resource_helpers.go
```diff
@@ -5,8 +5,20 @@
 
 package resource
 
+import (
+	"fmt"
+	"reflect"
+	"slices"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
```
nderjung: Mitchell's `mapstructure` module has been deprecated; prefer the `v2` fork: https://github.com/go-viper/mapstructure:
```diff
-	"github.com/mitchellh/mapstructure"
+	"github.com/go-viper/mapstructure/v2"
```

internal/resource/patch.go
```diff
@@ -14,6 +14,8 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+
+	"github.com/mitchellh/mapstructure"
```
nderjung: See previous comment

internal/resource/resource_helpers.go
nderjung: Ignore this suggestion. I can see that mapstructure is already a dependency of the project. We can leave this suggestion for another PR.

internal/resource/patch.go
nderjung: Ignore.

internal/cmd/services.go
```diff
@@ -41,27 +44,67 @@ type ServiceGroup struct {
...
+	s.Source = uint32(srcPort)
+
+	destPort, err := strconv.Atoi(dest)
+	if err != nil {
+		return fmt.Errorf("invalid destination port: %w", err)
+	}
+	s.Destination = uint32(destPort)
+
+	for handler := range strings.SplitSeq(handlers, "+") {
+		s.Handlers = append(s.Handlers, platform.ServiceHandlers(handler))
```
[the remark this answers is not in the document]
nderjung: It would also be nice to support aliases like `http` and `https` to imply `80` and `443`, e.g.:

```
https:8080
```
implies
```
443:8080/http+tls
```
but also happy to let that be future work.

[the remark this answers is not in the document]
jedevc: Future work!

jedevc:
Pushed a couple more commits - this now allows the full syntax I proposed originally:
<a href="https://asciinema.org/a/a3ol6ze24pyF0sXV" target="_blank"></a>

jedevc:
Aaaaand added some tests :tada:

internal/cmd/services.go
```diff
@@ -41,27 +46,71 @@ type ServiceGroup struct {
...
+	if err != nil {
+		return fmt.Errorf("invalid source port: %w", err)
+	}
+	s.Source = uint32(srcPort)
+
+	destPort, err := strconv.Atoi(dest)
+	if err != nil {
+		return fmt.Errorf("invalid destination port: %w", err)
+	}
+	s.Destination = uint32(destPort)
```
nderjung, line 93: Idk if my comment got lost - but i was wondering if it makes sense to add aliases here for 443 and 80, e.g.:
```
https:8080
```
which may translate to
```
443:8080/http+tls
```
Wdyt?

jedevc: Probably at some point. I just followed the syntax at https://unikraft.com/docs/platform/services#handlers

internal/cmd/services.go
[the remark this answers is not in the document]
nderjung: AH i see it now, ignore my last comment

internal/cmd/services.go
nderjung: ignore - i found my previous comment

nderjung (approved):
Brilliant, thanks!

PR #60, opened 2026-01-13
fix: Follow-ups from editing list fields

jedevc:
See individual commits for details.


nderjung (approved):
Thanks!

PR #62, opened 2026-01-13
feat: Add unikraft instance create

jedevc:
Please be gentle in review :pray: 

The changes to `instances.go` are somewhat garbage - this whole editing + patching system needs a heavy rework, but I'm going to do it later.

### DIFF
--- a/cmd/unikraft/testdata/TestGolden/instances
+++ b/cmd/unikraft/testdata/TestGolden/instances
@@ -1,0 +1,120 @@
+$ unikraft login
+
+stderr:
+	HH:MM INF using authentication token from UKC_TOKEN environment variable
+	HH:MM WRN no metros configured; please add them manually
+	HH:MM INF login successful
+
+$ unikraft instance list
+
+stdout:
+	METRO  NAME  STATE  IMAGE  ARGS  MEMORY  VCPUS  SERVICES  FQDN
+
+$ unikraft instance create --set name=test-$UNIQ_INST --set metro=test --set image=nginx:latest --set autostart=true --set resources.memory=128 --set resources.vcpus=1 --set service.services=443:8080/tls+http --set service.domains=name=$UNIQ_DOMAIN
+
+stdout:
+	metro:        test
+	name:         test-<INST>
+	uuid:         12345678-1234-1234-1234-123456789abc
+	state:        running
+	image:        nginx
+	resources:
+	  memory:     128
+	  vcpus:      1
+	service:
+	  uuid:       12345678-1234-1234-1234-123456789abc
+	  name:       <SERVICE_NAME>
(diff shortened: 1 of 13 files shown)


cmd/unikraft/testdata/TestGolden/instances
```diff
@@ -0,0 +1,125 @@
...
+	METRO  NAME               STATE    IMAGE  ARGS  MEMORY  VCPUS  SERVICES  FQDN
+	test   test-<INST>  running  nginx  []    128     1                <DOMAIN>.unikraft.internal
+
+$ unikraft instance inspect test-$UNIQ_INST
+
+stdout:
+	metro:        test
+	name:         test-<INST>
+	uuid:         12345678-1234-1234-1234-123456789abc
+	state:        running
```
nderjung, line 46: There's an edge-case where the state is in `starting` state (which could also be valid).

cmd/unikraft/testdata/TestGolden/instances
```diff
@@ -0,0 +1,125 @@
...
+	  - fqdn:     <DOMAIN>.unikraft.internal
+	networks:
+	- uuid:       12345678-1234-1234-1234-123456789abc
+	  private-ip: 10.X.X.X
+	  mac:        aa:bb:cc:dd:ee:ff
+
+$ unikraft instance logs test-$UNIQ_INST
+
+stdout:
+	Powered by Unikraft Kiviuq (0.20.0~f5319a0)55.252 gw 10.X.X.X
```
nderjung: Possible to just check if it `Powered by Unikraft` exists? Output will vary over-time with new Unikraft versions.

nderjung (commented):
As discussed offline, will merge as-is - just want to get the test to pass at least as this won't change (see inline comments).

cmd/unikraft/testdata/TestGolden/instances
jedevc: Just removed the instance logs entirely for now.

cmd/unikraft/testdata/TestGolden/instances
jedevc: Added a cleaner for this case. This made me think though, we should have another resource command - `wait`. This command adds a filter-like expression, and waits for it to become true.

So you could do:

```
unikraft instance wait <name> --until state==running
```

Then it would poll at regular intervals, and block until the result is true.

Or maybe this would just be a flag to `create` / `edit` / etc instead. Though I do like the idea of the standalone command.

nderjung: I like the idea of a standalone command. I've opened a ticket in triage.

nderjung (approved):
Thanks!

PR #63, opened 2026-01-14
refactor: Split resource package and add basic command tests

jedevc:
I've wanted to do the big resource package split for a while - this ensures that we can split up functionality to be a bit clearer.

The tests are also designed so that we can test command behavior outside of the integration tests.


nderjung (approved):
Thanks!

PR #65, opened 2026-01-14
fix: Don't hide empty fields with -fall set

jedevc:
This was an oversight, it should have worked like this when empty filtering was added.


nderjung:
Just needs a rebase :)

nderjung (approved):
Thanks!

PR #67, opened 2026-01-16
refactor: Simplify patch population

jedevc:
We can automatically pull these from the type of the field, we don't need to do the weird iteration + population in most cases.

Resolves https://linear.app/unikraft/issue/CLI-57/simpler-patch-creation


nderjung (approved):
Thanks!

PR #69, opened 2026-01-16
feat: Add --dry-run args to edit and delete

jedevc:


jedevc:
Tests are legitimately failing, will fix.

internal/kvwriter/kvwriter.go
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BSD-3-Clause
-// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
+// Copyright (c) 2025, Unikraft dthGmbH and The Unikraft CLI Authors.
```
nderjung: Unikraft Durchführung technischer Handelsgeschäfte Gesellschaft mit beschränkter Haftung

jedevc: Uhhhh. I blame vim.

internal/resource/cmd/cmd.go
```diff
@@ -266,6 +267,10 @@ func (cmd *ResourceEditCmd[R]) Run(ctx context.Context, cfg *config.Config, sand
 	}
 	patched = patch.FilterPatchableFields(patched)
 
+	if cmd.DryRun {
+		return printPatches(cfg.Stdout, patched, false)
+	}
```
nderjung, line 334: I am wondering - why not show the diff like before?

```diff
unikraft instance edit --set resources.memory=1024 fra#name:nginx-v98x9
  metro:        fra
  name:         nginx-v98x9
  uuid:         9f5baba7-2591-45ec-ac56-d00e9ce0e81d
  state:        stopped
  image:        nginx
  resources:
-   memory:     128
+   memory:     1024
    vcpus:      1
  service:
    uuid:       2451f2f8-ec54-4be9-9b64-c46c506c36b2
    name:       fragrant-field-6q4l1725
    domains:
    - fqdn:     fragrant-field-6q4l1725.fra.unikraft.app
  networks:
  - uuid:       07920f85-6a6e-4bce-88b1-20d67a933e3f
    private-ip: 10.0.1.233
    mac:        12:b0:0a:00:01:e9
```
```
unikraft instance edit --dry-run --set resources.memory=1024 fra#name:nginx-v98x9
resources.memory  := 512
```

`printPatches` and `dmp.DiffMain` feel like two different ways to output the same information - does it make sense to instead have `--diff=pretty` and `--diff=text` (or similar)?

jedevc: Ahh I mentioned this the other day - but it's because there's not a perfect alignment between the request you make and the response you get. e.g. for certificates you provide a PEM encoded certificate and get back fields from the parsed certificate. The types also don't always align, e.g. you can create a service when creating an instance, but you can't view those service details when viewing the instance later (so what would we show here? The only bit that actually appears is the name/uuid - but we might not know those if they're autogenerated).

We could kinda just do a best effort guess, e.g. for `set` we can just assume that it overrides the previous result. The main issue with that is that then `--add` and `--del` need to be entirely implemented in the CLI to emulate what the API would do 😭 

All of this is certainly possible - it's just a super non trivial amount of work. IMO if we wanted these perfect dry runs, we would want the API itself to support these, otherwise, we're just building the API again in the CLI.

(Note: we kinda support this for visual editing, which is why visual editing needs refactoring. It's a bit easier because we explicitly *don't* support add/del here, only absolute set. We also *don't* show any field that can't be edited).

But I get the point about wanting consistent output. Maybe dry-run is the wrong name for it? Essentially I just want to see the patches that would be applied, without actually attempting to apply them.

jedevc: FYI, this is kind of the comment explaining a bit more about the refactoring that needs doing around visual editing and why it's so much of a mess right now: https://linear.app/unikraft/issue/CLI-56/better-behavior-for-visual-editing

TL;DR - I'd originally assumed there was a 1-1 relationship between the field you edit and the field you get back. As it turns out, nope, not really.

nderjung (approved):
Merging as-is. 

I think we can later improve, if we want, how the output is presented to the user, but this works fine for and is functional.

PR #70, opened 2026-01-16
feat: Add unikraft run

jedevc:
Self explanatory hopefully :tada:

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -1,0 +1,183 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/multimetro"
+	"unikraft.com/cli/internal/resource"
+	"unikraft.com/cli/internal/resource/cmd"
+	"unikraft.com/cli/internal/resource/patch"
+)
+
+type RunCmd struct {
+	Image string   `arg:"" help:"Image to run."`
+	Args  []string `arg:"" optional:"" help:"Arguments to pass to the instance."`
+
+	Name  string `short:"n" help:"Name of the instance."`
+	Metro string `help:"Metro to deploy the instance in." required:"" placeholder:"metro"`
+
+	Env    []string `short:"e" help:"Set environment variables (KEY=VALUE)."`
+	Memory int      `short:"m" help:"Memory in MB."`
+	Volume []string `short:"v" help:"Attach a volume (NAME:AT[:ro] or NAME:SIZE:AT[:ro] or UUID:AT[:ro])."`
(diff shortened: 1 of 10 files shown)


internal/cmd/run.go
```diff
@@ -0,0 +1,317 @@
...
+	ServiceName         string   `help:"Service group name."`
+	ServiceUUID         string   `help:"Service group UUID."`
+	ServiceSoftLimit    uint32   `help:"Service group soft limit."`
+	ServiceHardLimit    uint32   `help:"Service group hard limit."`
+	RestartPolicy       string   `help:"Restart policy."`
+	ScaleToZero         bool     `help:"Enable scale-to-zero."`
+	ScaleToZeroPolicy   string   `help:"Scale-to-zero policy."`
+	ScaleToZeroStateful bool     `help:"Enable stateful scale-to-zero."`
+	ScaleToZeroCooldown int64    `help:"Scale-to-zero cooldown time in milliseconds."`
+	Autostart           bool     `help:"Start the instance automatically."`
```
nderjung: Should it maybe auto-start by default?

internal/cmd/run.go
```diff
@@ -0,0 +1,317 @@
...
+	DryRun bool `help:"Show the create preview without executing."`
+
+	Publish             []string `short:"p" help:"Publish a service port (SOURCE:DESTINATION[/HANDLERS])."`
+	Domain              []string `help:"Add a service domain (FQDN)."`
+	ServiceName         string   `help:"Service group name."`
+	ServiceUUID         string   `help:"Service group UUID."`
+	ServiceSoftLimit    uint32   `help:"Service group soft limit."`
+	ServiceHardLimit    uint32   `help:"Service group hard limit."`
+	RestartPolicy       string   `help:"Restart policy."`
+	ScaleToZero         bool     `help:"Enable scale-to-zero."`
```
nderjung: Also scale-to-zero on by default?

internal/cmd/run.go
jedevc: Yup, definitely.

internal/cmd/run.go
jedevc: So that definitely mimics the behavior of `kraft`. But I'm not entirely sure that that's expected given how `docker run` works.

I've made it significantly easier to enable, so now it's just `--scale-to-zero=`, and it's enable scale to zero.

nderjung:
After the latest series of merges, i'm now getting this:
```
unikraft.com/cli/internal/cmd
# unikraft.com/cli/internal/cmd
internal/cmd/run.go:213:41: cannot use "" (untyped string constant) as kvwriter.KeyValueOpt value in argument to kvwriter.KeyValueWriter
internal/cmd/run.go:218:52: resource.Field{…}.ValueString undefined (type resource.Field has no field or method ValueString)
make: *** [unikraft] Error 1
```

jedevc:
Marking as draft so as to not merge accidentally - this one needs a few tricky changes to make `--dry-run` work right. Will do this in my morning.

jedevc:
Ready now :tada:

nderjung (approved):
Awesome, thanks! I do think we may end up tweaking some of the top-level flags for `unikraft run` but we can leave this for later. Cheers!

PR #71, opened 2026-01-16
chore: Update README.md

jedevc:
With a nice little setup guide, and getting started section that folks can follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,107 @@
 
 > [!WARNING]
 >
-> Under construction!
+> Under construction! :construction_worker:
 
 This repository contains the code for a work-in-progress, internal tool for
-managing the Unikraft Cloud Platform.
+managing the [Unikraft Cloud Platform](https://unikraft.com/). Eventually, this will replace [kraft](https://github.com/unikraft/kraftkit).
+
+## Setup
+
+To clone the repository, run:
+
+```bash
+git clone https://github.com/unikraft/cli.git
+cd cli
+```
+
+This repo uses `task` (from <https://taskfile.dev/>) to manage builds and other tasks.
+To install `task`, follow the [installation guide](https://taskfile.dev/docs/installation).
+
+You'll also need to enable `task`'s [Remote Taskfiles](https://taskfile.dev/docs/experiments/remote-taskfiles) feature:
+
+```bash
+export TASK_X_REMOTE_TASKFILES=1
+```
+
+## Building
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -2,7 +2,106 @@
...
+```
+
+This repo uses `task` (from <https://taskfile.dev/>) to manage builds and other tasks.
+To install `task`, follow the [installation guide](https://taskfile.dev/docs/installation).
+
+You'll also need to enable `task`'s [Remote Taskfiles](https://taskfile.dev/docs/experiments/remote-taskfiles) feature:
+
+```bash
+export TASK_X_REMOTE_TASKFILES=1
+```
```
nderjung, line 26: If we have a top-level `Makefile` like this, then we can abstract using `task`:
```Makefile
Q    ?= @
GO   ?= go
ifeq ($(shell command -v task >/dev/null 2>&1 && echo yes),)
TASK ?= $(GO) run -v github.com/go-task/task/v3/cmd/task@v3.44.1 --yes
else
TASK ?= task --yes
endif

export TASK_X_REMOTE_TASKFILES=1

ifeq ($(Q),)
TASK := $(TASK) --verbose
endif

.DEFAULT_GOAL := help
.PHONY: help

help:
	$(Q)$(TASK) -l

%:
	$(Q)$(TASK) $(MAKECMDGOALS)
```

Calling:
- `make help` would show you all the targets
- `make XXX` (e.g. `make cli`) proxies things to `task`.

If `task` is not installed, it calls `go run .../cmd/task` so the user does not have to install anything.  Calling `make XXX` will always work therefore.

Additionally, it adds `export TASK_X_REMOTE_TASKFILES=1` just to make things also easier with the remote taskfile.

README.md
```diff
@@ -2,7 +2,106 @@
...
+To use the `unikraft` CLI, you'll first need to login:
+
+```bash
+unikraft login
+```
+
+Alternatively, you can manually create a profile:
+
+```yaml
+# ~/.config/unikraft/config.yaml
```
nderjung:
```
# Linux:
# ~/.config/unikraft/config.yaml
# macOS:
# ~/Library/Application\ Support/unikraft/config.yaml
```

README.md
jedevc: I can :sweat: But I was kind of hoping at some point to actually just remove the Makefile, and only have a Taskfile. I think it's a fair warning there, I'd rather not just silence it automatically.

nderjung (approved):
Thanks!

PR #72, opened 2026-01-16
refactor!: Update config

jedevc:
There were a few issues with the config loading + global flags that needed fixing, so this gets all of them:

- Some flags weren't used and won't be, e.g. `emoji`

- Some flags weren't working, e.g. `--config` which has a circular dependency - to know the file you need to parse the CLI, to parse the CLI you need to provide the config file.

- Almost every flag could be controlled by an env var, which makes sense for some, like `UNIKRAFT_LOG_LEVEL`, but really doesn't for others, like `UNIKRAFT_RAW` on `unikraft instance ls`. This behavior *shouldn't* come from env vars, it should be very explicit on the CLI for two main reasons. Firstly, it's hard to debug a user setup if functionality is potentially being inherited from an env var somewhere, and second, it means that commands aren't fully copy-pasteable. As much as possible, a command that works on one machine should easily work on another.

To do this, we split the previous `Config` struct up, and put most of the functionality into a `globalFlags` struct, while making sure that the config file contains only the bare minimum (currently the authentication) details. We then pass around a lot of the details of the command using `kong.Bind`, which allows dynamically binding to content that the root specifies.


cmd/unikraft/main.go
```diff
@@ -32,24 +31,27 @@ func main() {
...
 	if err != nil && !errors.Is(err, context.Canceled) {
-		if config.G(ctx).LogType == log.TextType || config.G(ctx).LogType == "" {
-			log.G(ctx).Error().Msg(" ")
-			log.G(ctx).Error().Msg(colors.ErrorFg("error:"))
-		}
+		// XXX: this is weird
+		// if config.G(ctx).LogType == log.TextType || config.G(ctx).LogType == "" {
+		log.G(ctx).Error().Msg(" ")
+		log.G(ctx).Error().Msg(colors.ErrorFg("error:"))
+		// }
```
jedevc: @nderjung curious what you were thinking of doing to improve the logging here.

I personally don't really like the (ab)use of text logs to display pretty errors here. It's weird that we need to change what we output based on the log type, and it would be good to avoid needing to propagate that logic everywhere.

It's also weird because if we wanted to add log ingestion (e.g. to sentry), we'd also end up mixing that logic in as well.

But yeah, having nicely formatted errors is nice 😅 I'm not 100% sure what my suggestion is here - but I think we need a way of displaying pretty errors, that writes directly to stderr (and also writes to the logs).

nderjung: Yeah, it's definitely a hacky way to make more whitespace around the error to increase visibility. As you say, it would be better to have a way to nicely unwind errors depending on context. This doesn't happen often, but i see more places where it would be nice to have more whitespace around the actual log line, so that it doesn't feel as condensed and stands out more for the user.

I'm also not sure the best approach would be.  Maybe we could add a wrapper around `log.G(ctx)...Msg()` when we do want to "pronounce" certain messages (e.g. `log.Pronounce(log.G(ctx)...Msg())`. Or if we could extend zerolog somehow to have `log.G(ctx)...HighlightMsg()`.

As for sentry, I think we're going to stop using this in favour of a different log system and when we do emit logs, it would only be in internal things we run. End-users would never have logs sent - only crash telemetry.  In our systems, we can ensure we don't mangle output by always using JSON. So the compromise for now I think is to leave it in just so we can have a nicer output in text-mode and re-visit when we can better formalise the need to "highlight"/"pronounce" certain messages. Wdyt?

jedevc: Note: somehow this is currently wrong on `staging`.

```
{"level":"error","time":"2026-01-28T12:52:17+02:00","message":" "}
{"level":"error","time":"2026-01-28T12:52:17+02:00","message":"\u001b[38;2;236;0;63merror:\u001b[0m"}
{"level":"error","time":"2026-01-28T12:52:17+02:00","message":"  parsing arguments: expected one of \"completion\", \"login\", \"logout\", \"profile\", \"run\", ..."}
{"level":"error","time":"2026-01-28T12:52:17+02:00","message":" "}
```

jedevc: Fixed this!

internal/cmd/profile.go
```diff
@@ -63,7 +63,7 @@ func (Profile) List(ctx context.Context) ([]resource.Resource, error) {
 
 		result := Profile{
 			Name:   profile.Name,
-			Active: profile.Name == cfg.Profile,
+			Active: profile.Name == cfg.DefaultProfile,
```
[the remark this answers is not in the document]
jedevc: This is intentional.

cmd/unikraft/main.go
```diff
@@ -94,19 +96,19 @@ func getMethod(value reflect.Value, name string) reflect.Value {
 	return method
 }
 
-func run(ctx context.Context, args []string, stdin io.Reader, stdout, stderr io.Writer) (context.Context, error) {
-	cli, opts, cleanup, err := cmd.NewRootCmd(ctx, args, stdin, stdout, stderr)
+func run(ctx context.Context, args []string, stdio config.Stdio) (context.Context, error) {
+	ctx2, cli, opts, cleanup, err := cmd.NewRootCmd(ctx, args, stdio)
+	if ctx2 != nil {
+		ctx = ctx2
+	}
```
nderjung: AFAICT `cmd.NewRootCmd` does not return a `nil` for `ctx`:
```diff
- 	ctx2, cli, opts, cleanup, err := cmd.NewRootCmd(ctx, args, stdio)
- 	if ctx2 != nil {
- 		ctx = ctx2
- 	}
+ 	ctx, cli, opts, cleanup, err := cmd.NewRootCmd(ctx, args, stdio)
```

jedevc: Good catch :pray: :tada:

nderjung (approved):
Thanks!

PR #75, opened 2026-01-16
feat: Show help when running parent command

jedevc:
Before:

```
❯ unikraft
8:10PM ERR  
8:10PM ERR error:
8:10PM ERR   parsing arguments: expected one of "login", "logout", "profile", "metros", "instances", ...
8:10PM ERR  
```

After:

```
❯ unikraft
Usage:
  unikraft <command> [flags]

Version:
  dev (linux/amd64)

Commands:
  login
          Login to Unikraft Cloud.
  logout
          Logout from Unikraft Cloud.
  profile, profiles
          Manage Unikraft Cloud profiles.
  metros, metro
          Manage Unikraft Cloud metros.
  instances, instance
          Manage Unikraft Cloud instances.
  volumes, volume
          Manage Unikraft Cloud volumes.
  services, service
          Manage Unikraft Cloud services.
  certificates, certificate
          Manage Unikraft Cloud certificates.
  images, image
          Manage Unikraft Cloud images.

Global flags:
  -h, --help
          Show context-sensitive help.
  -c, --config ($UNIKRAFT_CONFIG)
          Set the configuration file.
      --[no-]telemetry ($UNIKRAFT_TELEMETRY)
          Enable or disable telemetry.
          [default: true]
      --[no-]emojis ($UNIKRAFT_EMOJIS)
          Enable or disable emojis in the CLI output.
          [default: true]
      --log-level ($UNIKRAFT_LOG_LEVEL)
          Set the logging level.
          [default: info, choice: trace,debug,info,warn,error,fatal]
      --log-type ($UNIKRAFT_LOG_TYPE)
          Set the log type.
          [default: text, choice: text,json]
      --profile ($UNIKRAFT_PROFILE)
          Set the current profile.
          [default: default]
  -v, --version
          Print version information.

Run "unikraft <command> --help" for more information on a command.
```

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -7,11 +7,15 @@ package cmd
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"io"
 	"maps"
+	"os"
 	"path/filepath"
 	"runtime"
 	"slices"
+	"strings"
 
 	"github.com/alecthomas/kong"
 	kongyaml "github.com/alecthomas/kong-yaml"
@@ -53,6 +57,18 @@ func NewRootCmd(ctx context.Context, args []string, stdin io.Reader, stdout, std
 	parser.Stderr = stderr
 
 	kctx, err := parser.Parse(args)
+
+	var parseErr *kong.ParseError
+	if errors.As(err, &parseErr) {
+		// HACK: kong provides UsageOnError, but this shows help for *all* parse
+		// errors - we only want to show it only for parent commands.
+		// See https://github.com/alecthomas/kong/issues/33
+		if strings.HasPrefix(parseErr.Error(), "expected one of") {
+			_ = parseErr.Context.PrintUsage(false)
+			fmt.Fprintln(os.Stdout)
+		}
+	}
+
 	if err != nil {
 		return nil, nil, nil, jujuerrors.Annotate(err, "parsing arguments")
 	}
--- a/cmd/unikraft/testdata/TestGolden/empty
+++ b/cmd/unikraft/testdata/TestGolden/empty
@@ -1,5 +1,57 @@
 $ unikraft
 
+stdout:
+	Usage:
+	  unikraft <command> [flags]
+
+	Version:
+	  dev (GOOS/GOARCH)
+
+	Commands:
+	  login
+	          Login to Unikraft Cloud.
+	  logout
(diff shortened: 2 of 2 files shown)


internal/cmd/root.go
```diff
@@ -54,7 +56,17 @@ func NewRootCmd(ctx context.Context, args []string, stdin io.Reader, stdout, std
...
 	kctx, err := parser.Parse(args)
 	if err != nil {
-		return nil, nil, nil, jujuerrors.Annotate(err, "parsing arguments")
+		if shouldShowHelp(err) {
+			helpArgs := append(slices.Clone(args), "--help")
+			helpKctx, helpErr := parser.Parse(helpArgs)
+			if helpErr == nil {
+				kctx = helpKctx
+				err = nil
+			}
```
nderjung: Oh really don't love this (running `parser.Parse`) again.

Another solution is not very nice either, basically embedding the `parser.Parse(args)` function here directly. But then you lose all the "apply hooks" (not that we use them, but you never know) - unless you embed that functionality ALSO and making a custom help flag. [I already asked Alec if he could expose `applyHook` and he said no](https://github.com/alecthomas/kong/pull/546).  Embedding `parser.Parser` would allow us to better handle error messages, also with config parsing, which is why i originally wanted it (i.e. "could not parse `config.yaml`, syntax mistake at line X").

jedevc: > Oh really don't love this (running `parser.Parse`) again.

Yeah, likewise - hence big `HACK` comment. It definitely sucks to not have a better way to do this, but IMO the right way to do this is https://github.com/alecthomas/kong/issues/33. We could actually do it if there was a better typed error for what actually went wrong as well - but yeah, eventually this will take upstream changes.

I guess there's a few options here:
- Do one of the above hacks (the re-parse, or the full embedding of `Parse`, my preference on the first, since it's a bit shorter), then work on a better upstream solution in the background
- Do the upstreaming, and don't merge anything till then.

IMO, the upstreaming is gonna take a moment, since I'm so focused on shipping a usable CLI this next week. I would argue we should just pick one of the hacks for now.

jedevc: Actually woops. There's a better hack we can do - `parseErr` has a `Context` you can call `PrintUsage` on.

nderjung (approved):
Thanks!

PR #76, opened 2026-01-17
fix: Ensure image fields are correctly parsed

jedevc:
Due to a change in how fields were parsed, we were incorrectly inspecting the contents of the ImageRef struct, when it should be treated as entirely opaque.

To do handle this, we instead hopefully fix this once and for all by fixing the navigation rules:
- We always inspect anonymous structs
- Otherwise, the field needs the "embed" value in the field tag

Also, adds an integration test which would have caught this behavior in the first place.


nderjung (approved):
Thanks!

PR #79, opened 2026-01-18
feat: Add shell autocompletion

jedevc:
~~:warning: Requires #83.~~

This PR adds autocompletion, resolving https://linear.app/unikraft/issue/CLI-35/command-line-completion.

To use it:

```
❯ unikraft completion
Execute the following command to activate tab completion for unikraft in zsh:

    source <(unikraft completion -c zsh)

Note that this only takes effect for your current shell session. For permanent activation (beyond the current shell session), you can e.g. paste this command into your zsh’s init file, which usually is: ~/.zshrc
```

Tab-complete is supported on:
- Structural elements: command names, flag names
- Enum values
- Instance names in `get`/`list`/`edit`/`delete`
- ...and more later.

---

This is a bit tricky: unlike cobra, kong doesn't support generating command line completions out of the box (see https://github.com/alecthomas/kong/issues/43). There's a few third party libs out there:

- https://github.com/WillAbides/kongplete
  - Uses popular https://github.com/posener/complete library
  - Uses ancient version of kong, appears abandoned
  - Hard-forked into https://github.com/jotaen/kong-completion
    - More up-to-date.
    - I've made a few upstream changes to support our case: https://github.com/jotaen/kong-completion/pulls?q=sort%3Aupdated-desc+is%3Apr+author%3Ajedevc+
- https://github.com/miekg/king
  - Standalone.
  - No docs, limited programatic interface.


internal/resource/cmd/complete.go
```diff
@@ -0,0 +1,60 @@
...
+
+	resources, err := p.resourceType.List(ctx)
+	if err != nil {
+		return []string{}
+	}
+
+	keys := make([]string, 0, len(resources))
+	for _, r := range resources {
+		// XXX: can't return #, they get escaped?
+		// XXX: need a way of returning the *name*, not the UUID
```
jedevc: Argh, @nderjung this is related to key parsing + stringification.

1. `#` doesn't play well as an auto completion character. It gets escaped - which is annoying, this is just a normal shell thing. I *suspect* because it can be used as a comment character. I think we need to pick *again*. I'm wondering about just entirely changing how `image` works, and just going back to `/`.
2. There are multiple different "types" of key we want to be able to return here. e.g. `ls -q` should print the *most* specific output. But completions should be able to show both IDs and names.

nderjung: Again! I was just getting the hang of `#`! 😂 

I always liked `/` but understand the complexity in parsing it w.r.t. images.

What about `+` - it looks like it won't be escaped plus it's quite visually distinct and i can't see it being used in any other contexts? Also reads nicely "fra" and "this UUID"

internal/resource/cmd/complete.go
```diff
@@ -0,0 +1,60 @@
...
+
+type predictResourceKey struct {
+	ctx context.Context
+
+	resourceType resource.GettableResource
+}
+
+func (p predictResourceKey) Predict(a complete.Args) []string {
+	// XXX: need to predict the profile somehow!
+
```
nderjung: Why wouldn't we just select the default profile?

internal/resource/cmd/complete.go
nderjung: If we swapped the `Key`'s `String()` order to prefer `Name` (if its set) over `UUID`, this would make the output a bit more usable:
```diff
@@ -82,16 +82,16 @@ func (k Key) String() string {
        if k.Metro != "" {
                s += k.Metro + MetroKeySeparator
        }
-       if k.UUID != "" {
-               if requiresIDPrefix(k.UUID) {
-                       s += KeyUUIDPrefix
-               }
-               s += k.UUID
-       } else if k.Name != "" {
+       if k.Name != "" {
                if requiresNamePrefix(k.Name) {
                        s += KeyNamePrefix
                }
                s += k.Name
+       } else if k.UUID != "" {
+               if requiresIDPrefix(k.UUID) {
+                       s += KeyUUIDPrefix
+               }
+               s += k.UUID
        }
        return s
 }
```

In a long list of UUIDs, i have no idea which to pick.

nderjung:
As for:

> or we can attempt upstreaming changes.

I support this.

For now, having a gomod replace with a backlink to the fork is generally acceptable but we probably need a policy along the lines of "if the PR is not merged in X time frame we must perpetuate a fork" (aka `x`). Annoyingly this mostly to do with trying to keep dependencies as up-to-date as possible. `kong-completions` looks like it's maintained so I suspect the changes will be merged eventually and we can drop the `replace`; so this won't be a problem.

internal/resource/cmd/complete.go
jedevc: I've been using `unikraft --profile=prod instance ...`. Which is really nice, but the auto-complete in this scenario should auto-complete from the `prod` profile, not from whatever the default one is.

internal/resource/cmd/complete.go
jedevc: We need both :thinking: if a user starts writing a UUID, then we should complete the UUID if they tab on it.

I *think* it's somewhat possible to hide the UUIDs until the user starts typing one.

nderjung: Mmmm good point, yeah it should try both on tab - but by default (if nothing is typed yet) would be good to show names first - not sure if this is possible.

internal/resource/cmd/complete.go
nderjung: omg

nderjung: Sounds like early `os.Args` parsing :S

internal/resource/cmd/complete.go
jedevc: > not sure if this is possible

Yup! Seems to be! Have done this.

jedevc:
> > or we can attempt upstreaming changes.
> 
> I support this.

Excellent :tada: Looks like my patches are already merged, so no need for a fork!

internal/resource/cmd/complete.go
jedevc: Yup :cry: I've added a hack in to do this.

internal/config/config.go
```diff
@@ -30,7 +30,7 @@ type Config struct {
 	Stderr  io.Writer       `kong:"-" yaml:"-" json:"-"`
 
 	// Global configuration.
-	Config    string `group:"flag-global" name:"config" short:"c" help:"Set the configuration file." placeholder:"file" type:"yamlfile" yaml:"-" json:"-"`
+	Config    string `group:"flag-global" name:"config" help:"Set the configuration file." placeholder:"file" type:"yamlfile" yaml:"-" json:"-"`
```
jedevc, line 33: This is intentional - this would otherwise conflict with the `-c` opt available on the `completion` command.

IMO, this is fine though, `--profile` is more used, and there's no shorthand for that.

nderjung (approved):
Awesome. Thanks!

PR #81, opened 2026-01-19
fix(login): Ignore UKC_TOKEN

jedevc:
We can ignore `UKC_TOKEN` - it was only added for testing, and we can just construct our config in the test instead of hijacking that old legacy env var.

This was my "other idea" for https://github.com/unikraft/cli/pull/42.


nderjung (approved):
Thanks!

PR #84, opened 2026-01-20
feat: Add diffs for more instance commands

jedevc:
Before: no output.

After:

image

This creates a new `Differ` function, extracted from how we use `edit`, and apply it to `start`/`stop`/`restart` as well.


nderjung (approved):
This all works for me and looks good. The CI is failing because it's cursed. A rebase + force-push would fix it and show it all passes but think it's not necessary. Cheers!

PR #86, opened 2026-01-22
feat: Add custom (pretty text) log printer

nderjung:
Linear-Closes: CLI-70
Signed-off-by: Alexander Jung <alex@unikraft.com>

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -28,6 +28,7 @@ import (
 
 	"unikraft.com/cli/internal/cmd/login"
 	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/logfmt"
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
 	"unikraft.com/cli/internal/version"
@@ -55,7 +56,13 @@ type UnikraftCLI struct {
 }
 
 func NewRootCmd(ctx context.Context, args []string, stdin io.Reader, stdout, stderr io.Writer) (*kong.Context, *UnikraftCLI, func() error, error) {
-	cli := UnikraftCLI{}
+	cli := UnikraftCLI{
+		Config: config.Config{
+			Stdin:  stdin,
+			Stdout: stdout,
+			Stderr: stderr,
+		},
+	}
 
 	parser, err := NewParser(&cli)
 	if err != nil {
@@ -93,9 +100,6 @@ func NewRootCmd(ctx context.Context, args []string, stdin io.Reader, stdout, std
 	}
 
 	cli.Context = ctx
-	cli.Stdin = stdin
-	cli.Stdout = stdout
-	cli.Stderr = stderr
 
 	var level log.Level
 	switch cli.LogLevel.String() {
@@ -116,7 +120,7 @@ func NewRootCmd(ctx context.Context, args []string, stdin io.Reader, stdout, std
 	default:
 		level = log.InfoLevel
 	}
-	cli.Context = log.WithLogger(cli.Context, log.New(stderr, cli.LogType, level))
+	cli.Context = log.WithLogger(cli.Context, logfmt.New(stderr, cli.LogType, level))
 	cli.Context = ctrdlog.WithLogger(cli.Context, logrus.NewEntry(log.ToLogrus(
 		log.G(cli.Context),
 		log.WithLogrusLevelCap(logrus.DebugLevel),
@@ -167,6 +171,16 @@ func NewParser(cli *UnikraftCLI) (*kong.Kong, error) {
 	globalFlagGroup := kong.Group{
 		Key:   "flag-global",
 		Title: kingkong.Underline("Global flags") + ":",
+	}
+
+	// Replace the global logger getter with our own which leverages our own
+	// log formatter and configuration.
+	log.G = func(ctx context.Context) *log.Logger {
+		if v, ok := ctx.Value(log.ContextKey{}).(*log.Logger); ok {
+			return v
+		}
+
+		return logfmt.New(cli.Stderr, cli.LogType, cli.LogLevel)
 	}
 
 	configFile := filepath.Join(config.ConfigDir(), config.DefaultConfigFilename)
--- a/internal/logfmt/logfmt.go
+++ b/internal/logfmt/logfmt.go
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package logfmt
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/charmbracelet/lipgloss"
+	"github.com/rs/zerolog"
+
+	"unikraft.com/x/colors"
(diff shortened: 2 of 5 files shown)


internal/cmd/root.go
```diff
@@ -130,6 +134,16 @@ func NewParser(cli *UnikraftCLI) (*kong.Kong, error) {
...
 
+	// Replace the global logger getter with our own which leverages our own
+	// log formatter and configuration.
+	log.G = func(ctx context.Context) *log.Logger {
+		if v, ok := ctx.Value(log.ContextKey{}).(*log.Logger); ok {
+			return v
+		}
+
+		return logfmt.New(cli.Stderr, cli.LogType, cli.LogLevel)
+	}
```
jedevc, line 184: Curious what this does - I can remove it, and it all just... seems to keep working.

I think this is because everytime we call `log.G`, we're getting the version that was passed into `WithLogger`. So I don't think this is necessary.

internal/logfmt/logfmt.go
```diff
@@ -0,0 +1,79 @@
...
+// for a short level name.
+var formattedLevels = map[log.Level]string{
+	log.TraceLevel: "T",
+	log.DebugLevel: "D",
+	log.InfoLevel:  "i",
+	log.WarnLevel:  "W",
+	log.ErrorLevel: "E",
+	log.FatalLevel: "!",
+	log.PanicLevel: "X",
+}
```
jedevc: Argh. I see why we could kind of want this - it is sort of nice to be able to show the levels when we get copy-pasted logs. That said, these will show up anytime users pipe output anywhere and `NO_COLOR` gets automatically disabled.

I don't love the letters for this. What about instead using a cute set of unicode characters? Like this:

```suggestion
// FormattedLevels are used by ConsoleWriter's consoleDefaultFormatLevel
// for a short level name.
var formattedLevels = map[log.Level]string{
	log.TraceLevel: "○",
	log.DebugLevel: "◔",
	log.InfoLevel:  "◑",
	log.WarnLevel:  "◕",
	log.ErrorLevel: "●",
	log.FatalLevel: "●",
	log.PanicLevel: "●",
}

var formattedLevelUnknown = "◌"
```

This looks like:

Image

Which IMO is still clear what all the levels are, but doesn't look cluttered if you strip them out.

internal/logfmt/logfmt.go
```diff
@@ -0,0 +1,79 @@
...
+
+var levelColors = map[log.Level]func(str ...string) string{
+	log.TraceLevel: lipgloss.NewStyle().Background(colors.Info).Foreground(colors.Info).Render,
+	log.DebugLevel: lipgloss.NewStyle().Background(colors.Success).Foreground(colors.Success).Render,
+	log.InfoLevel:  lipgloss.NewStyle().Background(colors.Primary).Foreground(colors.Primary).Render,
+	log.WarnLevel:  lipgloss.NewStyle().Background(colors.Warning).Foreground(colors.Warning).Render,
+	log.ErrorLevel: lipgloss.NewStyle().Background(colors.Error).Foreground(colors.Error).Render,
+	log.FatalLevel: lipgloss.NewStyle().Background(colors.Error).Foreground(colors.Error).Render,
+	log.PanicLevel: lipgloss.NewStyle().Background(colors.Error).Foreground(colors.Error).Render,
+}
```
jedevc: These colors are all fine for now, but we should add specifically colors to the x pkg that encodes these eventually.

jedevc (changes requested):

internal/cmd/root.go
nderjung: We need this because if it fails anywhere between this and it being added to the context, the old log formatter is used; i.e. `kong.New` and `parser.Parse`. 

The one caveat is that we also use the `cli` pointer reference; so if it does fail in between this and the context seed, it will only use the default values from the config. Only after `parser.Parse` will the true values be set.

A simple test for you to is to comment this out and run the integration test and see the "empty" gold test fail.

jedevc: :cry: fair. IMO, it would be better to call `WithLogger` twice though, instead of modifying the global. First called with our "general" logger, then we update it once we've successfully parsed the CLI.

nderjung: The reason I didn't do this to begin with is that this just sets a value of `log.G` without evaluating it; and only evaluates it if an error does occur.  If we run `WithLogger`, not only is it evaluated but we add a second logger to the context, holding two structures in memory, both with potentially different configurations.

jedevc: Yeah, I just don't love that we poke into the internals of the logging package to modify globals. It's gonna be rare that two consumers are going to actually be using it, but it just feels like a code smell.

That said, happy to just take this, it's a tiny quibble. I think #72 will end up refactoring lots around this area anyways, so I might end up being able to clean it up.

> but we add a second logger to the context

We can just hold onto the original context - only use the first context to do the kong setup, then just create a new one from the original.

nderjung: That's true. Yeah, i'm happy for you to take over here and see that work touching all of this. Main goal of this PR is just to get `logfmt` in so that we can mess with the printer internally so we have something unique inside of the CLI.

internal/logfmt/logfmt.go
nderjung: I have mixed feelings about this.

One the one hand, it's fun, looks not too bad and the detail of different unicode characters makes it visually interesting.  On the other hand, i have no idea what any of them mean (esp. as a new user).  The letters at least were guessable as to what they meant and made it very clear when someone copypastas what verbosity level was set. Definitely the letters are a bit noisy compared to this.

Not sure yet

jedevc: If we do keep the letters can we at least make them all case consistent?

There's also a few other Unicode things we could do:
- Dots (more dots == more verbosity)
- Filled dots (different shapes are different levels)

Good point about not being able to tell at a glance - but I actually think that's the same with the colors. Given two colors, one can be more urgent than the other, but aside from red == error, orange == warning, it's less clear what the others are. So I don't think they need to be visually identifiable at a glance.

nderjung: i think red/orange (or yellow) are obvious for error-type signals, as you say.

For the rest, we could just use shades of grey for general verbosity (info, debug, trace); for this it doesn't matter too much the color, but you're right that using different colors like green/blue can be misleading. we can save green for success-type operations perhaps. For example:

image
image

FWIW, when it came to using diferent cases, i use `i` because it's more visually distinct.

With all of that said, I am wondering if we should just ditch the idea of a special symbol - are we gaining anything out of it other than potentially noise for the end-user? 

Also worth noting is how it's rendered in different environments. I primarily use two different terminals depending on context. My main driver is Ghostty, and the other is VSCode. VSCode seems to prioritise showing the character in order to maintain legibility, rather then respecting the ascii color escapes. For example, here is VSCode for both the original letter-based logs and then for your suggested circular symbol:

image

image

I also tried with using a common symbol, `│`, across all log-levels and wonder if we we lose anything by doing this.  Not only does it play well with VSCode:

image

But in a `NO_COLOR` environment, it doesn't feel so distracting:

image

wdyt?

nderjung: If this is all acceptable; I've updated the PR with the changes and marked as ready.

jedevc (approved):
Yeah LGTM! We can always change it later as well :tada: It's much better than before!

PR #88, opened 2026-01-23
fix: Common package for string-value manipulation

jedevc:
This patch creates a common `value` package for string <-> value manipulation.
    
We use this functionality for:
- Converting `--set` patches into values
- Parsing the input values for `unikraft run`
- Displaying values in key-value + table output
- Formatting patches for output in `--dry-run`
    
This simplifies and standardizes the logic - the input format is now the same as the output format, and we can handle trickier values (like structs).

For example, before:

image

And, after:

image


nderjung (approved):
Thanks!

PR #89, opened 2026-01-23
fix: Keep certificate timestamps hidden

jedevc:
This is consistent with all other objects that have timestamps included.


nderjung:
I'd argue these are pretty useful fields to be able to see generally when looking at certificates, specifically "not after". It would also really be nice to show this as a relative time as well (NOT AFTER "in 2 weeks").

PR #92, opened 2026-01-23
fix: Ensure prettydiff handles diffs containing newlines

jedevc:
This is needed so we don't weirdly end up splitting multiline diffs in super odd places.

Fixes https://linear.app/unikraft/issue/CLI-54/weird-diff-newline-in-wrong-place.


nderjung (approved):
Thanks!

PR #96, opened 2026-01-23
ci: Use arc runner for all jobs

jedevc:
Linear-Closes: CLI-90


nderjung (approved):
Just so it's noted here as it was discussed offline: the existing `runs-on` is kept in comment as this is a temporary fix and this change will be reverted once the repository goes open-source.

Cheers!

PR #105, opened 2026-01-26
fix: Gracefully handle multiple Create

jedevc:
This is not a regular scenario! But can occur, for example, when starting replicas of a single instance.

Resolves https://linear.app/unikraft/issue/CLI-66/update-creatableresource-to-return-multiple-values.

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -7,9 +7,11 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
+	"golang.org/x/sync/errgroup"
 	"unikraft.com/cli/internal/config"
 	"unikraft.com/cli/internal/logs"
 	"unikraft.com/cli/internal/multimetro"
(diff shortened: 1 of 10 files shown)


internal/cmd/run.go
```diff
@@ -95,25 +97,48 @@ func (c *RunCmd) Run(ctx context.Context, cfg *config.Config, sandbox *resource.
...
 		}
-		_, err = multimetro.DoKeyExact(ctx, cl, instance.key(), func(ctx context.Context, mc *multimetro.MetroClient) (any, error) {
-			r, err := logs.InstanceLogs(ctx, mc).Reader(instance.key().NameOrUUID(), 0, true)
-			if err != nil {
-				return nil, err
+		eg, ctx := errgroup.WithContext(ctx)
+		_, err = multimetro.DoKeys(ctx, cl, keys, func(_ context.Context, mc *multimetro.MetroClient, keys multimetro.Keys) ([]struct{}, []multimetro.Key, error) {
+			for _, key := range keys {
+				eg.Go(func() error {
+					r, err := logs.InstanceLogs(ctx, mc).Reader(key.NameOrUUID(), 0, true)
```
nderjung, line 124: It may be good to support prefixing the logs here to make reading the output of difference instances more distinct since they will now be interlaced.

nderjung (approved):
Thanks!

PR #107, opened 2026-01-26
refactor: Use grouped clients

jedevc:
This PR uses the new grouped clients in https://github.com/unikraft-cloud/go-sdk/pull/104, which are a heavy refactor of the existing multimetro system, entirely decoupled from the CLI.

We extract most of the "Key" functionality into "Ref"s, and convert most of the "Client" functionality into "Group"s. We still keep some of the original types, and use those to wrap the upstream ones.


nderjung (approved):
Thanks!

PR #109, opened 2026-01-27
feat: Parse memory value with IEC-compliant suffix

nderjung:
Linear-Closes: CLI-97


jedevc:
Okay, I've rebased, and also pushed some other changes - specifically, I implemented `encoding.TextMarshaler` and `encoding.TextUnmarshaller` instead of the kong mapper interface.

This means it can also be used to display in `list` and `get`, as well as used when printing patches during `--dry-run`s, setting patches using `--set`, and even during visual patch editing.

Approved, but let me know if those changes make sense before I merge!

jedevc (approved):

PR #113, opened 2026-01-27
fix: Options for run

jedevc:
`--volume` and `--scale-to-zero` both had issues.

`--volume` was just almost entirely broken, the prior implementation was not right.

`--scale-to-zero` took an `enabled` parameter, this actually doesn't exist in the platform, so we shouldn't set it. The correct syntax is `--scale-to-zero policy=on`.

This *also* finally fixes the `scale-to-zero` extra field appearing.


nderjung (approved):
Thanks!

PR #115, opened 2026-01-27
ci(release): Merge the "tag" and "release" workflows

nderjung:
Includes a smaller change that was discovered for this workflow.


jedevc (approved):

.github/workflows/release.yaml
```diff
@@ -55,16 +51,33 @@ jobs:
...
+        git config user.email "monkey@unikraft.com"
+        git remote set-url origin https://x-access-token:${{ secrets.GH_PAT }}@github.com/${{ github.repository }}
+
+    - name: Generate next stable version
+      shell: bash
+      if: ${{ github.ref == 'refs/heads/stable' }}
+      run: |
+        VERSION=$(make next-stable)
+        git tag -a "${VERSION}" -m "${VERSION}"
+        git push -u origin "${VERSION}"
```
jedevc, line 67: This is fine for now, but once we get to a public release, I'd like to set tags manually.

I don't love automatically determining versions if we can avoid it.

nderjung: Right now this allows us to pick when a release to stable happens manually, that's just a merge from staging into stable.  If you feel strongly about it, we can extend this workflow to accept an input tag to replace `make next-stable`.

PR #119, opened 2026-01-28
fix: Update create-only fields for instances

jedevc:
From the demo session earlier, `wait_timeout_ms` shouldn't appear on `unikraft instance get`.

However, there were actually multiple issues here.
- We update certain instance fields to correctly show as "create-only" `invisible`
- We update `invisible` fields to always be filtered out
- We also make sure that `wait-timeout` is used instead, and use a custom parser for durations.


internal/types/duration.go
```diff
@@ -0,0 +1,141 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
```
nderjung: We have successfully made a full solar orbit

nderjung: Also double check your sign-off :)

internal/resource/cmd/printers.go
```diff
@@ -38,11 +38,13 @@ const (
 	PrinterTypeRaw      PrinterType = "raw"
 	PrinterTypeQuiet    PrinterType = "quiet"
 	PrinterTypeTemplate PrinterType = "template"
+
+	PrintTypeDebug PrinterType = "debug"
```
nderjung, line 42: Why is debug lonely? ):

nderjung (approved):
Thanks!

internal/resource/cmd/printers.go
jedevc: It's lonely because it's special :heart: I've also not added this in the docs.

It leaks internal details of the field resources that honestly are not very useful to anyone *except for me*.

PR #120, opened 2026-01-28
ci: Avoid cancelling in-progress workflows

jedevc:
For release: we should *keep* running the already running release.
For test: we should tear down resources at the end of the test run.

I'm hoping this is the cause of https://linear.app/unikraft/issue/CLI-102/leaky-authenticated-cli-test-resources. But will keep an eye on it.


nderjung (approved):
Thanks!

PR #123, opened 2026-01-28
feat: Add log wrapping

jedevc:
Wraps logs to the screen size.

Note. Like the table formatting, this will still look bad when resizing after the command has run (either by resizing the terminal window), or when copy-pasting output. However, for the happy path logs just look better on smaller screen sizes.

Before:

```
│ unikraft CLI arch=amd64 built=unknown commit=uns
et plat=linux version=dev
│ initializing platform clients metros=["ukp-stabl
e","ukp-staging"]
│ getting instances keys=["http-go121-q8yf8"] metr
o=ukp-staging
│ getting instances keys=["http-go121-q8yf8"] metr
o=ukp-stable
│  
│ error:
│   keys not found: [http-go121-q8yf8]
│  
```

After:

```
❯ unikraft instance logs http-go121-q8yf8 --log-level=trace
│ unikraft CLI arch=amd64 built=unknown commit=unset
│ plat=linux version=dev
│ initializing platform clients metros=["ukp-
│ stable","ukp-staging"]
│ getting instances keys=["http-go121-q8yf8"] 
│ metro=ukp-staging
│ getting instances keys=["http-go121-q8yf8"] 
│ metro=ukp-stable
│  
│ error:
│   keys not found: [http-go121-q8yf8]
│  
```

### DIFF
--- a/internal/logfmt/wrapper.go
+++ b/internal/logfmt/wrapper.go
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package logfmt
+
+import (
+	"bytes"
+	"io"
+	"strings"
+
+	"github.com/charmbracelet/x/ansi"
+	"github.com/charmbracelet/x/term"
+	"unikraft.com/x/guesstermwidth"
+)
+
+type wrappedWriter struct {
+	out     io.Writer
+	width   int
+	pending []byte
+}
+
+func newWrappedWriter(out io.Writer, width int) io.Writer {
+	return &wrappedWriter{out: out, width: width}
+}
+
+func newScreenWrappedWriter(out io.Writer) io.Writer {
+	if !isTTY(out) {
+		return out
+	}
+	return newWrappedWriter(out, guesstermwidth.GuessTermWidth(out))
+}
+
+func (w *wrappedWriter) Write(p []byte) (int, error) {
+	if w.width < 1 {
+		return w.out.Write(p)
+	}
+
+	w.pending = append(w.pending, p...)
+	for {
+		newline := bytes.IndexByte(w.pending, '\n')
+		if newline == -1 {
+			break
+		}
+		line := string(w.pending[:newline+1])
+		if newline+1 >= len(w.pending) {
+			w.pending = w.pending[:0]
+		} else {
+			w.pending = w.pending[newline+1:]
+		}
+
+		if err := w.writeWrappedLine(line); err != nil {
+			return 0, err
+		}
+	}
+
+	return len(p), nil
+}
+
+func (w *wrappedWriter) writeWrappedLine(line string) error {
+	prefix, content := splitLogPrefix(line)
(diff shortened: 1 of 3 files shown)


jedevc:
Keeping in draft, so I get another chance to glance at this tomorrow morning and make sure my tired brain is correct.

But ready-for-review.

internal/logfmt/wrapper.go
```diff
@@ -0,0 +1,101 @@
...
+func splitLogPrefix(line string) (string, string) {
+	ss := strings.SplitAfterN(line, " ", 2)
+	if len(ss) < 2 {
+		return "", line
+	}
+	if !strings.Contains(ss[0], LogLevelSymbol) {
+		return "", line
+	}
+	return ss[0], ss[1]
+}
```
nderjung, line 93: This feels brittle (maybe another argument for absorbing zerolog... or `zerolog.ConsoleWriter`?)

Otherwise I think _first_ removing the ascii escapes _and then_ checking if the first character is `LogLevelSymbol` feels a little bit more robust... but still not happy with this approach since it feels like so much post processing.

Is there no way to plug into `zerolog.ConsoleWriter` sans `FormatLevel` ?

In an ideal world, we're wrapping the value of the log and then adding new lines with the `LogLevelSymbol` rather than checking it after the fact.

Even more ideal would be "fancy" processing here, like:

```
│ this is my message
│   key1=value1
│   key2=value2
| this is another message
│   key1=value1
│   key2=value2
```
or

```
│ this is my message
│   key1=value1 key2=value2
| this is another message
│   key1=value1 key2=value2
```
via
```go
log.G(ctx).
  Info().
  Str("key1", "value1").
  Str("key2", "value2").
  Msg("this is my message")
log.G(ctx).
  Info().
  Str("key1", "value1").
  Str("key2", "value2").
  Msg("this is another message")
```
If you think that we should just go ahead with this implementation for now, then that's also ok, we can always open a ticket for handling more sophisticated approach (also considering nested logging, like how @robertgzr mentioned).

jedevc: > Is there no way to plug into `zerolog.ConsoleWriter` sans `FormatLevel` ?

Nope :cry: Each part is formatted independently:

https://github.com/rs/zerolog/blob/a0d61dc2c7439bdea7e1a9920713b1134367be58/console.go#L121-L174

> In an ideal world, we're wrapping the value of the log and then adding new lines with the LogLevelSymbol rather than checking it after the fact.

Agreed, this just doesn't exist right now :cry:

That said, I actually think this approach is fine for now - yes, there's improvement to be done, but I don't think we should take the time *right now* to fork zerolog and add our custom formatting.

> Otherwise I think _first_ removing the ascii escapes _and then_ checking if the first character is `LogLevelSymbol` feels a little bit more robust... but still not happy with this approach since it feels like so much post processing.

I could update this to use `vtclean`, but this does allocate a new string - just checking to see if the prefix character is included is very cheap. I've added a comment - I don't actually think it's actually possible to not get that prefix? It's just kind of a sanity check to prevent very weird cases from appearing oddly (e.g. if somehow a log line has a newline in the middle).

nderjung: Let's leave for now - thanks! Allocating an extra string on a log line is performance sin :)

nderjung (approved):
Thanks!

PR #126, opened 2026-01-29
feat: Allow specifying login info explicitly

jedevc:
Split out from https://github.com/unikraft/cli/pull/104.

The idea is that this can be used in scripts / etc, to manually login where the token is known.

There's also a weird patch that seems to work for registry authentication when we can't find the organization in the config file.

### DIFF
--- a/cmd/unikraft/testdata/TestGolden/auth/help
+++ b/cmd/unikraft/testdata/TestGolden/auth/help
@@ -19,6 +19,10 @@ stdout:
 	          Allow insecure server connections when using SSL.
 	      --no-browser ($UNIKRAFT_NO_BROWSER)
 	          Do not open the browser automatically for login.
+	      --token ($UNIKRAFT_TOKEN)
+	          Path to a file containing the authentication token (or '-' for stdin).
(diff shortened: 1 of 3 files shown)


cmd/unikraft/testdata/TestGolden/auth/help
```diff
@@ -19,6 +19,10 @@ stdout:
 	          Allow insecure server connections when using SSL.
 	      --no-browser ($UNIKRAFT_NO_BROWSER)
 	          Do not open the browser automatically for login.
+	      --token ($UNIKRAFT_TOKEN)
```
nderjung, line 22: I'm wondering for compatibility reasons this should be `UKC_TOKEN` ?

jedevc: This is a path to a file - so actually `UKC_TOKEN` isn't the right one here.

IMO users shouldn't be able to specify this via env var anyways: this should be fixed in [`83a0a61` (#131)](https://github.com/unikraft/cli/pull/131/commits/83a0a615864e87703093f1905eab2e59ebe813f4).

nderjung (approved):
Thanks!

PR #127, opened 2026-01-30
feat(skills): Add generated skills file for the cli

craciunoiuc:
Closes: CLI-138

### DIFF
--- a/skills/unikraft-cli/SKILL.md
+++ b/skills/unikraft-cli/SKILL.md
@@ -1,0 +1,243 @@
+---
+name: unikraft-cli
+description: Manage Unikraft Cloud resources including instances, images, volumes, and services. Use when the user needs to deploy applications, manage cloud infrastructure, or interact with Unikraft Cloud.
+allowed-tools: Bash(unikraft*)
+---
+
+# Unikraft Cloud CLI
+
+The `unikraft` command provides a unified interface for managing resources on Unikraft Cloud. It enables deploying instances, managing persistent storage, and configuring services.
+
+## Installation
+
+```bash
+curl --proto '=https' --tlsv1.2 -fsSL https://unikraft.com/cli/install.sh | sh
+```
+
+## Quick Start
+
+```bash
+# Start by logging into Unikraft Cloud
+unikraft login --no-browser
+
+# Run a simple NGINX instance
+unikraft run --metro=fra --scale-to-zero policy=on,stateful=true,cooldown-time=10 -p 8080:80 nginx:latest
+
+# List running instances
+unikraft instances list
+```
+
+## Core Resources
+
+- **Instances**: MicroVMs on Unikraft Cloud run from `Dockerfile`'s.
+- **Images**: Unikernel images stored in the registry.
+- **Volumes**: Persistent storage that can be attached to instances.
+- **Services**: Networking abstractions for exposing instances.
+- **Certificates**: TLS certificates for secure connections.
+- **Metros**: Geographic locations where resources are deployed.
+
+## Commands
+
+### Building (`unikraft build`)
+
+The `build` command compiles a project from a Kraftfile into a unikernel image.
+
+```bash
+unikraft build [flags] [<path>]
+```
+
+**Common Flags:**
+
+- `-o, --output <dest>`: Output destination (registry tag or local OCI archive path).
+- `--build-arg <key=val>`: Set build-time variables.
+- `--no-cache`: Do not use cache when building the image.
+- `--secret <spec>`: Secret to expose to the build (format: `id=mysecret[,src=/local/secret]`).
+- `--ssh <spec>`: SSH agent socket or keys to expose to the build (format: `default|<id>[=<socket>|<key>[,<key>]]`).
+
+### Deployment (`unikraft run`)
+
+The `run` command is the primary entry point for deploying applications.
+
+```bash
+unikraft run [flags] <image> [<args>...]
+```
+
+**Common Flags:**
+
+- `--metro <code`>: Metro to deploy in (e.g., `fra`, `dal`, `sin`).
+- `-p, --publish <port>`: Publish a port (e.g., `443:8080/http+tls`).
+- `-e, --env <key=val>`: Set environment variables.
+- `-v, --volume <vol>`: Attach a volume.
+- `-m, --memory <size>`: Set memory size (e.g., `512MiB`).
+- `--scale-to-zero`: Enable scale-to-zero policies.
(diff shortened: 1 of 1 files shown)


jedevc:
Can you add instructions to `README.md` for how to use this?

craciunoiuc:
I think they are added automatically to https://skills.sh/

for example:
```
npx skills add https://github.com/anthropics/skills --skill frontend-design
```

we could add this I think:
```
npx skills add https://github.com/unikraft/cli
```

or this if we move it to a `cli` directory:
```
npx skills add https://github.com/unikraft/cli --skill cli
```

Does it sound good?

skills/unikraft-cli/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+name: unikraft-cli
+description: Manage Unikraft Cloud resources including instances, images, volumes, and services. Use when the user needs to deploy applications, manage cloud infrastructure, or interact with Unikraft Cloud.
+allowed-tools: Bash(unikraft*)
+---
+
+# Unikraft Cloud CLI
+
+The `unikraft` command provides a unified interface for managing resources on Unikraft Cloud. It enables deploying instances, managing persistent storage, and configuring services.
+
+## Installation
```
nderjung, line 11: I would reduce "installation" to the "official procedure" once we have one based on the `curl` command (PR incoming).

The build (`task cli`) i think should be moved to the bottom under "Hacking" or "Building from source" to be explicit - an agent should also be able to understand that difference.

Separately, we may wish to include the per-distro mechanisms here. For now we only have debian/ubuntu.

skills/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+# The binary is placed in dist/unikraft
+
+# Or install via script (general usage)
+curl -sL https://get.unikraft.io | bash
+```
+
+## Quick Start
+
+```bash
+# Login to Unikraft Cloud
```
nderjung:
```suggestion
# Start by logging into Unikraft Cloud
```
That said, as an agent, they may not be able to use `unikraft login` as it opens a browser. So as an agent, they may need to use `unikraft login --no-browser` or `unikraft login --token`. Or maybe an env var.

skills/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+unikraft login
+
+# Run a simple NGINX instance
+unikraft run --metro=fra -p 8080:80 nginx:latest
```
nderjung: Add scale-to-zero props as a default value-add.

skills/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+- **Instances**: Virtual machines running Unikernel images.
+- **Images**: Unikernel images stored in the registry.
+- **Volumes**: Persistent storage that can be attached to instances.
+- **Services**: Networking abstractions for exposing instances.
+- **Certificates**: TLS certificates for secure connections.
+- **Metros**: Geographic locations where resources are deployed.
+
+## Commands
+
+### Deployment (`run`)
```
nderjung:
```suggestion
### Deployment (`unikraft run`)
```

skills/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+unikraft run --metro=fra --follow my-app:latest
+
+# Get detailed info
+unikraft instances inspect my-instance-name
+```
+
+## Troubleshooting
+
+**Authentication Issues?**
+Run `unikraft login` to refresh credentials. Check if your token is expired.
```
nderjung: What does "Check if your token is expired." mean? There is no way to do this or to systematically check it (as an agent).

skills/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+# Run a simple NGINX instance
+unikraft run --metro=fra -p 8080:80 nginx:latest
+
+# List running instances
+unikraft instances list
+```
+
+## Core Resources
+
+- **Instances**: Virtual machines running Unikernel images.
```
nderjung:
```suggestion
- **Instances**: microVMs on Unikraft Cloud run from `Dockerfile`'s
```

skills/SKILL.md
craciunoiuc: Remember I AI generated this 😅

skills/SKILL.md
```diff
@@ -0,0 +1,163 @@
...
+---
+
+# Unikraft Cloud CLI
+
+The `unikraft` command provides a unified interface for managing resources on Unikraft Cloud. It enables deploying instances, managing persistent storage, and configuring services.
+
+## Installation
+
+```bash
+curl -sL https://get.unikraft.io | sh
```
jedevc: This doesn't exist.

skills/unikraft-cli/SKILL.md
```diff
@@ -0,0 +1,163 @@
...
+Run `unikraft login` to refresh credentials.
+
+**Deployment Failures?**
+Use `--dry-run` to validate configuration.
+Check `unikraft instances logs <id>` for application startup errors.
+
+**Resource Not Found?**
+Ensure you are targeting the correct metro with `--metro`. Resources are often metro-specific.
+
+## Tasks (Developer)
```
jedevc, line 236: This probably doesn't belong here. I think this is probably for AGENTS.md?

jedevc (commented):
LGTM, nits only.

skills/SKILL.md
craciunoiuc: Yes, we need to find something to replace it with (for when we release this)

skills/unikraft-cli/SKILL.md
craciunoiuc: No idea, it was present in other skills files, see here https://skills.sh/browser-use/browser-use/browser-use

skills/SKILL.md
jedevc: @craciunoiuc `install.sh` should be working now. Can you update this PR?

nderjung:
```
curl --proto '=https' --tlsv1.2 -fsSL https://unikraft.com/cli/install.sh | sh
```

nderjung (commented):
I think we just need a build example as one last item

craciunoiuc:
No idea how that looks ;_;

nderjung:
> No idea how that looks ;_;

https://github.com/unikraft/cli/blob/staging/internal/cmd/build.go#L33-L66

craciunoiuc:
Ok added that and whatever else was missing

nderjung (approved):
Thanks!

PR #132, opened 2026-02-02
feat: Allow virtually patching the instance state

jedevc:
This isn't *actually* patchable using the PATCH method, but, we can issue a start/stop command instead. This is pretty neat!

Closes https://linear.app/unikraft/issue/CLI-127/add-virtual-patch-for-instance-state.


nderjung (approved):
That's fun!

This works for me, approving ahead-of-time so you can rebase and merge.

One thing I noticed during usability is that it might be good to show in the visual editing comments possible values for enums (i.e. so a new user knows to write `running` if the instance has stopped or that they can only run `running` and not `standby`).

PR #133, opened 2026-02-02
fix: Make template behave like docker templates

jedevc:
And like kubectl templates. Instead of requiring the user to iterate / index over all the results themselves, the template should be applied to each resource independently.

Fixes https://linear.app/unikraft/issue/CLI-142/template-for-inspect-should-accept-one-list-should-accept-multiple.


nderjung (approved):
Thanks!

PR #137, opened 2026-02-03
feat: Add build

jedevc:
Fixes https://linear.app/unikraft/issue/CLI-21/implement-unikraft-build.

Depends on:
- https://github.com/unikraft-cloud/x/pull/143
- https://github.com/unikraft-cloud/x/pull/138


internal/cmd/build.go
```diff
@@ -0,0 +1,98 @@
...
+
+type BuildCmd struct {
+	Input  string `arg:"" default:"." help:"Path to the input directory."`
+	Output string `short:"o" help:"Output destination"`
+
+	// similar to docker compose build
+	BuildArg []string `help:"Set build-time variables."`
+	NoCache  bool     `help:"Do not use cache when building the image."`
+	Secret   []string `help:"Secret to expose to the build (format: \"id=mysecret[,src=/local/secret]\")."`
+	SSH      []string `help:"SSH agent socket or keys to expose to the build (format: \"default|<id>[=<socket>|<key>[,<key>]]\")."`
```
nderjung, line 28: ~~Maybe to make the CLI flag more explicit `--ssh-agent` ? I just wonder if `--ssh` is a bit generic and might assume that ssh is available during the build~~

Nevermind `docker build --ssh` :)

internal/builder/rootfs.go
```diff
@@ -0,0 +1,281 @@
...
+				Type:      client.ExporterLocal,
+				OutputDir: localDest,
+			},
+		},
+		LocalDirs:     localDirs,
+		Frontend:      "dockerfile.v0",
+		FrontendAttrs: attrs,
+	}
+
+	pw, err := progresswriter.NewPrinter(context.WithoutCancel(ctx), os.Stderr, "auto")
```
nderjung, line 90: Is it possible to leverage `progressui.NewDisplay` instead and pump the logs through our log system? Something like `log.G(ctx).With().Logger().Level(log.InfoLevel)` ?

jedevc: Absolutely. We can and should integrate this a bit better. I've put this under https://linear.app/unikraft/issue/TOOL-617/unikraft-build-followups, I don't want to block getting some functionality in, since the progress ui is definitely a *pain* from the buildkit point of view.

internal/cmd/build.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (c *BuildCmd) Run(ctx context.Context, cfg *config.Config) error {
```
nderjung, line 66: Think we should add some `Examples()` :)

Similarly, the top-level `README.md` should also be updated with one or two examples too

jedevc: Done!

internal/cmd/build.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	defer func() {
+		for _, img := range imgs {
+			img.Close()
+		}
+	}()
+
+	if c.Output == "" {
+		return nil
+	}
```
nderjung, line 118: If `c.Output` is empty, essentially the output is discarded. The action of building simply verifies that the build is possible; without any artefact.  So:

1. Should the `-o|--output` flag be mandatory field? Or,
2. Should there be a default output artefact?
3. Keep-as is..? what's the reason for the current approach?

jedevc: I would like to keep as-is, that's kind of the behavior of docker build if you don't specify a tag with `-t` (maybe more recent versions just invent a random name, and put it in the image store).

The main reason to not output it, is - what should we output? What should we name it, where should we put it? We don't have an image store. We probably shouldn't push it to `index.unikraft.io` without a name, especially given that that's a slow operation.

I don't like the idea of a default output artifact really, but maybe `--output` could be mandatory? I do like being able to check that the build works though.

nderjung (approved):
Thanks!

PR #139, opened 2026-02-04
test: Fix leaky certificates from TestGolden/services/edit

jedevc:
Fixes https://linear.app/unikraft/issue/CLI-102/leaky-authenticated-cli-test-resources.

This was occurring because editing a service can update the linked resource, by creating a new one. We need to recognize that this can happen, and scan for new resources after editing, adding any new resources to the sandbox.


nderjung (approved):
Thanks!

PR #150, opened 2026-02-16
ci(release): Capture relevant version output from stdout

nderjung:
The version outputs in the gather versions step gets mangled with the prompt/"assume yes" from the remote taskfile.  Luckily, [the prompt occurs only in the stderr][0] so if we filter this out, we'll get clean values.

[0]: https://github.com/go-task/task/blob/6dedcafd7dbc05aa233e1e2a6e5b5d75a50454a8/internal/input/input.go#L63

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -62,16 +62,16 @@ jobs:
       id: versions
       run: |
         {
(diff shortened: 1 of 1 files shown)


jedevc (approved):

PR #153, opened 2026-02-17
feat: Add ability to delete all resources

jedevc:
Fixes https://linear.app/unikraft/issue/CLI-154/implement-unikraft-purge.

There are two parts:
- Added a unified `unikraft resource` command which allows interacting with *all* resources. e.g. `unikraft resource list` to list all resources.
- Added `--all` and `--filter` flags for the `remove` cmd. This is implemented as a *variant* of `remove`, because not all deleteable resources are listable (e.g. images for when we do that).

Then, from both of these features, we get a `unikraft resource rm --all` (essentially what `unikraft purge` would do).

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -304,10 +304,10 @@ func NewParser(cli *UnikraftCLI) (*kong.Kong, error) {
 }
 
 var SandboxedResources = []resource.Resource{
-	Certificate{},
 	Instance{},
-	ServiceGroup{},
 	Volume{},
+	ServiceGroup{},
+	Certificate{},
 }
 
 type staticKey string
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -53,6 +53,8 @@ type UnikraftCLI struct {
 	Certificates CertificatesCmd `cmd:"" help:"Manage Unikraft Cloud certificates." aliases:"certificate,certificates,crt,crts,cert,certs"`
 	Images       ImagesCmd       `cmd:"" help:"Manage Unikraft Cloud images." aliases:"image,images,img,imgs"`
 
+	Resources AnyResourceCmd `cmd:"" hidden:"" help:"Manage any Unikraft Cloud resource." aliases:"resource,resources"`
+
 	SendAnalytics SendAnalyticsCmd `cmd:"" help:"Send analytics payload (used internally for detached analytics)." name:"_send_analytics" hidden:""`
 }
 
@@ -118,6 +120,7 @@ func NewRootCmd(ctx context.Context, args []string, stdio config.Stdio) (context
 		kongcompletion.WithPredictor("resource-key-service", cmd.PredictResourceKey[ServiceGroup](ctx)),
 		kongcompletion.WithPredictor("resource-key-certificate", cmd.PredictResourceKey[Certificate](ctx)),
 		kongcompletion.WithPredictor("resource-key-image", cmd.PredictResourceKey[ImageEntry](ctx)),
+		kongcompletion.WithPredictor("resource-key-resource", cmd.PredictResourceKey[AnyResource](ctx)),
 	)
 
 	kctx, err := parser.Parse(args)
--- a/internal/resource/cmd/cmd.go
+++ b/internal/resource/cmd/cmd.go
@@ -6,6 +6,7 @@
 package cmd
 
 import (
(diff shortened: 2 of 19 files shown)


jedevc:
FYI @craciunoiuc. Hope this seems reasonable for you.

craciunoiuc:
Beautiful that you remembered to do this :D

internal/resource/cmd/cmd.go
```diff
@@ -565,3 +572,102 @@ func (cmd *ResourceCreateCmd[R]) Run(ctx context.Context, stdio config.Stdio, sa
...
+
+	err = cmd.Output.
+		WithDefault(PrinterTypeTable).
+		Print(stdio.Stdout, cmd.Field, empty, resources...)
+	if err != nil {
+		return err
+	}
+
+	if !cmd.Force {
+		fmt.Fprintf(stdio.Stdout, "\nType \"YES\" (in capitals) to confirm deletion: ")
```
craciunoiuc: can we make it accept `y/yes` and all capitalization variations?

internal/cmd/root.go
```diff
@@ -52,6 +52,8 @@ type UnikraftCLI struct {
 	Services     ServicesCmd     `cmd:"" help:"Manage Unikraft Cloud services." aliases:"service,services,svc,svcs"`
 	Certificates CertificatesCmd `cmd:"" help:"Manage Unikraft Cloud certificates." aliases:"certificate,certificates,crt,crts,cert,certs"`
 	Images       ImagesCmd       `cmd:"" help:"Manage Unikraft Cloud images." aliases:"image,images,img,imgs"`
+
+	Resources AnyResourceCmd `cmd:"" help:"Manage any Unikraft Cloud resource." aliases:"resource,resources"`
```
craciunoiuc: I unfortunately find it hard to understand what is the difference between a `resource` or already a command like `instance` for example. Can you elaborate please?

internal/cmd/root.go
```diff
@@ -293,10 +293,10 @@ func NewParser(cli *UnikraftCLI) (*kong.Kong, error) {
 }
 
 var SandboxedResources = []resource.Resource{
-	Certificate{},
+	Volume{},
 	Instance{},
 	ServiceGroup{},
-	Volume{},
+	Certificate{},
```
craciunoiuc: why not order alphabetically :D

craciunoiuc (commented):
Some comments my side

internal/cmd/root.go
jedevc: The idea is that a `resource` is *all* the resources supported in cloud.

So, for example you could do this to get *all* the objects in your account, across all metros:

```
❯ unikraft resource ls
TYPE      KEY
volume    volume:ukp-stable/vol-sde0a
volume    volume:ukp-stable/test-434d46374d49
volume    volume:ukp-stable/vol-dvj8x
instance  instance:ukp-stable/www-ux7k3
service   service:ukp-stable/red-haze-5hz0l1hq
```

Without this, you'd have to individually work with each item type, and stitch the results together.

One of the reasons this was added is so we can have a global remove - so you just need to do `unikraft resource rm --all` - this deletes everything. But something like `unikraft instance rm --all` would only delete all instances.

internal/resource/cmd/cmd.go
jedevc: Yup! Done.

internal/cmd/root.go
jedevc: Reordered these. But they're not alphabetic: they need to be the order we can safely delete in.

e.g. we can't delete a certificate before we delete the service group that refers to it. 

Ideally, we should build a directed dependency graph using the links (that are kind of sort of working), but this is much more complicated.

craciunoiuc: Ok, makes sense, sounds good

internal/cmd/root.go
craciunoiuc: Ooooh, ok, interesting, nice

craciunoiuc (approved):
All good here. Thanks!

nderjung:
As discussed offline, just waiting for the `hidden` go tag so it can be internal for now as we gather feedback.

nderjung:
Last commit is missing sign-off

nderjung (approved):
Thanks!

PR #154, opened 2026-02-17
feat: Add metro connection status

jedevc:
Fixes https://linear.app/unikraft/issue/CLI-130/metro-connection-info.

image

### DIFF
--- a/internal/cmd/metros.go
+++ b/internal/cmd/metros.go
@@ -7,10 +7,17 @@ package cmd
 
 import (
 	"context"
+	"net"
+	"net/http"
+	"net/url"
+	"time"
 
 	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/httpclient"
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
+	"unikraft.com/cli/internal/types"
+	"unikraft.com/cli/internal/xsync"
 	"unikraft.com/x/kingkong"
 )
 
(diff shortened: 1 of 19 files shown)


internal/cmd/metros.go
```diff
@@ -43,7 +50,142 @@ func (i Metro) Raw() any {
...
-	return resource.FieldsFromStruct(i)
+	fields, err := resource.FieldsFromStruct(i)
+	if err != nil {
+		return nil, err
+	}
+
+	baseClient := httpclient.DefaultHTTPClient
+	if i.Insecure {
+		baseClient = httpclient.InsecureHTTPClient
+	}
```
craciunoiuc: Could this be somehow abstracted away? I'm assuming these same lines happen everywhere

internal/cmd/metros.go
```diff
@@ -43,7 +50,142 @@ func (i Metro) Raw() any {
...
+		if err != nil {
+			return "", nil
+		}
+		conn.Close()
+		return types.PingLatency(elapsed), nil
+	})
+
+	online := xsync.OnceCtxValues(func(ctx context.Context) (any, error) {
+		client := &http.Client{
+			Timeout:   3 * time.Second,
```
craciunoiuc: should the timeout be configurable?

internal/cmd/metros.go
```diff
@@ -43,7 +50,142 @@ func (i Metro) Raw() any {
...
+		if err != nil {
+			return nil, err
+		}
+		resp, err := client.Do(req)
+		if err != nil {
+			return types.MetroStatusOffline, nil
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode >= 200 && resp.StatusCode < 500 {
```
craciunoiuc, line 158: Looks ok, I wonder if we can use the constants for these :D

internal/cmd/metros.go
craciunoiuc: Also, maybe we can make it a constant somewhere instead of `3 * second` everywhere

internal/cmd/metros.go
```diff
@@ -82,6 +224,10 @@ func (Metro) Examples() map[cmd.CmdType][]kingkong.Example {
 				Description: "List all metros",
 				Commands:    []string{"unikraft metro list"},
 			},
+			{
+				Description: "List metros with connection status",
+				Commands:    []string{"unikraft metro list -f +status"},
```
craciunoiuc, line 229: This `+` thing looks weird, I guess this is used also in more places?

internal/cmd/metros.go
```diff
@@ -43,7 +50,142 @@ func (i Metro) Raw() any {
...
+
+		ips, err := resolveIPs(ctx)
+		if err != nil || len(ips) == 0 {
+			return "", nil
+		}
+
+		addr := net.JoinHostPort(ips[0], port)
+		start := time.Now()
+		conn, err := (&net.Dialer{Timeout: 3 * time.Second}).DialContext(ctx, "tcp", addr)
+		elapsed := time.Since(start)
```
craciunoiuc, line 132: I see `dial tcp` I guess you tested and times are similar to what you would get using `ping` on the server?

craciunoiuc (commented):
some food for thought :D

internal/cmd/metros.go
jedevc: Right out of kraftkit 🎉 I will check tho

internal/cmd/metros.go
jedevc: Made it a constant!

internal/cmd/metros.go
jedevc: Abstracted :)

internal/cmd/metros.go
jedevc: We can, I don't personally love it though. HTTP status codes are well known, and it looks weird to do `resp.StatusCode < StatusInternalServerError`.

internal/cmd/metros.go
jedevc: Yeah, this is the syntax for adding more fields to the output. If you just to `-f status`, you get *only* status. Alternative suggestions welcome (for a separate PR).

internal/cmd/metros.go
craciunoiuc: true, makes sense

internal/cmd/metros.go
craciunoiuc: good question, only other idea I have is to have a separate `--extra` flag, but that's not clean either

craciunoiuc: But if I see `-f`  as `--fields` the `+` sign makes sense, instead of `--filter` :D

Though I understand the need to keep it uniform everywhere

internal/cmd/metros.go
```diff
@@ -43,7 +50,142 @@ func (i Metro) Raw() any {
...
+		client := &http.Client{
+			Timeout:   timeout,
+			Transport: baseClient.Transport,
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+		}
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, i.Endpoint, nil)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
jedevc: Wrong, this is deliberate. If we fail to create the request, then this is legitimate.

internal/cmd/metros.go
jedevc: Actually this is such a good question. As it turns out `api.fra.unikraft.cloud` drops ICMP packets, so `ping` doesn't give me *anything*.

nderjung: Yeah ICMP ping is disabled by default on metros by design.

nderjung: There is [`/healthz`](https://unikraft.com/docs/api/platform/v1/node#health-check) - but unauthenticated, it won't provide any details other than "online" (aka 401 unauthorized, which actually indicates it's online) (fine for now, maybe in future we provide some more details like `healthy`, `upgrading`, `degraded`, etc.)

nderjung:
Fixed the other CI issues, tests now report change in CLI output adding the extra field

internal/cmd/metros.go
jedevc: I might be missing it?

```
❯ curl https://api.fra.unikraft.cloud/healthz -H "Authorization: Bearer $UKC_TOKEN"
{"status":"error","message":"No API endpoint","errors":[{"status":404}]}
```

jedevc:
@nderjung should be good now!

craciunoiuc (approved):
All good here. Thanks!

PR #152, opened 2026-02-17
fix: Preserve yaml metadata on config save

jedevc:
Fixes https://linear.app/unikraft/issue/TOOL-592/streamline-quick-metroprofile-switching.

Mostly, this preserves comments. Whitespace isn't fully preserved sadly, this isn't preserved by the parser.

As reported by @jschlumpp here: https://discord.com/channels/879723098881028167/1448722316782735463/1473248133801709691.

PR #157, opened 2026-02-18
feat: Add config interpolation

jedevc:
:warning: Depends on #152.

:warning: Needs de-sloppification.


jedevc:
Closed, we're doing https://github.com/unikraft/cli/pull/232 instead.

PR #158, opened 2026-02-19
fix: Show partial output before error

jedevc:
:hammer_and_wrench: Fixes https://linear.app/unikraft/issue/TOOL-587/errors-in-list-inspect-should-still-show-un-errored-items.

If we error out, we should still show *some* output.


jedevc:
@nderjung this PR also includes changes that prevent evaluating lazy fields *multiple* times.

So, e.g. if you run `unikraft metro ls --profile=prod -f +status --log-level=trace --sort status.ping --filter "status.online==online"`, we should only ping the server *once*. Without this, we actually do it 3 times!

nderjung (approved):
Tests are failing due to #189 which is already merged. No rebased needed. Thanks!

PR #169, opened 2026-02-23
fix(cmd/root): Set default path when initializing config

nderjung:
This PR fixes an issue where on a fresh installation, the configuration object is empty.  If the object (which is bound to subcommands) is saved, it will return an error due to a missing config path.  Since the config path has been already determined in the sequence before this, we can simply set the path.


craciunoiuc (approved):
All good here. Thanks!

PR #180, opened 2026-03-02
feat: Add flags for sorting asc/desc resource items

nderjung:
Add a `--sort` flag for resource listings that supports:
- Prefix notation for direction: -field (descending), +field or field (ascending);
- Multiple comma-separated fields for multi-level sorting; and,
- Mixed ascending/descending in a single sort specification.

Examples:
```
  --sort name                   # ascending by name
  --sort -name                  # descending by name
  --sort state,timing.uptime    # by state, then uptime
  --sort state,-timing.uptime   # ascending state, descending uptime
```

This mirrors the `--field` selection syntax and enables expressive sorting of resource output.

### DIFF
--- a/internal/resource/cmd/cmd.go
+++ b/internal/resource/cmd/cmd.go
@@ -127,9 +127,10 @@ type FormatOpts struct {
 }
 
 type ResourceListCmd[R resource.GettableListableResource] struct {
-	Name   []string       `arg:"" optional:"" completion-predictor:"resource-key-${name}" help:"Names of the ${names} to list."`
-	Filter []string       `help:"Filter output based on a field value (e.g. --filter state==running)." sep:"none"`
-	Watch  *time.Duration `short:"w" help:"Watch for changes and refresh output." type:"optional"`
+	Name   []string            `arg:"" optional:"" completion-predictor:"resource-key-${name}" help:"Names of the ${names} to list."`
+	Filter []string            `help:"Filter output based on a field value (e.g. --filter state==running)." sep:"none"`
+	Watch  *time.Duration      `short:"w" help:"Watch for changes and refresh output." type:"optional"`
+	Sort   xkong.GreedyStrings `help:"Sort output by field values (e.g. --sort name,-timestamps.created-at). Use - prefix for descending, + for ascending."`
 
 	FormatOpts
 
@@ -162,6 +163,12 @@ func (cmd *ResourceListCmd[R]) Run(ctx context.Context, stdio config.Stdio, sand
 
 	ctx = resource.WithFilter(ctx, filter)
 
+	// Parse sort configuration
+	sortSpecs, err := parseSortSpecs(cmd.Sort...)
+	if err != nil {
+		return err
+	}
+
 	var empty R
 	render := func(out io.Writer) error {
 		var resources []resource.Resource
@@ -179,6 +186,12 @@ func (cmd *ResourceListCmd[R]) Run(ctx context.Context, stdio config.Stdio, sand
 		resources, err = filterResources(ctx, resources, filter)
 		if err != nil {
 			return err
+		}
+		if len(sortSpecs) > 0 {
+			resources, err = sortResources(ctx, resources, sortSpecs)
+			if err != nil {
+				return err
+			}
 		}
 		return cmd.Output.
 			WithDefault(PrinterTypeTable).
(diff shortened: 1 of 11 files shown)


internal/resource/cmd/cmd.go
```diff
@@ -352,6 +375,64 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
...
+			if err != nil {
+				return err
+			}
+
+			matched := resource.GetFieldByPath(fields, path)
+			if len(matched) == 1 {
+				items[i].value = matched[0].Value
+			}
+			return nil
+		})
```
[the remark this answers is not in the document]
nderjung: Not necessary in latest version of Go. (forvar: copying variable is unneeded)

internal/resource/cmd/cmd.go
```diff
@@ -115,9 +117,11 @@ type FormatOpts struct {
...
 
 type ResourceListCmd[R resource.GettableListableResource] struct {
-	Name   []string       `arg:"" optional:"" completion-predictor:"resource-key-${name}" help:"Names of the ${names} to list."`
-	Filter []string       `help:"Filter output based on a field value (e.g. --filter state==running)." sep:"none"`
-	Watch  *time.Duration `short:"w" help:"Watch for changes and refresh output." type:"optional"`
+	Name     []string       `arg:"" optional:"" completion-predictor:"resource-key-${name}" help:"Names of the ${names} to list."`
+	Filter   []string       `help:"Filter output based on a field value (e.g. --filter state==running)." sep:"none"`
+	Watch    *time.Duration `short:"w" help:"Watch for changes and refresh output." type:"optional"`
+	SortAsc  string         `help:"Sort output by field value in ascending order (e.g. --sort-asc name)."`
+	SortDesc string         `help:"Sort output by field value in descending order (e.g. --sort-desc timestamps.created-at)."`
```
jedevc: I would like it if `--sort` "just worked". A few options, in order of preference:
- We only have `--sort`. Then to do a descending sort you just do `--sort -<field>` to reverse it. To do an explicit ascending sort, you'd do `--sort +<field>`. This kind of mimics how `--field` selection works which is nice.
- We have `--sort`/`--sort-asc` (as aliases) and `--sort-desc`.

I also think we should support *multiple fields* to sort by. That way you could do `--sort state,timing.uptime` which would sort by state *then* for equivalent states would sort by uptime. One of the reasons I like doing the `-` prefix for descending is that then we can support combining ascending and descending operations. e.g. `--sort state,-timing.uptime` to sort by ascending state, and then by descending uptime, which is fun.

internal/resource/cmd/cmd.go
```diff
@@ -141,6 +145,19 @@ func (cmd *ResourceListCmd[R]) Run(ctx context.Context, stdio config.Stdio, sand
 	}
 	ctx = resource.WithFilter(ctx, filter)
 
+	// Determine sort configuration
+	var sortPath resource.FieldPath
+	var sortDesc bool
+	if cmd.SortAsc != "" && cmd.SortDesc != "" {
+		return fmt.Errorf("cannot use both --sort-asc and --sort-desc at the same time")
+	}
```
jedevc: If we do have separate flags (see above), we can use `xor` groups in kong instead of doing this check ourselves.

internal/resource/cmd/cmd.go
nderjung: Addressed in https://github.com/unikraft/cli/commit/ce46e4b3136f2bb2f2431537339a096d3151e96d

nderjung: Note: this has the caveat of not being able to use the flag to sort descending by itself because it will be interpreted as a short flag, i.e.:
```
unikraft instance ls --sort -name
```
it has to be used like so:
```
unikraft instance ls --sort=-name
```

nderjung: A workaround could be to use an alternative character, I suggest `^` but is confusing when used with `+`.

jedevc: Ah, grr. Yeah, I think this is a kong limitation. I need to take a look at this.

For *now* can we use `-`? If we can't fix this in kong, we need a fix for `-f` as well. At least then it's consistent.

jedevc: Okay we can use `GreedyString`/`GreedyStrings` from https://github.com/unikraft/cli/pull/186.

nderjung: I've rebased the work to use the greedy strings :) Thanks!

jedevc (approved):
Neat!

PR #182, opened 2026-03-02
docs: Restore info on manual profile creation

jedevc:
This was deleted as part of https://github.com/unikraft/cli/pull/170.

We should keep this info. I was linking internally to folks who had manually deployed metros. The file *is* designed to be human readable + editable, so I think it's fine to mention here.


nderjung (approved):
Ooops, yes, thanks!

PR #184, opened 2026-03-02
feat: New `--dangling` flag for `unikraft image list`

jedevc:
We weren't really allowing list dangling images in any way. This now adds a `--dangling` flag that lets the user do this.

Also, some metros (staging) are producing some odd output. These are wrong, and need fixing, but we shouldn't completely fail out on this, we just should discard the digest.


nderjung (approved):
Thanks!

PR #186, opened 2026-03-03
fix: Allow removing fields using -f -field-name

jedevc:
Due to a kong "issue" (intentional design), this was previously only achievable by `--field=-field-name`. Now it's possible, even if the next component *looks* like a flag.


internal/resource/cmd/cmd.go
```diff
@@ -120,9 +121,7 @@ func (cmd ResourceCmd[R]) Examples() []kingkong.Example {
 }
 
 type FormatOpts struct {
-	// FIXME: not able to pass values beginning with -
```
jedevc, line 123: One down. Many to go 😢

nderjung (approved):
Yay, thanks!

PR #187, opened 2026-03-03
fix: Ensure cooldown-time has correct casing

jedevc:
The automatic naming for unpacking values wasn't quite right. We need to explicitly use `strcase.ToKebab` to match the automagic field naming.

This manifested like this:

    $ unikraft run --metro=fra0-bench --scale-to-zero policy=on,cooldown-time=300 nginx:latest
    │
    │ error:
    │   unknown fields: [cooldown-time]

### DIFF
--- a/internal/resource/value/parse.go
+++ b/internal/resource/value/parse.go
@@ -12,6 +12,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/ettle/strcase"
+
 	xmaps "unikraft.com/cli/internal/x/maps"
 )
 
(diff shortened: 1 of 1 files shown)


internal/resource/value/parse.go
```diff
@@ -157,7 +159,12 @@ func parseReflect(input []string, value reflect.Value) error {
 					if !field.IsExported() {
 						continue
 					}
-					if strings.SplitN(field.Tag.Get("field"), ",", 2)[0] == k || strings.EqualFold(field.Name, k) {
+					name := strings.SplitN(field.Tag.Get("field"), ",", 2)[0]
```
nderjung: `strings.Cut` ?

jedevc: @copilot let it be so.

jedevc: Well no that's awful, I don't like that at all.

nderjung (approved):
Thanks!

PR #193, opened 2026-03-06
ci: Use public runners

jedevc:
For all jobs, except `test`s where we want to connect to `ukp-stable`.


nderjung (approved):
Thanks!

PR #194, opened 2026-03-06
feat: Homebrew, "`install.sh`" and upgrade subcmd

nderjung:


internal/cmd/upgrade.go
```diff
@@ -0,0 +1,490 @@
...
+	"unikraft.com/cli/internal/version"
+)
+
+// ErrChecksumNotAvailable indicates the checksum file could not be fetched.
+var ErrChecksumNotAvailable = errors.New("checksum file not available")
+
+type UpgradeCmd struct {
+	Channel string `help:"Release channel to upgrade from." default:"stable" enum:"stable,staging"`
+	Force   bool   `short:"f" help:"Force upgrade even if already at latest version."`
+	Use     string `short:"v" name:"use" help:"Upgrade to a specific version."`
```
jedevc: `--use` or `-v`? Why not `--version`?

internal/cmd/upgrade.go
```diff
@@ -0,0 +1,490 @@
...
+	return nil
+}
+
+// compareSemver compares two semantic version strings.
+// Returns:
+//
+//	-1 if a < b
+//	 0 if a == b
+//	 1 if a > b
+func compareSemver(a, b string) (int, error) {
```
jedevc: https://pkg.go.dev/golang.org/x/mod/semver pls 🙏

internal/cmd/upgrade.go
nderjung: because we have `--version` is a top-level and kong panics as it already exists

jedevc (approved):
LGTM :tada:

PR #196, opened 2026-03-11
feat: Allow setting instance vsock

jedevc:
This patch introduces support for enabling vsock on created instances.

For example:

```bash
unikraft instance create \
  --set metro=fra0-bench \
  --set image=demo/nginx:latest \
   --set name=jedevc-test \
   --set service.services=8080:8080/tls \
   --set vsock=true
```

While working on this, I realized that `vsock` was appearing in visual edits, despite the fact that there's no backing state. This was weird, so I started hacking on https://linear.app/unikraft/issue/TOOL-570/better-behavior-for-visual-editing-and-create-only-fields as well. This now gives much better behavior - a user can add `vsock` fields to the file they're editing, but they're not there by default, because we can't obtain this value from the API.

This has a few other cool benefits:
- `unikraft instance create -e` just *works*, instead of failing complaining about required fields. We defer validating the required fields, so we can handle this a bit better.
- `unikraft instance edit <name> --cmd "yq ..."` allows editing using an external command. The content is added to stdin, and what comes out of stdout is applied.
- `unikraft instance edit <name> --save <path>` + `unikraft instance edit <name> --load <path>` allows editing the resource more easily. This is quite useful for LLMs.

Also this fixes a bug where `unikraft instance edit --del runtime.env=A,B` (note, two keys) would panic weirdly.


nderjung (approved):
Thanks!

PR #199, opened 2026-03-13
fix: Consistent profile order

jedevc:
The order of the profiles was previously dependent on map-iteration order, which is non-deterministic.


nderjung (approved):
Thanks!

PR #200, opened 2026-03-13
fix: Correctly show equal lines

jedevc:
Observed in the real world with 4 env vars, `A`, `B`, `C` and `D` - then I deleted `A` and `D`.

This created this weird diff, which I've now fixed and added a test for.


nderjung (approved):
Thanks!

PR #201, opened 2026-03-13
ci: Make task lint ignore default disables

jedevc:
We don't want those default disables, especially since our CI checks against the standard config. This can result in inconsistencies.


nderjung (approved):
Thanks!

PR #203, opened 2026-03-13
gomod: Remove replace directives

jedevc:
Now it can be `go install`ed :)


nderjung (approved):
Thanks!

PR #205, opened 2026-03-16
feat: Support unikraft.io images

jedevc:
`unikraft.io` and `index.unikraft.io` both point to the same place.

We should prefer `unikraft.io`.


nderjung (approved):
Thanks!

PR #209, opened 2026-03-18
fix: Ensure generic resource type continues on errors

jedevc:
An error on get/list/delete shouldn't be a reason to show no output. The underlying resources handle this neatly, so should we.


craciunoiuc (approved):
All good here. Thanks!

PR #212, opened 2026-03-19
feat!: Add shortcut flags for patches

jedevc:
For `create` and `edit` commands, we can easily support some shortcut flags.

There is one major breaking change - `unikraft run` now takes an explicit `--image` flag, instead of an arg, which is so that `run` and `instance create` are essentially the *same* underlying code.

### DIFF
--- a/internal/cmd/certificates.go
+++ b/internal/cmd/certificates.go
@@ -11,6 +11,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/alecthomas/kong"
+
 	"unikraft.com/cloud/sdk/platform"
 	"unikraft.com/cloud/sdk/platform/group"
 	"unikraft.com/x/kingkong"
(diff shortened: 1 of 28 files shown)


internal/cmd/volumes.go
```diff
@@ -30,19 +32,59 @@ import (
...
+// VolumeCreateCmd extends the generic resource create command with shortcut
+// flags for commonly used volume fields. Each field tagged with
+// `shortcut:"<path>"` is translated into a --set <path>=<value> entry before
+// the standard create pipeline runs.
+type VolumeCreateCmd struct {
+	cmd.ResourceCreateCmd[Volume]
+
+	Metro string              `group:"flag-create" shortcut:"metro" help:"Metro to create in." placeholder:"metro" example:"fra,sfo,nyc"`
+	Name  string              `group:"flag-create" shortcut:"name" short:"n" help:"Volume name." placeholder:"name"`
+	Size  types.SizeMebibytes `group:"flag-create" shortcut:"size" help:"Volume size." placeholder:"size" example:"10GiB,100GiB"`
```
craciunoiuc:
```suggestion
	Size  types.SizeMebibytes `group:"flag-create" shortcut:"size" help:"Volume size." placeholder:"size" example:"10GiB,100MiB"`
```

craciunoiuc: Same in other places, maybe good to give two unit sizes for better udnerstanding that it's not fixed?

internal/cmd/certificates.go
```diff
@@ -294,12 +304,18 @@ func (Certificate) Examples() map[cmd.CmdType][]kingkong.Example {
...
 					`unikraft certificate create \
-  --set name=demo-cert \
-  --set cn=demo.unikraft.dev. \
-  --set-file chain=cert.pem \
-  --set-file pkey=cert.key \
-  --set metro=fra`,
+	  --name demo-cert \
+	  --cn demo.unikraft.dev. \
+	  --chain "$(cat cert.pem)" \
+	  --pkey "$(cat cert.key)" \
```
craciunoiuc: tbh this feels a bit dirty, I know we had this in `kraft` but I never really liked passing private keys as arguments 😓

craciunoiuc (changes requested):
some comments my side

internal/cmd/certificates.go
jedevc: I changed these to be paths to files.

internal/cmd/volumes.go
jedevc: Done.

craciunoiuc (approved):
As one great Marc would say: YOLO!

PR #213, opened 2026-03-19
feat: Add config resource

jedevc:
Example: 
image

The idea is also to support doing editing this from the command line, but things are in flux there: see #196.


nderjung:
```
$ ./dist/unikraft config inspect profile
│
│ error:
│   config file not found: profile
│
```

jedevc:
Just do `unikraft config inspect` to get the current *full* config. What did you expect the command you ran to do?

I *am* thinking about having a `unikraft config profile ...` command (instead of `unikraft profile`).

nderjung:
> Just do `unikraft config inspect` to get the current _full_ config. What did you expect the command you ran to do?

Based on the help menu, `unikraft config get [<name> ...] [flags]`, i was expecting to be able to retrieve elements of the config, so these should have worked:
- `unikraft config get path`
- `unikraft config get profile`
- `unikraft config get profiles[0].name` (this one is fake, but you get the gist, i can target elements inside the config).

> I _am_ thinking about having a `unikraft config profile ...` command (instead of `unikraft profile`).

Yeah that makes sense

nderjung (approved):
Thanks!

PR #216, opened 2026-03-20
feat(build): Cache pulled artifacts

jedevc:
If we're pulling rootfs or kernel from an OCI registry, then we should probably cache it locally. There's no real reason that we should try and pull it from the harbor registry over and over again.

It's all content hashed anyways.

### DIFF
--- a/internal/x/imagespec/cache.go
+++ b/internal/x/imagespec/cache.go
@@ -1,0 +1,132 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package imagespec
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"sync"
+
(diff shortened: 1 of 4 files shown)


internal/x/imagespec/cache.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/containerd/errdefs"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	spec "unikraft.com/x/image-spec"
+	"unikraft.com/x/log"
+)
+
+// WrapCached wraps a File with caching support. If the file has a backing
+// provider and descriptor, it returns a new ContentStoreFile that uses a
+// pull-through cache at ~/.cache/unikraft. Otherwise, returns the original file.
+func WrapCached(ctx context.Context, file spec.File) spec.File {
```
jedevc, line 25: Hm, we should try and *only* cache if it's fetched from a remote registry. If it's a *local* content store, then probably not.

internal/x/imagespec/cache.go
jedevc, line 1: Is it okay to have no automatic cache cleanup? Probably not.

PR #221, opened 2026-03-22
ci: Disambiguate `UNIKRAFT_CONFIG` env var

jedevc:
Don't use it for *both* the config and the path.


nderjung (approved):
The fact that it's failing is correct behaviour (it's failing for a separate reason, which is something we were trying to catch, which we couldn't see before).

Thanks!

PR #223, opened 2026-03-23
feat(internal)!: Add timeouts to all supported endpoints

jedevc:
Set timeouts in:
- `start`
- `delete`
- `volimport`
- instance template creation
The rest already had timeouts.

Also remove some hacks while refactoring.

Depends on: https://github.com/unikraft-cloud/platform/pull/937

Closes: TOOL-1134

### DIFF
--- a/cmd/unikraft/integration/instance_test.go
+++ b/cmd/unikraft/integration/instance_test.go
@@ -959,6 +959,25 @@ func TestInstances(t *testing.T) {
 		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
 	})
 
+	t.Run("create-waits-for-running", func(t *testing.T) {
+		r := runner(t, true, []string{staging, stable})
+		instName := uniq()
+
+		r.Run(t, []string{
+			"unikraft", "instance", "create",
+			"--output", "quiet",
+			"--set", "name=test-" + instName,
+			"--set", "metro=" + r.Config.MetroName,
+			"--set", "image=nginx:latest",
+			"--set", "autostart=true",
+			"--set", "resources.memory=128",
+			"--set", "resources.vcpus=1",
+		})
+		out := r.Run(t, []string{"unikraft", "instance", "inspect", "test-" + instName})
+		assert.Regexp(t, `state:\s+running`, out)
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
+	})
+
 	t.Run("suspend", func(t *testing.T) {
 		r := runner(t, true, []string{staging, stable})
 		instName := uniq()
--- a/cmd/unikraft/integration/instance_test.go
+++ b/cmd/unikraft/integration/instance_test.go
@@ -86,7 +86,10 @@ func tunnelProxyUUIDs(t *testing.T, cfg *config.Config, metro string) map[string
 
 func TestInstances(t *testing.T) {
 	t.Run("create", func(t *testing.T) {
-		r := runner(t, true, []string{staging, stable, prod})
+		// TODO: Add 'prod' back when it runs platform version 13. Older
+		// versions send a duplicate "status" member that breaks every wait.
+		// See https://github.com/unikraft-cloud/platform/pull/937
+		r := runner(t, true, []string{staging, stable})
 		instName := uniq()
 
 		out := r.Run(t, []string{"unikraft", "instance", "list", "--output", "quiet"})
@@ -244,7 +247,10 @@ func TestInstances(t *testing.T) {
 	})
 
 	t.Run("create-oom", func(t *testing.T) {
-		r := runner(t, true, []string{staging, stable})
+		// TODO: Add 'stable' back when it runs platform version 13. Older
+		// versions send a duplicate "status" member that breaks every wait.
+		// See https://github.com/unikraft-cloud/platform/pull/937
+		r := runner(t, true, []string{staging})
 		instName := uniq()
 
 		r.Run(t, []string{
@@ -266,7 +272,10 @@ func TestInstances(t *testing.T) {
 	})
 
 	t.Run("connect", func(t *testing.T) {
-		r := runner(t, true, []string{staging, stable, prod})
+		// TODO: Add 'prod' back when it runs platform version 13. Older
+		// versions send a duplicate "status" member that breaks every wait.
+		// See https://github.com/unikraft-cloud/platform/pull/937
+		r := runner(t, true, []string{staging, stable})
 		instName := uniq()
 		domainName := uniq()
 
@@ -300,7 +309,10 @@ func TestInstances(t *testing.T) {
 	})
 
 	t.Run("getting-started", func(t *testing.T) {
-		r := runner(t, true, []string{staging, stable, prod})
+		// TODO: Add 'prod' back when it runs platform version 13. Older
+		// versions send a duplicate "status" member that breaks every wait.
+		// See https://github.com/unikraft-cloud/platform/pull/937
+		r := runner(t, true, []string{staging, stable})
 		instName := uniq()
 
 		r.Run(t, []string{
@@ -421,7 +433,10 @@ func TestInstances(t *testing.T) {
 	})
 
 	t.Run("restart-follow", func(t *testing.T) {
-		r := runner(t, true, []string{staging, stable})
+		// TODO: Add 'stable' back when it runs platform version 13. Older
+		// versions send a duplicate "status" member that breaks every wait.
+		// See https://github.com/unikraft-cloud/platform/pull/937
+		r := runner(t, true, []string{staging})
 		instName := uniq()
 		volName := uniq()
 		baseImage := integ.Busybox.Build(t, r)
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -988,7 +988,11 @@ func instancePatchSpec(path string, op patchOp, value any) (platform.MutableInst
 		}
 		return platform.MutableInstancePropertyAnnotations, value.(map[string]string), nil
 	case "image":
-		return platform.MutableInstancePropertyImage, value.(types.ImageRef[reference.Named]).Reference.String(), nil
+		ref := value.(types.ImageRef[reference.Named]).Reference
+		if ref == nil {
+			return zero, nil, fmt.Errorf("image cannot be empty")
+		}
+		return platform.MutableInstancePropertyImage, ref.String(), nil
 	case "runtime.args":
 		return platform.MutableInstancePropertyArgs, []string(value.(InstanceArgs)), nil
 	case "runtime.env":
@@ -1098,7 +1102,9 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 		case "metro":
 			metro = string(field.Create.Set.(LinkName[Metro]))
 		case "image":
-			imageURL = field.Create.Set.(types.ImageRef[reference.Named]).Reference.String()
+			if ref := field.Create.Set.(types.ImageRef[reference.Named]).Reference; ref != nil {
+				imageURL = ref.String()
+			}
 		case "pull-policy":
 			pullPolicy = field.Create.Set.(*platform.PullPolicy)
 		case "type":
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -41,6 +41,7 @@ import (
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
 	"unikraft.com/cli/internal/resource/value"
+	"unikraft.com/cli/internal/timeouts"
 	"unikraft.com/cli/internal/tunnel"
 	"unikraft.com/cli/internal/types"
 )
@@ -152,7 +153,6 @@ type Instance struct {
 	SchedPriority *platform.SchedPriority `mirror:"instance.sched_priority" field:"sched-priority,long" create:"set" edit:"set" flag:"sched-priority" help:"Scheduling priority for the instance." placeholder:"priority" example:"normal,medium,high,admin"`
 	Autostart     bool                    `field:"autostart,invisible,valueless" create:"set" flag:"autostart" help:"Start instance automatically."`
 	Replicas      int64                   `field:"replicas,invisible,valueless" create:"set" flag:"replicas" help:"Number of replicas." placeholder:"n" example:"1,3"`
-	WaitTimeout   types.DurationS         `field:"wait-timeout,invisible,valueless" create:"set"`
 	Features      []string                `field:"features,invisible,valueless" create:"set" flag:"feature" sep:"none" help:"Instance feature." placeholder:"feature"`
 	Vsock         bool                    `field:"vsock,invisible,valueless" create:"set" edit:"set"`
 	Template      string                  `field:"template,invisible,valueless" create:"set" flag:"template" help:"Create from instance template." placeholder:"name"`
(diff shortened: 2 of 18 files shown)


internal/cmd/instances.go
```diff
@@ -594,7 +593,15 @@ func (Instance) Delete(ctx context.Context, targets []resource.Resource) error {
 	}
 	return group.DoRefs(ctx, g, keys.Refs(), func(ctx context.Context, c multimetro.MetroClient, refs group.Refs) (group.Refs, error) {
 		log.G(ctx).Trace().Msg("deleting instances")
-		instances, err := c.DeleteInstances(ctx, refs.NameOrUUIDs(), platform.DeleteInstancesOpts{})
+		reqs := make([]platform.DeleteInstanceRequestItem, 0, len(refs))
+		for _, ref := range refs.NameOrUUIDs() {
+			reqs = append(reqs, platform.DeleteInstanceRequestItem{
+				Name:     ref.Name,
+				Uuid:     ref.Uuid,
+				TimeoutS: ptr.ToPtr(int64(-1)),
```
jedevc: This has max 10 seconds. So for *all* cases where we set this, we need to fallback to a Wait request as well, to extend that.

craciunoiuc:
I think we need to keep the sleeps so I'll drop the commit (in kraft we had also a 1s sleep after the creation commands or so)

For the delete commit this was already done and merged so it can be dropped

Next I'm doing the loop retry for 5 minutes

internal/cmd/instances.go
```diff
@@ -1469,14 +1504,226 @@ func (args *StopOpts) toReq(nameOrUUID platform.NameOrUUID) platform.StopInstanc
...
+							UUID:  inst.Uuid,
+						})
+					default:
+						return nil, fmt.Errorf("instance %q transitioned to unexpected state %q while waiting to start", inst.Uuid, inst.State)
+					}
+				}
+				mu.Lock()
+				nextPending = append(nextPending, stillPending...)
+				mu.Unlock()
+			}
```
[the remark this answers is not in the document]
craciunoiuc: return early and error the wait sounds good

internal/cmd/instances.go
```diff
@@ -1469,14 +1504,227 @@ func (args *StopOpts) toReq(nameOrUUID platform.NameOrUUID) platform.StopInstanc
...
+			_, waitErr := c.WaitInstances(ctx, waitReqs)
+			if waitErr == nil {
+				return refs, nil
+			}
+
+			if time.Now().After(deadline) {
+				return nil, fmt.Errorf("timed out waiting for instance(s) to start: %w", waitErr)
+			}
+
+			getResp, getErr := c.GetInstances(ctx, refs.NameOrUUIDs(), platform.GetInstancesOpts{})
```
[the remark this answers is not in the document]
craciunoiuc: yes, set to false

internal/cmd/instances.go
```diff
@@ -1246,7 +1269,8 @@ func (cmd *InstancesLogsCmd) Run(ctx context.Context, stdio config.Stdio) error
 }
 
 type InstancesStartCmd struct {
-	Targets []string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instances to start."`
+	Targets []string      `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instances to start."`
+	Wait    time.Duration `help:"Time to wait for instances to reach running state." default:"5m"`
```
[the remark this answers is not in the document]
craciunoiuc: golden files will be updated later

internal/cmd/instances.go
```diff
@@ -1469,14 +1504,227 @@ func (args *StopOpts) toReq(nameOrUUID platform.NameOrUUID) platform.StopInstanc
...
+
+			getResp, getErr := c.GetInstances(ctx, refs.NameOrUUIDs(), platform.GetInstancesOpts{})
+			if getErr != nil {
+				return nil, getErr
+			}
+			var stillPending multimetro.Keys
+			for _, inst := range getResp.Data.Instances {
+				if inst.Status != nil && *inst.Status != platform.ResponseStatusSuccess {
+					continue
+				}
```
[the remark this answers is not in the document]
craciunoiuc: treat as error

craciunoiuc:
@jedevc  I don't really have an answer to how to cleanly configure the wait time for delete and create

They are very abstract (and kinda weird to me)

In kraft what I did is to not wait for `autostart` on create as the create operation does not really make sense to be blocking. The fact that the instance starts automatically does not alter the idea that the command just creates an instance.

Similar for the delete operation. I also have my doubts with that. It's fine to wait for the max API time for it to go away, but the operation is anyway async, so why should we make it synchronous? But if we do want to do it, then why don't we do it for all API operations, theoretically we can.

jedevc:
It's useful to make it synchronous for scripting and similar. If you delete an instance, but don't wait, and then try and delete a volume that was mounted you can get "still mounted" as an error. We should wait, so that when it returns it's all guaranteed to be definitely gone and all it's resources disposed of.

jedevc:
It would be useful to have some sort of option to skip this synchronizing tbf - but imo it does make sense to synchronize as a default.

jedevc:
Maybe this is more complex though. And needs some sort of visual indicator while it's waiting to start but after the instance has been created.

craciunoiuc:
Ah sorry, didn't think you'd get/answer to the notification :sweat: 

The option is right now I believe to set the `wait` to `0`, so all good (I still need to do some tweaks)

> And needs some sort of visual indicator while it's waiting to start but after the instance has been created.

Agreed, maybe some spinning thingy like the one in kraft that did `propagating` but in this case waiting for this to start

craciunoiuc:
The good part though is that 90% of cases finish in less than 1 second, and 99.9% of cases finish in less than 10s

Above that it's safe to say that te rest will take less than 20s

craciunoiuc:
Keeping in draft until box is fixed, but can be reviewed @jedevc

jedevc:
@craciunoiuc can you revive this one plsss :heart: 

It'll fix https://linear.app/unikraft/issue/TOOL-1134/vm-template-create-can-return-references-not-found.

jedevc:
As discussed: if a call fails because of a timeout, just *warn*, and keep going.

craciunoiuc:
Ok we have a problem here then

I revisited this and saw the comment

If we try to continue it means we'll have to rearch how our usage of commands works. For example: when a create endpoint call errors out with a `timeout`, it will not provide its instance information and it will just return the error message

This means that in order to "continue" we need to do a Get when the call times out and try to backfill data on the create response which feels ugly?

Anyway, have a rebase:

jedevc:
Hmmmm, I thought a timeout would still return a partial success with some info. Does it not? If not, that does feel like a thing that the platform API should be updated to support.

craciunoiuc:
ok, so there is a partial success, I'm picking again up to add this

jedevc:
Be aware of unikraft-cloud/cli#400. Which adds some partial success stuff.

craciunoiuc:
draft till I know 100% it works

craciunoiuc:
CI is being annoying, I'm investigating, but note that, stable will fail on that test till updated.

I will take stable/prod out of it until bumped there also

internal/cmd/instances.go
```diff
@@ -1174,14 +1173,13 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
-			autostart := field.Create.Set.(bool)
+			autostart = field.Create.Set.(bool)
 			req.Autostart = &autostart
 		case "replicas":
 			replicas := field.Create.Set.(int64)
 			req.Replicas = &replicas
 		case "wait-timeout":
-			timeout := field.Create.Set.(types.DurationS)
-			req.TimeoutS = new(int64(timeout))
+			waitTimeout = new(int64(field.Create.Set.(types.DurationS)))
```
jedevc: Let's remove this field (and similar `wait-timeout` fields anywhere else).

It shouldn't be configurable, it should always wait as long as possible.

internal/cmd/instances.go
```diff
@@ -1242,34 +1254,95 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 	if err != nil {
 		return nil, err
 	}
+	// Instances as the create call reported them, keyed by multimetro key. Used
+	// as a fallback when the follow-up Get cannot see them yet.
+	createdData := make(map[string]createdResource)
```
jedevc: Can this ever happen? This is weird, we created something, but `Get` can't see it??? Even if that's an async operation going on, that feels weird from the API surface point-of-view.

It makes for a lot of complexity, when really, IMO, once something is returned as "it exists", the API should be externally consistent.

If this code path is actually being exercised, I think we need to report this to the platform team.

internal/cmd/instances.go
```diff
@@ -1242,34 +1254,95 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+					reqRetry.WaitTimeoutMs = new(int64(-1)) //nolint:staticcheck // Required for compatibility with metros that do not support timeout_s.
+				} else {
+					waitTimeoutMs := timeoutS * 1000
+					reqRetry.WaitTimeoutMs = &waitTimeoutMs //nolint:staticcheck // Required for compatibility with metros that do not support timeout_s.
+				}
+			} else {
+				reqRetry.WaitTimeoutMs = new(int64(-1)) //nolint:staticcheck // Required for compatibility with metros that do not support timeout_s.
+			}
+			resp, err = c.CreateInstance(ctx, reqRetry)
+		}
```
jedevc: Let's make this a helper func.

internal/cmd/instances.go
```diff
@@ -1242,34 +1254,95 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+			// A UUID means the instance exists, whatever the per-item status
+			// says.
+			if instance.Uuid == "" && failed {
+				message := ptr.ZeroIfNil(instance.Message)
+				if message == "" {
+					message = "unknown error"
+				}
+				errs = append(errs, fmt.Errorf("instance create failed for %s: %s", name, message))
+				continue
+			}
```
jedevc, line 1397: I wonder if we could simplify a lot of this flow by having a reusable `reflect`-based function something like `resp, err := removeTimeoutErrors(resp, err)`.

Then we iterate over everything in the response + rebuild the error, removing the timeouts, so we can warn nicely, but then continue on as normal.

internal/cmd/instances.go
jedevc, line 1987: Wondering if we should set `drain_timeout_ms`? Or `force`? Or whatever.

Not really in scope here, but maybe worth thinking about, since `start` is now updated too.

internal/cmd/instances.go
craciunoiuc: Sure, I believe this field was hidden anyway

internal/cmd/instances.go
craciunoiuc: Yes it can happen, I tested a case here: https://discord.com/channels/879723098881028167/1534550849299284129/1534897084212187290

A few examples:
* You create an instances with `delete-on-stop` and it crashes/finishes on boot (deleted before GET)
* You start an async operation that finishes after he timeout and GET
* You accidentally delete the instance you created before the GET happened (hard but doable)
* You used autokill and it got autokilled before the GET happened
and so on

I think there are plenty of cases where this can happen, resulting in the original bug report `TOOL-1134`.

I would really like to keep this fix (or an alternative to it) in.

internal/cmd/instances.go
craciunoiuc: Hmm, sounds interesting

Will need to check, hopefully it actually reduces complexity :D

internal/cmd/instances.go
craciunoiuc: Hmm true, but I guess it would change functionality from `default stop` to:
* force stop if timeout is 0
* drain connections if timeout >0 or -1

This would change a bit how it works though. So maybe we leave force out of it for now and we only work around drain? Worth having a discussion on

internal/cmd/instances.go
craciunoiuc: Added some helpers with some magic mirror reflection

Looks cleaner I think? :eyes:

internal/cmd/instances.go
jedevc: URGH you're not wrong. Okay fine, do we just get the UUID/name in the Get or is it the full object? Like, could we *always* just return that?

If not, I guess there's not much we can do, and we need this ugly hack.

craciunoiuc: I think you meant in the `Create` not `Get`?

Yeah, you get the uuid/name but also some other info for some reason lol. I sent on discord but have here also:
```
metro:     dev-ccraciunoiu1
name:      nginx-57jo6
uuid:      98be47fc-b7d7-443b-aadb-4cafc6a84bb8
state:     starting
service:
  name:    muddy-grass-3jjl1whm
  uuid:    75f77131-753b-43a1-9646-5ea3684ec356
  domains:
  - fqdn:  muddy-grass-3jjl1whm.dev-ccraciunoiu1.apw.unikraft.internal
```
  
So also the service information and the state

jedevc: But no memory or vcpus 🤔 weird. Yeah, fine I guess we can have this fallback 😞

craciunoiuc:
rebased

main change is that it now applies to instance checkpoints (which do not have a fallback btw)

internal/cmd/instances.go
```diff
@@ -1258,6 +1256,13 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 		return nil, fmt.Errorf("only one of --image, --template, --branch, or --checkpoint may be specified")
 	}
 
+	// The API only waits for an instance that it also starts, and rejects
+	// timeout_s outright when autostart is unset. -1 makes the metro wait as
+	// long as its own configured API timeout allows.
```
jedevc: AI comments.

craciunoiuc: Massacred them

craciunoiuc:
Will revisit after we decide if we should actually remove the hacks now

craciunoiuc:
Killed comments, did a round which found some small quirk things, fixed CI

Should be good for a review but no rush

You might dislike still the things that do the retry in place, but at least now not many things will call them

craciunoiuc:
I kept some comments that I thought to be informative

craciunoiuc:
Note: the whole `utils.go` and `utils_test.go` and where they are used are there just for the hacky retry

Without it this change would be like ~400 lines I think

craciunoiuc:
you can do a round now @jedevc

cmd/unikraft/integration/instance_test.go
```diff
@@ -62,7 +62,9 @@ func tunnelProxyUUIDs(t *testing.T, cfg *config.Config, metro string) map[string
 
 func TestInstances(t *testing.T) {
 	t.Run("create", func(t *testing.T) {
-		r := runner(t, true, []string{staging, stable, prod})
+		// NOTE(craciuniuc): Enable back prod once "status" duplication
+		// is solved (node is updated).
```
jedevc: 1. I don't like named `NOTE`s, this isn't just for you.
2. Add enough detail, linking to a bug report, pr, etc, so that anyone in the team would be able to do this task.

internal/cmd/instances_test.go
jedevc, line 1: Useless tests?

internal/cmd/util.go
```diff
@@ -84,6 +89,129 @@ func matchRef(refs group.Refs, name, uuid string) *group.Ref {
 	return nil
 }
 
+// timeoutMessage is the message the platform puts at the top level of a
+// response when a wait ran out.
+const timeoutMessage = "Operation timed out"
+
+// withoutTimeout clears the error from a response that only failed because its
+// wait ran out, warning instead.
+func withoutTimeout[T any](ctx context.Context, resp *platform.Response[T], err error, msg string) (bool, error) {
```
jedevc: Description and name are misleading. It doesn't modify the response.

internal/cmd/util.go
```diff
@@ -84,6 +89,129 @@ func matchRef(refs group.Refs, name, uuid string) *group.Ref {
...
+
+// recoverCreated builds resources for refs that a listing could not find but
+// that a create call already reported. Refs with nothing to fall back on are
+// returned unchanged.
+func recoverCreated[R resource.Resource](
+	ctx context.Context,
+	refs group.Refs,
+	createdData map[string]createdResource,
+	load func(*group.Ref, platform.Instance, *config.Metro, *config.Profile) (R, error),
+) ([]resource.Resource, group.Refs) {
```
jedevc, line 104: Confused. Why is this somewhat generic, but all hardcoded for `platform.Instance` anyways?

internal/volimport/volimport.go
```diff
@@ -103,10 +121,17 @@ func Terminate(ctx context.Context, c multimetro.MetroClient, instUUID string) e
...
-	_, err := c.WaitInstanceByUUID(ctx, instUUID, platform.WaitInstanceByUUIDRequestBody{
-		State:     state,
-		TimeoutMs: new(stopTimeoutMs),
-	})
+	req := platform.WaitInstanceByUUIDRequestBody{
+		State:    state,
+		TimeoutS: new(int64(-1)),
+	}
+	_, err := c.WaitInstanceByUUID(ctx, instUUID, req)
+	if timeoutSUnsupported(err) {
```
jedevc: Why not use the same helpers? These are duplicated.

cmd/unikraft/integration/instance_test.go
craciunoiuc: 1. Ah, I always thought the idea is to write that to show to original comment writer/noter
2. Sure

internal/cmd/util.go
craciunoiuc: hmm, true, missed that, nice catch

craciunoiuc:
ready for another round @jedevc Addressed your comments

I moved the timeout functions to a separate shared package and made them generic enough

cmd/unikraft/integration/instance_test.go
jedevc: 1. `git blame` :heart:

jedevc:
Rebased it. LGTM!

```
```

craciunoiuc:
end of an era :0

PR #224, opened 2026-03-23
feat: Determine default metro from profile

jedevc:
So it needs to be manually specified *less* on create.

The best mechanism for doing with was allowing `context.Context` on the `Fields` calls.


nderjung (approved):
Thanks!

PR #225, opened 2026-03-23
fix: Reorder top-level help

jedevc:
Which should make it much more readable.


nderjung (approved):
Thanks!

PR #230, opened 2026-03-23
fix: Handle images on staging correctly

jedevc:
Split out of https://github.com/unikraft/cli/pull/208, which isn't ready because of the reasons noted.


nderjung (approved):
Thanks!

PR #235, opened 2026-03-27
feat: Add more resource links and handle misc fallout

jedevc:
Fix https://linear.app/unikraft/issue/TOOL-674/introduce-more-links-between-resources.

Add more links through the TUI - which makes it nice and easy and fun to click through.

Since links are also used for automatically building the sandbox so that we can properly tear  *everything* down when we're done, this also requires a few weird tweaks here.


craciunoiuc (approved):
All good here. Thanks!

PR #241, opened 2026-04-01
fix: Use bulk methods instead of per-uuid methods

jedevc:
Since this doesn't always work when using keys - since the key might not contain it, it's what the user directly specified.

This was a regression (I think) introduced in https://github.com/unikraft/cli/pull/191.


craciunoiuc (approved):
All good here. Thanks!

PR #243, opened 2026-04-02
feat: Add volume filesystem field

jedevc:
We *should* be doing this via https://github.com/unikraft-cloud/proto/pull/235, but we can't update the go sdk for reasons: https://github.com/unikraft-cloud/go-sdk/pull/144#discussion_r3000624088.

But `AdditionalProperties` gives us a nice little hack that we can do here anyways!


procub3r (approved):
10 / 10 no notes. thanks!

nderjung (approved):
Thanks!

PR #245, opened 2026-04-03
feat: Add 'volume import' command

craciunoiuc:
This adds initial support for volume importing. As before, only CPIO files are supported for importing due to how the "backend" of this is built.

~To add to this, only files and directories are supported right now. I could not figure out how to correctly split the `dockerfile` implementation to make it stop before creating the OCI archive using `imagespec`, as I needed it to give me a `.cpio` file.~
Dockerfiles also work now.

~To make this fully work, `proto` needs to be updated and the `delete_on_stop` flag needs to be renamed to `delete-on-stop`.~
~Switched to new way of providing features that works.~
Sdk fixed.

~Finally, I don't really like how things sit but I couldn't find a better way to break things apart whilst also keeping the general repo structure (aka not refactoring everything 😈)~
I did the refactoring hehe.

~tldr: if you give it no `--source` it will use the workdir and assume it's a Kraftfile in there (or fail if not)~
Removed.

If you give it a `--source` it uses that, be it a CPIO (directly uses it), or Directory (Packages it), or Dockerfile (Builds it).

TODO: Regenerate tests before merging

Blocked on: https://github.com/unikraft-cloud/platform/pull/705
Closes: TOOL-694

### DIFF
--- a/cmd/unikraft/testdata/TestGolden/volumes/import/dir
+++ b/cmd/unikraft/testdata/TestGolden/volumes/import/dir
@@ -1,0 +1,29 @@
+$ unikraft volume create --output quiet --set name=test-$UNIQ_VOLUME --set size=10 --set metro=test
+
+stdout:
+	test/test-<VOLUME>
+
+$ unikraft volume import test-$UNIQ_VOLUME --source .
+
+stderr:
+	│ importing data into volume size=276B volume=test-<VOLUME>
+	│ import complete free=10MiB total=10MiB volume=test-<VOLUME>
+
+$ unikraft volume inspect test-$UNIQ_VOLUME
+
+stdout:
+	metro:        test
+	name:         test-<VOLUME>
+	uuid:         12345678-1234-1234-1234-123456789abc
+	state:        available
+	size:         10MiB
+	filesystem:   virtiofs
+	quota-policy: static
+	persistent:   true
+	timestamps:
+	  created:    RELATIVE_TIME
+
+$ unikraft volume delete test-$UNIQ_VOLUME
+
+stdout:
+	test-<VOLUME>
--- a/cmd/unikraft/volumes_test.go
+++ b/cmd/unikraft/volumes_test.go
@@ -5,7 +5,11 @@
 
 package main
 
-import "testing"
+import (
+	"regexp"
+	"slices"
+	"testing"
+)
 
 func volumesTests(t *testing.T, r *testRunner) {
 	t.Run("help", func(t *testing.T) {
@@ -16,6 +20,7 @@ func volumesTests(t *testing.T, r *testRunner) {
 			{args: []string{unikraftCmd, "volume", "wait", "--help"}},
 			{args: []string{unikraftCmd, "volume", "create", "--help"}},
 			{args: []string{unikraftCmd, "volume", "clone", "--help"}},
+			{args: []string{unikraftCmd, "volume", "import", "--help"}},
 			{args: []string{unikraftCmd, "volume", "edit", "--help"}},
 			{args: []string{unikraftCmd, "volume", "delete", "--help"}},
 		})
@@ -58,5 +63,120 @@ func volumesTests(t *testing.T, r *testRunner) {
 				{args: []string{unikraftCmd, "volume", "inspect", "test-$UNIQ_VOLUME", "test-$UNIQ_VOLUME_CLONE"}},
 				{args: []string{unikraftCmd, "volume", "delete", "test-$UNIQ_VOLUME", "test-$UNIQ_VOLUME_CLONE"}},
 			})
+	})
+
+	t.Run("import", func(t *testing.T) {
+		// Offline: missing --source errors before any network call.
+		t.Run("missing-source", func(t *testing.T) {
+			r.run(t, []command{
+				{args: []string{unikraftCmd, "volume", "import", "my-volume"}, allowErr: true},
+			})
+		})
+
+		// Offline: port below the allowed range errors before any network call.
+		t.Run("invalid-port", func(t *testing.T) {
+			r.run(t, []command{
+				{args: []string{unikraftCmd, "volume", "import", "my-volume", "--source", ".", "--port", "80"}, allowErr: true},
+			})
+		})
+
+		// Offline: port above the allowed range errors before any network call.
+		t.Run("invalid-port-high", func(t *testing.T) {
+			r.run(t, []command{
+				{args: []string{unikraftCmd, "volume", "import", "my-volume", "--source", ".", "--port", "99999"}, allowErr: true},
+			})
+		})
+
+		// Import a small directory into a freshly created volume.
+		t.Run("dir", func(t *testing.T) {
+			r.
+				online().
+				withCleaners(
+					[]cleaner{
+						// free size differs on every run
+						{
+							pattern: regexp.MustCompile(`free=[0-9]+.?[0-9]*MiB`),
+							repl:    "free=10MiB",
+						},
+					},
+				).
+				withContext(map[string]string{
+					"hello.txt": "hello from volume import\n",
+				}).
+				run(t, []command{
+					{args: []string{unikraftCmd, "volume", "create", "--output", "quiet", "--set", "name=test-$UNIQ_VOLUME", "--set", "size=10", "--set", "metro=" + metroName}},
+					{args: []string{unikraftCmd, "volume", "import", "test-$UNIQ_VOLUME", "--source", "."}},
+					{args: []string{unikraftCmd, "volume", "inspect", "test-$UNIQ_VOLUME"}},
+					{args: []string{unikraftCmd, "volume", "delete", "test-$UNIQ_VOLUME"}},
+				})
+		})
+
+		// Import a file into a freshly created volume and test connection.
+		t.Run("serve", func(t *testing.T) {
+			r.
+				online().
+				withCleaners(
+					append(slices.Clone(instanceCleaners), cleaner{
+						// free size differs on every run
+						pattern: regexp.MustCompile(`free=[0-9]+.?[0-9]*MiB`),
+						repl:    "free=50MiB",
+					}),
+				).
+				withContext(map[string]string{
+					"index.html": "<html><body>hello from volume import</body></html>\n",
+				}).
+				run(t, []command{
+					// Create a volume to hold the custom web content
+					{args: []string{
+						unikraftCmd, "volume", "create",
+						"--output", "quiet",
+						"--set", "name=test-$UNIQ_VOL",
+						"--set", "size=50",
+						"--set", "metro=" + metroName,
+					}},
+					// Import the custom index.html into the volume
+					{args: []string{unikraftCmd, "volume", "import", "test-$UNIQ_VOL", "--source", "."}},
+					{args: []string{
+						unikraftCmd, "instance", "create",
+						"--set", "name=test-$UNIQ_INST",
+						"--set", "metro=" + metroName,
+						"--set", "image=test/nginx-compat:latest",
+						"--set", "autostart=true",
+						"--set", "resources.memory=256",
+						"--set", "resources.vcpus=1",
+						"--set", "volumes=test-$UNIQ_VOL:/wwwroot:ro",
+						"--set", "service.services=443:8080/tls+http",
+						"--set", "service.domains=name=$UNIQ_DOMAIN",
+					}},
+					// Capture the assigned FQDN
+					{
+						args: []string{
+							unikraftCmd, "instance", "inspect", "test-$UNIQ_INST",
+							"--output", "template=" + `{{ (index .service.domains 0).fqdn }}`,
+						},
+						captureEnv: "FQDN",
+					},
+					{args: []string{unikraftCmd, "instance", "wait", "--until", "state==running", "--timeout", "30s", "test-$UNIQ_INST"}},
+					// Curl the instance and write the body to a file for content verification.
+					{args: []string{
+						"curl",
+						"-k",
+						"--fail",
+						"--silent",
+						"--show-error",
+						"--output", "response.html",
+						"--retry", "10",
+						"--retry-delay", "2",
+						"--retry-all-errors",
+						"--connect-timeout", "5",
+						"--max-time", "10",
+						"https://$FQDN",
+					}},
+					// Assert the imported content is served.
+					{args: []string{"grep", "hello from volume import", "response.html"}},
+					{args: []string{unikraftCmd, "instance", "delete", "test-$UNIQ_INST"}},
+					{args: []string{unikraftCmd, "volume", "delete", "test-$UNIQ_VOL"}},
+				})
+		})
 	})
 }
--- a/cmd/unikraft/volumes_test.go
+++ b/cmd/unikraft/volumes_test.go
@@ -7,7 +7,6 @@ package main
 
 import (
 	"regexp"
-	"slices"
 	"testing"
 )
 
@@ -91,15 +90,13 @@ func volumesTests(t *testing.T, r *testRunner) {
 		t.Run("dir", func(t *testing.T) {
 			r.
 				online().
-				withCleaners(
-					[]cleaner{
-						// free size differs on every run
-						{
-							pattern: regexp.MustCompile(`free=[0-9]+.?[0-9]*MiB`),
-							repl:    "free=10MiB",
-						},
+				withCleaners([]cleaner{
+					// free size differs on every run
+					{
+						pattern: regexp.MustCompile(`free=[0-9]+.?[0-9]*MiB`),
+						repl:    "free=10MiB",
 					},
-				).
+				}).
 				withContext(map[string]string{
 					"hello.txt": "hello from volume import\n",
 				}).
@@ -115,13 +112,14 @@ func volumesTests(t *testing.T, r *testRunner) {
 		t.Run("serve", func(t *testing.T) {
 			r.
 				online().
-				withCleaners(
-					append(slices.Clone(instanceCleaners), cleaner{
+				withCleaners(instanceCleaners).
+				withCleaners([]cleaner{
+					{
 						// free size differs on every run
 						pattern: regexp.MustCompile(`free=[0-9]+.?[0-9]*MiB`),
 						repl:    "free=50MiB",
-					}),
-				).
+					},
+				}).
 				withContext(map[string]string{
 					"index.html": "<html><body>hello from volume import</body></html>\n",
 				}).
--- a/internal/builder/build.go
+++ b/internal/builder/build.go
@@ -31,6 +31,7 @@ type BuildOpts struct {
 
 type RootfsOpts struct {
 	Path string
+	Type RootfsType
 
 	// Output params
 	Format     kraftfile.FsType
@@ -51,6 +52,9 @@ type RootfsType string
 
 const (
 	RootfsTypeDockerfile RootfsType = "dockerfile"
+	RootfsTypeCpio       RootfsType = "cpio"
+	RootfsTypeErofs      RootfsType = "erofs"
+	RootfsTypeDir        RootfsType = "dir"
 )
 
 // Build a unikraft image based on the provided build options.
--- a/internal/builder/cpio/cpio.go
+++ b/internal/builder/cpio/cpio.go
@@ -40,6 +40,7 @@ func CreateFSFromDirectory(ctx context.Context, w io.Writer, source string, opts
 	}
 
 	cw := cpio.NewWriter(w)
+	defer cw.Close()
 
 	type hardlinkKey struct {
 		device int
--- a/internal/builder/kraftfile.go
+++ b/internal/builder/kraftfile.go
@@ -8,8 +8,6 @@ package builder
 import (
 	"fmt"
 	"path/filepath"
-	"slices"
-	"strings"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 
@@ -59,14 +57,18 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
 
 	if kf.Rootfs != nil {
 		opts.Rootfs.Format = kf.Rootfs.Format
-		base := filepath.Base(kf.Rootfs.Source)
-		if base == "Dockerfile" || slices.Contains(strings.Split(base, "."), "Dockerfile") {
-			opts.Rootfs.Path = filepath.Dir(kf.Rootfs.Source)
-		} else {
-			return BuildOpts{}, fmt.Errorf("unable to determine rootfs type for source %q", kf.Rootfs.Source)
+		sourcePath := filepath.Join(dir, kf.Rootfs.Source)
+		typ, err := DetectRootfsType(sourcePath)
+		if err != nil {
+			return BuildOpts{}, fmt.Errorf("detecting rootfs type for %q: %w", kf.Rootfs.Source, err)
 		}
-
-		opts.Rootfs.Path = filepath.Join(dir, opts.Rootfs.Path)
+		opts.Rootfs.Type = typ
+		switch typ {
+		case RootfsTypeDockerfile:
+			opts.Rootfs.Path = filepath.Dir(sourcePath)
+		default:
+			opts.Rootfs.Path = sourcePath
+		}
 	}
 
 	return opts, nil
--- a/internal/builder/rootfs.go
+++ b/internal/builder/rootfs.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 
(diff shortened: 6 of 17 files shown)


internal/cmd/volumes.go
```diff
@@ -608,3 +619,249 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+		Image:         new(image),
+		MemoryMb:      new(volimportMemoryMB),
+		Args:          args,
+		Autostart:     new(true),
+		TimeoutS:      new(volimportStartTimeoutS),
+		RestartPolicy: new(platform.CreateInstanceRequestRestartPolicyNever),
+		Features:      []platform.CreateInstanceRequestFeatures{
+			// TODO(craciunoiuc): Feature is actually `delete-on-stop`.
+			// Update the SDK and rename accordingly and edit the line below.
+			// platform.CreateInstanceRequestFeaturesDelete_on_stop,
```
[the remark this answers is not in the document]
craciunoiuc: TODO: Update when Proto is updated

internal/cmd/volumes.go
```diff
@@ -608,3 +620,263 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+			return fmt.Errorf("waiting for import instance to stop: %w", err)
+		}
+		return nil
+	})
+}
+
+// runVolimport creates a short-lived Unikraft Cloud instance running the
+// volimport service with the given volume attached at /mnt, and returns the
+// instance UUID and the public FQDN on which the import service listens.
+func runVolimport(ctx context.Context, c multimetro.MetroClient, image, volUUID, authStr string, timeoutS uint64, servicePort int) (instUUID, fqdn string, err error) {
```
jedevc: Same as in https://github.com/unikraft/cli/pull/248/changes#r3044440629, can you split this impl out?

And link to the underlying implementation as well in that package, like in https://github.com/unikraft/cli/pull/248/changes#r3044467175.

internal/builder/cpio/cpio.go
```diff
@@ -151,5 +156,318 @@ func CreateFSFromDirectory(ctx context.Context, w io.Writer, source string, opts
 		return fmt.Errorf("could not walk output path: %w", err)
 	}
 
+	return cw.Close()
+}
+
+// IsCPIOFile reports whether path is an SVR4 CPIO archive by inspecting its
+// magic bytes (newc: "070701", newcrc: "070702").
+func IsCPIOFile(path string) bool {
```
jedevc: Woop, this belongs in https://github.com/unikraft/go-cpio please!

There's already a global var for it as well. https://github.com/unikraft/go-cpio/blob/ff0b6bb000b45c2399e8556c16d2e44c0a0d86fc/svr4.go#L21-L21

internal/builder/cpio/cpio.go
```diff
@@ -151,5 +156,318 @@ func CreateFSFromDirectory(ctx context.Context, w io.Writer, source string, opts
...
+}
+
+// stopResponse is the final message sent by the volimport service once all
+// archive entries have been processed.  Its layout is identical to
+// startResponse.
+type stopResponse struct {
+	Free   uint64
+	Total  uint64
+	Maxlen uint64
+}
```
jedevc: This isn't `cpio` related, so should go in a different pkg.

internal/builder/rootfs.go
```diff
@@ -205,14 +232,34 @@ func BuildRootfs(ctx context.Context, c *client.Client, opts BuildOpts) (_ []*im
...
+	defer func() {
+		if rerr != nil {
+			for _, root := range roots {
+				if root.File != nil {
+					root.File.Close()
+					os.Remove(root.File.Name())
+				}
+			}
+		}
+	}()
```
jedevc: suggestion: Add a cleanup method to `root` so that this is simplified.

internal/builder/import.go
```diff
@@ -0,0 +1,137 @@
...
+	"unikraft.com/cli/internal/buildkit"
+)
+
+// BuildImportRootfs builds a CPIO archive from source and returns its file
+// path and byte size.
+// Source may be:
+//
+//   - a Dockerfile context — built with BuildKit and exported as a CPIO archive
+//   - a local directory path — walked and archived with cpio.CreateFSFromDirectory
+//   - an existing CPIO file — returned as-is (caller must NOT remove it)
```
jedevc: Ah, can we change how this is done, to instead take a `imagespec.File`? This means that we keep all the automagic detection *out of here*. Ideally, the detection logic is extracted into a common function with https://github.com/unikraft/cli/blob/3d0d6f44c61c6dfaee23ae0752e804871eaeec83/internal/builder/kraftfile.go#L58-L66, which does the detection, and creates the opts, which then we can use to commonly call `BuildRootfs` or whatever it's been refactored to.

This keeps the separation between "detecting" the type of the source, the build of a filesystem (for volumes, roms, initrd), and the upload of the volume. Meaning more code reuse, less code in general, and fewer places to patch if we ever need to.

internal/cmd/volumes.go
```diff
@@ -608,3 +620,263 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+	volimportMemoryMB      = int64(128)
+	volimportStopTimeoutMs = int64(1100)
+	volimportStartTimeoutS = int64(3)
+)
+
+// VolumeImportCmd imports data from a local source into a volume by
+// temporarily spinning up a volimport instance, streaming a CPIO archive
+// over TLS, and cleaning up the instance afterwards.
+type VolumeImportCmd struct {
+	Volume  string `arg:"" completion-predictor:"resource-key-volume" help:"Name or UUID of the volume to import into."`
```
jedevc: Hm, let's make this required. I don't like the automatic behavior of "just read the current dir", it's too easy to accidentally import all your local files then by accident.

internal/cmd/volumes.go
```diff
@@ -608,3 +620,263 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+	Source  string `short:"s" default:"." help:"Data source: local directory or CPIO archive." placeholder:"path"`
+	Port    int    `short:"p" default:"42069" help:"Port to connect to the volume import service on." placeholder:"port"`
+	Image   string `default:"official/utils/volimport:1.0" help:"Volume import image to use." placeholder:"image"`
+	Timeout uint64 `short:"t" default:"10" help:"Inactivity timeout in seconds for the import service." placeholder:"seconds"`
```
jedevc: Like in the other PR, what's the use case for setting all these? They seem internal.

We can hide them if we really need to.

internal/cmd/volumes.go
```diff
@@ -608,3 +620,263 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+	Force   bool   `short:"f" help:"Force import even if the data might exceed volume capacity."`
+	Workdir string `short:"w" help:"Working directory used when building the CPIO archive." placeholder:"dir"`
```
jedevc: :eyes: what's this for? Why do we need this at all?

internal/builder/cpio/cpio.go
craciunoiuc: 👀 oke

internal/builder/cpio/cpio.go
craciunoiuc: True

internal/builder/rootfs.go
craciunoiuc: hmm, oke

internal/cmd/volumes.go
craciunoiuc: not sure on what you commented? maybe on the `force`? we need that because it doesn't know if you're overwriting files

If you have 1TB total space, and one 600MB file in it, if you're overwriting it, the volimport will initially think that you will run out of space because `600MB` > the remaining `400MB`

internal/cmd/volumes.go
craciunoiuc: same argument like there, they might be needed, but are not mandatory

If we can hide them but keep them, it would be ideal

internal/cmd/volumes.go
jedevc: I mean for `Workdir`.

internal/cmd/volumes.go
craciunoiuc: Hmm, I don't think it's importing the current dir? It's importing the Kraftfile rootfs in the current dir

internal/cmd/volumes.go
craciunoiuc: same here: a separate package in `internal` or in `x`?

jedevc: Internal is good for now.

internal/builder/import.go
craciunoiuc: Hmm, if you say so, hopefully that works 👀

internal/cmd/volumes.go
craciunoiuc: hmm, I thought it was useful? 😅 

I guess it's done different in the `cli`

jedevc: What's it supposed to be for?

craciunoiuc: Workdir is for specifying the env you are building in?

I don't get the question so I will just remove it and it will only work in `cwd`

internal/cmd/volumes.go
craciunoiuc: Ok, will add that, I don't have anything against it

internal/builder/import.go
craciunoiuc: Ok, I read this at least 6-7 times and I still did not get it, so I will assume that what you wanted to say is:
* I revert the change to package images again like before
* I access the rootfs layer that is currently in the export, basically putting the two together

jedevc: I think that makes sense, yes.

craciunoiuc: Something is seriously wrong with that `imgspec` package, I need to look into it but I couldn't get it to behave properly.

craciunoiuc:
I split all the suggested changes into separate commits. I will need to squash some of them up before it's ready to review again.

Currently a bit broken as the `imgspec` does not return the path to the computer layer.

Also rootfs building does not work without dockerfiles. (aka with directories or files)

With these out of the way it should work again.

jedevc:
> Currently a bit broken as the imgspec does not return the path to the computer layer.

Why do you need the "computer layer"? Do you need to open the file? If so, you can call `Open`. What specifically needs the path here?

> Also rootfs building does not work without dockerfiles. (aka with directories or files)

Indeed :tada: I was hoping this was one of the things you might take a look at at some point, or even as part of this (though maybe in a separate PR).

craciunoiuc:
>  I was hoping this was one of the things you might take a look at at some point

Hmm, then something is weird, because you told me to revert the file/directory/dockerfile detection that I previously had. (which is why I did them as commits on top instead of overwriting hehe.

> Why do you need the "computer layer"?

Lol, not sure how I wrote that. I wanted to say the `rootfs` layer. I was hoping the `imagespec` directory would give me to path to the file if I call `images[0].Initrd.Path()` on it for examples. But it seems that it's always empty.

jedevc:
The detection should exist, yes, but it should be extracted - it shouldn't be done just for the volume imports. We should have *one* reusable function that does "pass this thing in, and build a rootfs from it". We may need to rework those functions and everything so that the signatures all work out. What we need to avoid is rewriting the same detection everywhere.

jedevc:
`Path()` is not documented well, but it's the path *inside the tarball*. e.g. for a packaged kernel, the kernel lives inside the tarball at `/unikraft/bin/kernel` for example. `Path()` says which of those it's at.

Looking at how it's used actually... we could remove that method entirely.

internal/builder/import.go
jedevc: > Something is seriously wrong with that `imgspec` package, I need to look into it but I couldn't get it to behave properly.

Very possible. Curious what you ran into, but yeah, bugs are to be expected :grin: we've only just started moving stuff over to it.

internal/builder/import.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if len(images) == 0 {
+		return "", -1, fmt.Errorf("no images were built from the provided source")
+	}
+
+	if len(images) > 1 {
+		return "", -1, fmt.Errorf("multiple images were built from the provided source; expected exactly one")
+	}
+
+	initrd := images[0].Initrd
```
jedevc: Okay, this is *so* much better than before. Keeping all the stuff in `BuildRootfs` feels so much neater.

I think there's a couple more bits though that would make it even neater:
- We should actually call `BuildRootfs` in `volumes.go`.
- Then this function would only take the `imagespec.File` directly, it wouldn't do the build itself.

Although, I actually think it's much simpler - this function doesn't really need to exist. We don't need to return a path and create a temp file. The only thing we're using it for is passing it into `volimport.Copy`, which only uses it to open an `io.ReadCloser`. We should just directly pass `io.ReadCloser` into `Copy`, which we can get from `initrd.Open`.

Does that make sense?

cmd/unikraft/volumes_test.go
jedevc, line 1: Can you also add a full test into `volumes_test.go`? e.g. look at how `clone` works.

One way to test this would be to create a volume, import some sample files, then mount that volume into an nginx instance so that it's serving those files. Then we just need to request those files like in the [`connect` test](https://github.com/unikraft/cli/blob/b4d1255b0000a55f77cbcd71083d0a688113c2d3/cmd/unikraft/instances_test.go#L89-L89).

internal/builder/cpio/cpio.go
```diff
@@ -151,5 +151,5 @@ func CreateFSFromDirectory(ctx context.Context, w io.Writer, source string, opts
 		return fmt.Errorf("could not walk output path: %w", err)
 	}
 
-	return nil
+	return cw.Close()
```
jedevc: Eek, not great that this was missing 😢 that's my bad.

We should make to close it even in the error cases though.

internal/builder/rootfs.go
```diff
@@ -26,25 +26,185 @@ import (
...
+//
+//  1. A pre-packaged rootfs file - returned as-is.
+//  2. A directory - walked and archived.
+//  3. Anything else - treated as a Dockerfile context and built with BuildKit.
 func BuildRootfs(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image, rerr error) {
 	if len(opts.Platform) == 0 {
 		return nil, fmt.Errorf("at least one platform must be specified")
 	}
 
+	if opts.Rootfs.Path != "" {
```
jedevc: This feels pretty similar to this check: https://github.com/unikraft/cli/blob/fe6dbedf3fe17f0f21a3e6913f678c891fab8a48/internal/builder/kraftfile.go#L60-L70

Instead of doing it in both places, we should do the detection all in that block, and set a `opts.Rootfs.Type`, to `dockerfile`, or `cpio`, or `erofs`, or `dir`, or whatever.

Then, here, we can just do a massive switch statement for `case "dockerfile":`, `case "cpio":`, probably splitting out the actual implementation for the dockerfile case like you've done for `buildRootfsFromPackaged` and similar.

internal/builder/rootfs.go
```diff
@@ -26,25 +26,185 @@ import (
...
+			imagespec.WithPlatform(p),
+			imagespec.WithInitrd(imagespec.NewTempOSFile(f)),
+		))
+	}
+	return imgs, nil
+}
+
+// PackagedRootfs holds a rootfs archive for a single platform produced by
+// PackageRootfs, together with the OCI image config extracted during the build.
+type PackagedRootfs struct {
```
jedevc: Unused? I think the original was as well tbf.

internal/builder/rootfs.go
```diff
@@ -26,25 +26,185 @@ import (
...
+		cfg.Env = env
+	}
+	cfg.Labels = opts.Labels
+	return cfg
+}
+
+// buildRootfsFromPackaged returns images backed by an already-packaged rootfs
+// file. The file is opened read-only per platform.
+// The caller must not delete it.
+func buildRootfsFromPackaged(_ context.Context, opts BuildOpts) (_ []*imagespec.Image, rerr error) {
```
jedevc: nit: these functions are good, but they should be *below* the `BuildRootfs` function, since they're helpers for just that function. It also helps a reader looking around to see the public methods generally towards the top, with the least significant stuff towards the bottom, so that the file can kind of be read top-to-bottom, instead of needing to jump around a lot (also helpful for LLMs, not just people).

r.e. naming, it's a personal preference, but we don't need excess words like `from` here. `buildRootfsPackaged` is fine, we're not programming java ☕

internal/builder/rootfs.go
```diff
@@ -26,25 +26,185 @@ import (
...
+			}
+
+			if err = erofs.CreateFSFromDirectory(ctx, f, opts.Rootfs.Path,
+				erofs.WithAllRoot(!opts.Rootfs.KeepOwners),
+			); err != nil {
+				return nil, fmt.Errorf("could not create EroFS archive: %w", err)
+			}
+		default:
+			return nil, fmt.Errorf("unknown filesystem type %q", format)
+		}
```
jedevc: This entire block is identical for the dockerfile build right? We should split it out to something like `packageRootfs`, and share it between the two, otherwise we risk drift and bugs.

internal/cmd/volumes.go
```diff
@@ -610,3 +620,181 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+
+// VolumeImportCmd imports data from a local source into a volume by
+// temporarily spinning up a volimport instance, streaming a CPIO archive
+// over TLS, and cleaning up the instance afterwards.
+type VolumeImportCmd struct {
+	Volume  string `arg:"" completion-predictor:"resource-key-volume" help:"Name or UUID of the volume to import into." create:"set,required"`
+	Source  string `short:"s" help:"Data source: local directory or CPIO archive." placeholder:"path" create:"set,required"`
+	Port    int    `short:"p" default:"42069" help:"Port to connect to the volume import service on." placeholder:"port" hidden:"true"`
+	Image   string `default:"official/utils/volimport:1.0" help:"Volume import image to use." placeholder:"image" hidden:"true"`
+	Timeout uint64 `short:"t" default:"10" help:"Inactivity timeout in seconds for the import service." placeholder:"seconds" hidden:"true"`
```
jedevc, line 632: nit: `hidden` dev options should appear at the very bottom in their own block.

internal/cmd/volumes.go
```diff
@@ -610,3 +620,181 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+	}
+
+	workdir, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("getting current working directory: %w", err)
+	}
+
+	var importOpts *builder.BuildOpts
+	if c.Source == "" {
+		kf, err := kraftfile.ParseDirectory(workdir, kraftfile.WithSkippedVersionCheck())
```
jedevc: This logic isn't used anymore? See previous comment in https://github.com/unikraft/cli/pull/245#discussion_r3044668377

internal/cmd/volumes.go
```diff
@@ -610,3 +620,181 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+	vol, ok := resources[0].(Volume)
+	if !ok {
+		return fmt.Errorf("unexpected resource type %T", resources[0])
+	}
+	if vol.UUID == "" {
+		return fmt.Errorf("volume %q is missing a UUID", vol.Name)
+	}
+	if vol.Metro == nil {
+		return fmt.Errorf("volume %q has no metro information", vol.Name)
+	}
```
jedevc: No need to check these, they can't happen. Or if they do, something utterly completely disastrous is happening.

internal/cmd/volumes.go
```diff
@@ -610,3 +620,181 @@ func volumePatchSpec(path string, _ patchOp, value any) (platform.UpdateVolumesR
...
+	}()
+
+	// Open a TLS connection to the instance and stream the CPIO archive.
+	instAddr := instFQDN + ":" + strconv.Itoa(c.Port)
+	log.G(ctx).Info().
+		Str("size", units.BytesSize(float64(cpioSize))).
+		Str("addr", instAddr).
+		Msg("importing data into volume")
+
+	conn, err := tls.Dial("tcp4", instAddr, nil)
```
jedevc: nit: why not just `tcp`? Even if we don't support ip6 now, we shouldn't assume it in our cli code.

internal/volimport/volimport.go
```diff
@@ -0,0 +1,133 @@
...
+
+	log.G(ctx).Trace().Msg("creating volume data import instance")
+	resp, err := c.CreateInstance(ctx, platform.CreateInstanceRequest{
+		Image:         new(image),
+		MemoryMb:      new(memoryMB),
+		Args:          args,
+		Autostart:     new(true),
+		TimeoutS:      new(startTimeoutS),
+		RestartPolicy: new(platform.CreateInstanceRequestRestartPolicyNever),
+		Features:      []platform.CreateInstanceRequestFeatures{platform.CreateInstanceRequestFeatures("delete-on-stop")},
```
jedevc:
```suggestion
		Features:      []platform.CreateInstanceRequestFeatures{platform.CreateInstanceRequestFeaturesDelete_on_stop},
```

internal/cmd/volumes.go
craciunoiuc: Link doesn't work for me, but basically this covers the case where you don't specify a source, so it checks the cwd for a kraftfile to read the source from

Are you saying we should mandate a `--source` and ignore Kraftfiles completely for this command?

internal/volimport/volimport.go
craciunoiuc: Perfect, this was added, nice

internal/builder/import.go
craciunoiuc: Thought you would said this, but kept it in just in case >:D

will revisit

craciunoiuc:
All comments make sense, will address, next time you see this PR it will have the changed commit history

internal/cmd/volumes.go
jedevc: > Hm, let's make this required. I don't like the automatic behavior of "just read the current dir", it's too easy to accidentally import all your local files then by accident.

(yeah, weird, link doesn't work - copy-pasted above)

internal/builder/rootfs.go
craciunoiuc: Hmm, wait, but this only works if you use a kraftfile if I move it there

What if you don't use a kraftfile like in almost all cases of the volimport actually

jedevc: Ahhhhh. Great. Okay, then we should move the logic out of there then, and move it into `BuildRootfs` right? We're just trying to consolidate the logic, it doesn't really matter where it gets consolidated to.

internal/builder/rootfs.go
```diff
@@ -26,25 +27,104 @@ import (
...
+	switch opts.Rootfs.Type {
+	case RootfsTypeCpio, RootfsTypeErofs:
+		return buildRootfsPackaged(ctx, opts)
+	case RootfsTypeDir:
+		return buildRootfsDirectory(ctx, opts)
+	case RootfsTypeDockerfile:
+		opts.Rootfs.Path = filepath.Dir(opts.Rootfs.Path)
+	default:
+		log.G(ctx).Warn().Str("type", string(opts.Rootfs.Type)).Str("path", opts.Rootfs.Path).Msgf("unknown rootfs type; treating as Dockerfile context")
+	}
```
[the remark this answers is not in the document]
craciunoiuc: Wanted.

craciunoiuc:
^ Will do monday.

craciunoiuc:
Integration is really weird, at least 2-3 things related to the server itself

PR should be ready

cmd/unikraft/testdata/TestGolden/volumes/import/dir
```diff
@@ -0,0 +1,29 @@
...
+stdout:
+	test/test-<VOLUME>
+
+$ unikraft volume import test-$UNIQ_VOLUME --source .
+
+stderr:
+	│ importing data into volume size=276B volume=test-<VOLUME>
+	│ import complete free=9.996MiB total=10MiB volume=test-<VOLUME>
+
+$ unikraft volume inspect test-$UNIQ_VOLUME
```
jedevc, line 12: Can you also check that the contents are imported correctly? I would start an nginx instance with the volume mounted and request the contents.

internal/builder/rootfs.go
```diff
@@ -26,25 +27,108 @@ import (
...
+		if f, err := os.Stat(opts.Rootfs.Path); err != nil {
+			if os.IsNotExist(err) {
+				return nil, fmt.Errorf("dockerfile does not exist")
+			}
+			return nil, fmt.Errorf("checking dockerfile path %q: %w", opts.Rootfs.Path, err)
+		} else if !f.IsDir() {
+			opts.Rootfs.Path = filepath.Dir(opts.Rootfs.Path)
+		}
+	default:
+		log.G(ctx).Warn().Str("type", string(opts.Rootfs.Type)).Str("path", opts.Rootfs.Path).Msgf("unknown rootfs type; treating as dockerfile context")
```
jedevc: This should be an error?

internal/builder/rootfs.go
```diff
@@ -26,25 +27,108 @@ import (
...
+		return buildRootfsDirectory(ctx, opts)
+	case RootfsTypeDockerfile:
+		if f, err := os.Stat(opts.Rootfs.Path); err != nil {
+			if os.IsNotExist(err) {
+				return nil, fmt.Errorf("dockerfile does not exist")
+			}
+			return nil, fmt.Errorf("checking dockerfile path %q: %w", opts.Rootfs.Path, err)
+		} else if !f.IsDir() {
+			opts.Rootfs.Path = filepath.Dir(opts.Rootfs.Path)
+		}
```
jedevc, line 127: Avoid fallthrough here, we can just directly call something like `buildRootfsDockerfile`.

internal/builder/build.go
```diff
@@ -51,6 +52,11 @@ type RootfsType string
 
 const (
 	RootfsTypeDockerfile RootfsType = "dockerfile"
+	RootfsTypeCpio       RootfsType = "cpio"
+	RootfsTypeErofs      RootfsType = "erofs"
+	RootfsTypeDir        RootfsType = "dir"
+	// Unused for now, but may be useful in ROMs
+	RootfsTypeFile RootfsType = "file"
```
jedevc: We should remove this for now if it's not being used.

internal/builder/kraftfile.go
```diff
@@ -59,14 +57,18 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
 
 	if kf.Rootfs != nil {
 		opts.Rootfs.Format = kf.Rootfs.Format
-		base := filepath.Base(kf.Rootfs.Source)
-		if base == "Dockerfile" || slices.Contains(strings.Split(base, "."), "Dockerfile") {
-			opts.Rootfs.Path = filepath.Dir(kf.Rootfs.Source)
-		} else {
-			return BuildOpts{}, fmt.Errorf("unable to determine rootfs type for source %q", kf.Rootfs.Source)
+		sourcePath := filepath.Join(dir, kf.Rootfs.Source)
+		typ, err := DetectRootfsType(sourcePath)
```
jedevc, line 61: Can you follow-up with a PR to x/kraftfile to add a field for this explicitly?

jedevc (commented):
Only nits, happy to merge when they're resolved.

internal/builder/rootfs.go
craciunoiuc: Unsure really how the stuff underneath works, wanted to be optimistic :D

I can make it error out sure

internal/builder/kraftfile.go
craciunoiuc: shouldn't I do this before, so I can rebase this?

internal/builder/rootfs.go
craciunoiuc: would mean moving the stuff below also to a function, sure

cmd/unikraft/testdata/TestGolden/volumes/import/dir
craciunoiuc: Sure thing, will test again

It worked the last time I tried it (so not recent enough >:D )

internal/builder/kraftfile.go
jedevc: I would keep the detection logic here for when that field is empty. Up to you on ordering, I have no preference.

craciunoiuc:
finally, it feels like 1 morbillion years have passed

jedevc (approved):

PR #248, opened 2026-04-07
feat(internal/cmd): Add 'instance tunnel' command

craciunoiuc:
Nodes need this option `--net-internal-traffic=vm` in order to test it.

```
tunnel-3uhz1 │ [    0.141932] ERR:  [apptunlr] Target remote not in the same subnet.
tunnel-3uhz1 │ [    0.142826] ERR:  [apptunlr] Failed to hook TCP2TCP.
```

~The Tunlr also needs updating to remove the check for inter-vm comms.~
Done.

~TODO: Update tests when platform is stable.~

~Blocked on: https://github.com/unikraft-cloud/platform/pull/705~
Closes: TOOL-714

### DIFF
--- a/cmd/unikraft/integration/instance_test.go
+++ b/cmd/unikraft/integration/instance_test.go
@@ -6,6 +6,7 @@
 package integration
 
 import (
+	"context"
 	_ "embed"
 	"strings"
 	"testing"
@@ -15,11 +16,47 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"unikraft.com/cloud/sdk/platform"
+	"unikraft.com/cloud/sdk/platform/group"
+
+	"unikraft.com/cli/internal/config"
 	integ "unikraft.com/cli/internal/integration"
+	"unikraft.com/cli/internal/multimetro"
 )
 
 //go:embed testdata/counter_server.py
 var counterServerPy []byte
+
+// tunnelProxyUUIDs queries the platform directly (bypassing the CLI's
+// resource sandbox, which never tracks the tunnel command's internal proxy
+// instance since it's created via the raw platform client rather than
+// through a sandboxed resource) for every currently-running tunnel-service
+// instance in the given metro.
+func tunnelProxyUUIDs(t *testing.T, cfg *config.Config, metro string) map[string]struct{} {
+	t.Helper()
+	ctx := config.WithConfig(t.Context(), cfg)
+	g, err := multimetro.NewClient(ctx)
+	require.NoError(t, err)
+
+	uuids := make(map[string]struct{})
+	err = group.DoMetro(ctx, g, metro, func(ctx context.Context, c multimetro.MetroClient) error {
+		resp, err := c.GetInstances(ctx, nil, platform.GetInstancesOpts{Details: new(true)})
+		if err != nil {
+			return err
+		}
+		if resp == nil || resp.Data == nil {
+			return nil
+		}
+		for _, inst := range resp.Data.Instances {
+			if strings.Contains(inst.Image, "utils/tunnel") {
+				uuids[inst.Uuid] = struct{}{}
+			}
+		}
+		return nil
+	})
+	require.NoError(t, err)
+	return uuids
+}
 
 func TestInstances(t *testing.T) {
 	t.Run("create", func(t *testing.T) {
@@ -1087,6 +1124,75 @@ cmd: ["cat", "/marker.txt"]
 		case <-time.After(10 * time.Second):
 			// Command still running
 		}
+	})
+
+	t.Run("tunnel", func(t *testing.T) {
+		r := runner(t, true, []string{staging, stable})
+		instName := uniq()
+		instName2 := uniq()
+
+		// Create two nginx instances without a public service, in the same
+		// metro, so a single tunnel invocation with two targets exercises the
+		// proxy's port-grouping/sequential-assignment logic for real.
+		for _, name := range []string{instName, instName2} {
+			r.Run(t, []string{
+				"unikraft", "instance", "create",
+				"--output", "quiet",
+				"--set", "name=test-" + name,
+				"--set", "metro=" + r.Config.MetroName,
+				"--set", "image=nginx:latest",
+				"--set", "autostart=true",
+				"--set", "resources.memory=128",
+				"--set", "resources.vcpus=1",
+			})
+		}
+		r.Run(t, []string{
+			"unikraft", "instance", "wait", "--until", "state==running", "--timeout", "60s",
+			"test-" + instName, "test-" + instName2,
+		})
+
+		before := tunnelProxyUUIDs(t, r.Config.Config, r.Config.MetroName)
+
+		// Use metro/instance:port/tcp syntax for the tunnel targets.
+		tunnel := r.StartBackground(t,
+			[]string{
+				"unikraft", "instance", "tunnel",
+				"18081:" + r.Config.MetroName + "/test-" + instName + ":8080/tcp",
+				"18082:" + r.Config.MetroName + "/test-" + instName2 + ":8080/tcp",
+			},
+			"127.0.0.1:18081",
+			0,
+		)
+
+		// Verify that we can reach both instances through the tunnel.
+		body := integ.HTTPGet(t, "http://127.0.0.1:18081")
+		assert.Regexp(t, "Thank you for using nginx.", body)
+		body = integ.HTTPGet(t, "http://127.0.0.1:18082")
+		assert.Regexp(t, "Thank you for using nginx.", body)
+
+		// Identify the proxy instance the tunnel command created, so we can
+		// confirm below that tearing down the tunnel actually deletes it
+		// (a regression in Close() would otherwise go unnoticed, since the
+		// CLI's own sandboxed "instance list"/"inspect" never see this
+		// instance either way).
+		during := tunnelProxyUUIDs(t, r.Config.Config, r.Config.MetroName)
+		var proxyUUID string
+		for u := range during {
+			if _, ok := before[u]; ok {
+				continue
+			}
+			require.Empty(t, proxyUUID, "expected exactly one new tunnel proxy instance, found multiple")
+			proxyUUID = u
+		}
+		require.NotEmpty(t, proxyUUID, "expected a new tunnel proxy instance to appear while the tunnel is running")
+
+		tunnel.Stop()
+
+		after := tunnelProxyUUIDs(t, r.Config.Config, r.Config.MetroName)
+		_, stillThere := after[proxyUUID]
+		assert.False(t, stillThere, "tunnel proxy instance %s was not deleted after the tunnel was torn down", proxyUUID)
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName, "test-" + instName2})
 	})
 
 	t.Run("branch", func(t *testing.T) {
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -19,7 +19,9 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/MakeNowJust/heredoc"
 	"github.com/alecthomas/kong"
 	"github.com/distribution/reference"
 	"github.com/go-json-experiment/json/jsontext"
@@ -39,6 +41,7 @@ import (
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
 	"unikraft.com/cli/internal/resource/value"
+	"unikraft.com/cli/internal/tunnel"
 	"unikraft.com/cli/internal/types"
 )
 
@@ -60,6 +63,7 @@ type InstancesCmd struct {
 	Stop    InstancesStopCmd    `cmd:"" help:"Stop one or more instances."`
 	Suspend InstancesSuspendCmd `cmd:"" help:"Suspend one or more instances."`
 	Restart InstancesRestartCmd `cmd:"" help:"Restart one or more instances."`
+	Tunnel  InstancesTunnelCmd  `cmd:"" aliases:"port-forward" help:"Forward a local port to an unexposed instance."`
 	History InstanceHistoryCmd  `cmd:"" help:"Show checkpoint history for an instance."`
 }
 
@@ -1956,4 +1960,79 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
 		return suspended, suspended.Refs(), nil
 	})
 	return multimetro.Keys(suspended), err
+}
+
+type InstancesTunnelCmd struct {
+	Targets []string `arg:"" name:"target" min:"1" help:"Forwarding target(s) in the form [LOCAL_PORT:]<INSTANCE>:DEST_PORT[/TYPE]." placeholder:"<target>"`
+
+	TunnelProxyPorts   []string `short:"p" name:"tunnel-proxy-port"   help:"Remote port(s) exposed by the tunnel service. When a single value is given it is used as the starting port for multiple targets." default:"4444" hidden:""`
+	ProxyControlPort   uint     `short:"P" name:"tunnel-control-port" help:"Command-and-control port of the tunnel service." default:"4443" hidden:""`
+	TunnelServiceImage string   `          name:"tunnel-image"        help:"Image to use for the tunnel service." default:"official/utils/tunnel:1.0" hidden:""`
+}
+
+func (InstancesTunnelCmd) Help() string {
+	return heredoc.Docf(`
+		Forward a local port to an unexposed instance through an intermediate
+		TLS tunnel service.
+
+		When you need to access an instance on Unikraft Cloud which is not
+		publicly exposed to the internet, you can use the
+		%[1]sunikraft instance tunnel%[1]s subcommand to forward from a local
+		port to a port which the instance listens on.
+	`, "`")
+}
+
+func (cmd InstancesTunnelCmd) Examples() []kingkong.Example {
+	return []kingkong.Example{
+		{
+			Description: "Forward local port 8080 to instance \"nginx\" port 8080",
+			Commands:    []string{"unikraft instance tunnel nginx:8080"},
+		},
+		{
+			Description: "Forward local port 8333 to instance \"nginx\" port 8080",
+			Commands:    []string{"unikraft instance tunnel 8333:nginx:8080"},
+		},
+		{
+			Description: "Forward multiple ports from multiple instances",
+			Commands:    []string{"unikraft instance tunnel 8080:my-instance1:8080/tcp 8443:my-instance2:8080/tcp"},
+		},
+		{
+			Description: "Forward local port 8080 to instance \"my-instance1\" port 8080 on fra metro using TCP",
+			Commands:    []string{"unikraft instance tunnel 8080:fra/my-instance1:8080/tcp"},
+		},
+		{
+			Description: "Use a custom relay port to avoid collisions",
+			Commands:    []string{"unikraft instance tunnel -p 5500 my-instance:8080"},
+		},
+	}
+}
+
+func (cmd *InstancesTunnelCmd) Run(ctx context.Context, stdio config.Stdio) error {
+	targets, err := tunnel.ParseTargets(cmd.Targets, cmd.TunnelProxyPorts)
+	if err != nil {
+		return fmt.Errorf("could not parse targets: %w", err)
+	}
+
+	tun, err := tunnel.New(ctx, targets)
+	if err != nil {
+		return fmt.Errorf("could not create tunnel: %w", err)
+	}
+
+	g, err := multimetro.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		// Detach from ctx so cleanup survives the same Ctrl-C that triggered
+		// it, but bound it so a stuck API call can't leave a still-billed
+		// proxy instance orphaned.
+		closeCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+		defer cancel()
+		if err := tun.Close(closeCtx, g); err != nil {
+			log.G(ctx).Error().Err(err).Msg("could not terminate tunnel proxy")
+		}
+	}()
+
+	return tun.Run(ctx, g, cmd.ProxyControlPort, cmd.TunnelServiceImage)
 }
--- a/internal/integration/env.go
+++ b/internal/integration/env.go
@@ -8,11 +8,13 @@ package integration
 import (
 	"bytes"
 	"context"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -145,4 +147,89 @@ func (env *TestEnv) Run(t *testing.T, args []string, opts ...CmdOption) string {
 		require.NoError(t, err, "command %q failed\n%s", strings.Join(args, " "), out)
 	}
 	return out
+}
+
+// BackgroundProcess is a handle to a process started by StartBackground.
+type BackgroundProcess struct {
+	cmd    *exec.Cmd
+	output *bytes.Buffer
+	t      *testing.T
+	once   sync.Once
+}
+
+// Stop terminates the background process, waiting for it to exit. It is safe
+// to call multiple times, and is registered automatically as test cleanup by
+// StartBackground, so it only needs to be called explicitly for early
+// termination.
+func (p *BackgroundProcess) Stop() {
+	p.once.Do(func() {
+		p.t.Helper()
+		p.t.Logf("stopping background process")
+		_ = p.cmd.Cancel()
+		err := p.cmd.Wait()
+		if err != nil && p.output.Len() > 0 {
+			p.t.Logf("background process output:\n%s", p.output.String())
+		}
+	})
+}
+
+// StartBackground starts a command in the background and registers cleanup to
+// stop it when the test ends. If waitAddr is non-empty, it polls that TCP
+// address until it is reachable or the timeout (default 90s) expires.
+// The returned handle's Stop method can be called for early termination.
+func (env *TestEnv) StartBackground(t *testing.T, args []string, waitAddr string, timeout time.Duration) *BackgroundProcess {
+	t.Helper()
+	t.Logf("starting background: %s", strings.Join(args, " "))
+
+	var c *exec.Cmd
+	if args[0] == "unikraft" {
+		c = exec.CommandContext(t.Context(), env.unikraftPath, args[1:]...)
+	} else {
+		c = exec.CommandContext(t.Context(), args[0], args[1:]...)
+	}
+
+	c.Env = os.Environ()
+	c.Env = slices.DeleteFunc(c.Env, func(s string) bool {
+		return strings.HasPrefix(s, "UNIKRAFT_")
+	})
+	c.Env = append(c.Env, "NO_COLOR=1")
+	c.Env = append(c.Env, "UNIKRAFT_CONFIG="+env.configPath)
+	c.Env = append(c.Env, "BUILDKIT_PROGRESS=quiet")
+	c.Env = append(c.Env, resource.UnikraftSandboxEnv+"="+env.sandboxPath)
+	c.Cancel = func() error {
+		return c.Process.Signal(os.Interrupt)
+	}
+	c.WaitDelay = 30 * time.Second
+
+	var output bytes.Buffer
+	c.Stdout = &output
+	c.Stderr = &output
+
+	require.NoError(t, c.Start(), "failed to start background command %q", strings.Join(args, " "))
+
+	proc := &BackgroundProcess{cmd: c, output: &output, t: t}
+	t.Cleanup(proc.Stop)
+
+	if waitAddr != "" {
+		waitForAddr(t, waitAddr, timeout)
+	}
+
+	return proc
+}
+
+func waitForAddr(t *testing.T, addr string, timeout time.Duration) {
+	t.Helper()
+	if timeout == 0 {
+		timeout = 90 * time.Second
+	}
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
+		conn, err := net.DialTimeout("tcp", addr, time.Second)
+		if err == nil {
+			conn.Close()
+			return
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
+	t.Fatalf("timed out waiting for %s to become reachable", addr)
 }
--- a/internal/tunnel/parse.go
+++ b/internal/tunnel/parse.go
@@ -1,0 +1,226 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package tunnel
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"unikraft.com/cli/internal/multimetro"
+	"unikraft.com/cloud/sdk/platform"
+	"unikraft.com/cloud/sdk/platform/group"
+	"unikraft.com/x/log"
+)
+
+// proxyInfo holds the metadata for a running tunnel proxy instance.
+type proxyInfo struct {
+	uuid string
+	fqdn string
+	// closed indicates whether this proxy instance has been closed.
+	closed bool
+}
+
+// ParseTargets parses raw target strings and proxy port specifications into
+// structured target objects.
+// proxyPorts configures which ports the tunnel service exposes:
+// if a single port is given it is used as the starting port for sequential
+// assignment; otherwise there must be exactly one port per target.
+func ParseTargets(rawTargets []string, proxyPorts []string) ([]Target, error) {
+	if len(rawTargets) == 0 {
+		return nil, fmt.Errorf("at least one target must be specified")
+	}
+
+	if len(proxyPorts) == 0 {
+		return nil, fmt.Errorf("at least one proxy port must be specified")
+	}
+
+	targets := make([]Target, len(rawTargets))
+	for i, raw := range rawTargets {
+		t, err := parseTarget(raw)
+		if err != nil {
+			return nil, fmt.Errorf("parsing target %q: %w", raw, err)
+		}
+		targets[i] = t
+	}
+
+	parsedPorts, err := parseProxyPorts(proxyPorts)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(parsedPorts) > 1 && len(parsedPorts) != len(targets) {
+		return nil, fmt.Errorf("number of proxy ports must be either 1 or equal to the number of targets")
+	}
+
+	if len(parsedPorts) == 1 {
+		if int(parsedPorts[0])+len(targets)-1 > 65535 {
+			return nil, fmt.Errorf("starting proxy port %d does not leave enough room for %d target(s)", parsedPorts[0], len(targets))
+		}
+	} else {
+		seen := make(map[uint16]struct{}, len(parsedPorts))
+		for _, p := range parsedPorts {
+			if _, dup := seen[p]; dup {
+				return nil, fmt.Errorf("duplicate proxy port %d", p)
+			}
+			seen[p] = struct{}{}
+		}
+	}
+
+	// Assign exposed proxy ports.
+	for i := range targets {
+		if len(parsedPorts) == 1 {
+			targets[i].exposedProxyPort = parsedPorts[0] + uint16(i)
+		} else {
+			targets[i].exposedProxyPort = parsedPorts[i]
+		}
+	}
+	return targets, nil
+}
+
+func parseProxyPorts(ports []string) ([]uint16, error) {
+	parsed := make([]uint16, 0, len(ports))
+	for _, port := range ports {
+		p, err := strconv.ParseUint(port, 10, 16)
+		if err != nil {
+			return nil, fmt.Errorf("%q is not a valid port number", port)
+		}
+		parsed = append(parsed, uint16(p))
+	}
+	return parsed, nil
+}
+
+// formatProxyArgs formats the arguments to pass to a tunnel service instance
+// for the given set of targets.
+// NOTE(craciunoiuc): Proxy arguments can be found at:
+// https://github.com/unikraft-cloud/catalog/tree/prod-staging/library/official/utils/tunlr/1.0
+func formatProxyArgs(targets []resolvedTarget, authStr string, proxyControlPort uint) []string {
+	connections := make([]string, 0, len(targets))
+	for _, tgt := range targets {
+		connections = append(connections, fmt.Sprintf("TCP2%s:%s:%d:%d:%d",
+			strings.ToUpper(tgt.network),
+			tgt.ip,
+			tgt.dest,
+			tgt.exposedProxyPort,
+			27,
+		))
+	}
+	return []string{
+		// HEARTBEAT_PORT:CTLR_AUTH_TIMEOUT
+		fmt.Sprintf("%d:%d", proxyControlPort, 5),
+		// AUTH_TIMEOUT:AUTH_COOKIE
+		fmt.Sprintf("%d:%s", 5, authStr),
+		// EVS_TIMEOUT
+		"600",
+		// [CONNSTR0|CONNSTR1|...]
+		"[" + strings.Join(connections, "|") + "]",
+	}
+}
+
+// parseTarget parses a single CLI forwarding target string of the form
+// [LOCAL_PORT:][METRO/]INSTANCE:DEST_PORT[/TYPE].
+func parseTarget(raw string) (Target, error) {
+	rest := raw
+	network := "tcp"
+
+	// Split connection type from the last "/" — but only if the suffix looks
+	// like a protocol name rather than a metro/instance path component.
+	if idx := strings.LastIndex(rest, "/"); idx >= 0 {
+		suffix := rest[idx+1:]
+		if strings.EqualFold(suffix, "tcp") || strings.EqualFold(suffix, "udp") {
+			network = strings.ToLower(suffix)
+			rest = rest[:idx]
+		}
+	}
+
+	if network != "tcp" {
+		return Target{}, fmt.Errorf("unsupported connection type %q: only tcp is supported", network)
+	}
+
+	segments := strings.SplitN(rest, ":", 3)
+	switch len(segments) {
+	case 2:
+		// INSTANCE:DEST_PORT — no local port override, use 0 for random port.
+		rport64, parseErr := strconv.ParseUint(segments[1], 10, 16)
+		if parseErr != nil {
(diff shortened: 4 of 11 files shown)


internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+		stderr: stdio.Stderr,
+	}
+	ready := make(chan struct{}, 1)
+	go func() {
+		if err := cr.controlUp(ctx, ready); err != nil {
+			log.G(ctx).Error().Err(err).Msg("control relay error")
+		}
+	}()
+	// Wait for the control relay to establish its connection before sending traffic.
+	<-ready
```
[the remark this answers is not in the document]
craciunoiuc: This was code imported, will skip the fix

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+
+	_, err = io.Copy(c.conn, rc)
+	if err != nil {
+		if !tunnelIsNetTimeoutError(err) {
+			log.G(ctx).Error().Err(err).Msg("failed to copy data from remote host to client")
+		}
+	} else {
+		// Remote closed the connection cleanly; return to close our side.
+		return
+	}
```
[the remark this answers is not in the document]
craciunoiuc: This was code imported, will skip the fix

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+
+	return r.up(ctx)
+}
+
+// tunnelGeneratePort returns the next sequential proxy port for single-port mode.
+func (cmd *InstancesTunnelCmd) tunnelGeneratePort(startPort uint16) uint16 {
+	defer func() { cmd.portIterator++ }()
+	return startPort + cmd.portIterator
+}
```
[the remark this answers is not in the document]
craciunoiuc: This was code imported, will skip the fix

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+		uuid := ptr.ZeroIfNil(inst.Uuid)
+
+		var fqdn string
+		if inst.ServiceGroup != nil && len(inst.ServiceGroup.Domains) > 0 {
+			fqdn = ptr.ZeroIfNil(inst.ServiceGroup.Domains[0].Fqdn)
+		}
+		if fqdn == "" {
+			return proxyInfo{}, fmt.Errorf("tunnel proxy has no service group domain")
+		}
+		return proxyInfo{uuid: uuid, fqdn: fqdn}, nil
```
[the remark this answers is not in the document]
craciunoiuc: Can't happen

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+			rest = parts[0] + "/" + parts[1]
+			ctype = "tcp"
+		}
+	} else {
+		rest = parts[0]
+		ctype = "tcp"
+	}
+	if strings.ToLower(ctype) != "tcp" {
+		log.G(ctx).Warn().Msg("only TCP connections are supported at the moment")
+	}
```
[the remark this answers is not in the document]
craciunoiuc: Already rejected

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+			return fmt.Errorf("could not resolve instances: %w", err)
+		}
+		for i, r := range resolved {
+			inst := r.(Instance)
+			metro = inst.key.Metro
+			if len(inst.Instance.NetworkInterfaces) == 0 || inst.Instance.NetworkInterfaces[0].PrivateIp == nil {
+				return fmt.Errorf("instance %q has no private IP", namesToResolve[i])
+			}
+			cmd.instances[indexesToResolve[i]] = *inst.Instance.NetworkInterfaces[0].PrivateIp
+		}
```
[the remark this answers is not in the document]
craciunoiuc: Will restrict to single-metro

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,586 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+	}
+
+	if metro == "" {
+		return fmt.Errorf("could not determine target metro: include the metro prefix in the instance name (e.g. fra/my-instance:8080)")
+	}
```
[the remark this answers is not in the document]
craciunoiuc: Remove referring by fqdn and private IP

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,582 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+	// localPorts to listen on locally.
+	localPorts []uint16
+	// ctypes holds the connection type ("tcp"/"udp") per target.
+	ctypes []string
+	// instanceProxyPorts holds the port on the instance to forward to.
+	instanceProxyPorts []uint16
+	// exposedProxyPorts holds the port exposed by the tunnel service per target.
+	exposedProxyPorts []uint16
+	// portIterator tracks the next port offset when a single proxy port is given.
+	portIterator uint16
```
jedevc: :eyes: why are these all on this struct?

The only things that should be here are the args that can be read by kong.

internal/cmd/instances.go
jedevc, line 1: Can you split the tunnel implementation *out* to `internal/tunnel` or something? Similar to how we have `internal/logs`.

That simplifies this package, so that we ideally only have the resource plumbing bits here.

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,582 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
 }
+
+const tunnelDeprecatedImage = "official/utils/tunnel:latest"
+
+type InstancesTunnelCmd struct {
+	Targets []string `arg:"" name:"target" min:"1" help:"Forwarding target(s) in the form [LOCAL_PORT:]<INSTANCE|PRIVATE_IP|PRIVATE_FQDN>:DEST_PORT[/TYPE]."`
+
+	TunnelProxyPorts   []string `short:"p" name:"tunnel-proxy-port"   help:"Remote port(s) exposed by the tunnel service. When a single value is given it is used as the starting port for multiple targets. (default: 4444)"`
+	ProxyControlPort   uint     `short:"P" name:"tunnel-control-port" help:"Command-and-control port of the tunnel service." default:"4443"`
+	TunnelServiceImage string   `          name:"tunnel-image"         help:"Image to use for the tunnel service." default:"official/utils/tunnel:1.0"`
```
jedevc: Why would a user want to specify these?

Is this a common use case? If it's only for dev, we should hide these (and avoid validation for them, the dev can be trusted).

internal/cmd/instances.go
```diff
@@ -1286,3 +1295,582 @@ func stopInstances(ctx context.Context, g *group.Group[multimetro.MetroClient],
...
+	lAddr    string
+	rAddr    string
+	ctype    string
+	auth     string
+	name     string
+	nameAddr string
+	stderr   io.Writer
+}
+
+const tunnelHeartbeat = "\xf0\x9f\x91\x8b\xf0\x9f\x90\x92\x00"
```
jedevc: There's a lot of magic values / strings / things.

When you move this to a separate package can you add a link to the upstream code that implements the server side of this.

internal/cmd/instances.go
craciunoiuc: Hmm, sure I was wondering how these would sit

internal/cmd/instances.go
craciunoiuc: equivalent of passing them by arg everywhere

If I move to a separate package I can keep them there.

internal/cmd/instances.go
craciunoiuc: the tunnel service image is mostly for debugging hence no short flag

the other two Simon back then suggested we should make this configurable in case the client box blocks 4444/4443 for some reason

internal/cmd/instances.go
craciunoiuc: You mean an actual `x` package or just an internal package to place everything?

jedevc: Just an internal package.

internal/cmd/instances.go
jedevc: IMO, these kind of act like global variables. It's unclear where they get created / passed through. Passing them through as args would be better if we can.

If they grow to be very long, we can group them into structs.

craciunoiuc: sure, if i'm moving this to a separate package it will need them as arguments anyway

internal/cmd/instances.go
craciunoiuc: Made hidden

internal/cmd/instances.go
craciunoiuc: Moved to internal package

internal/cmd/instances.go
craciunoiuc: Moved

internal/cmd/instances.go
craciunoiuc: Ok, I simplified it as much as I could, let's see what Copilot says.

I think the exposed interface of the library is quite simple now:
```
New <- you create the instance
RunProxy <- start the proxying instance
TerminateProxy <- stop the proxying instance
StartRelay <- block and send messages
```

craciunoiuc:
I need to fix these + do one more pass so no rush

craciunoiuc:
Should be as decent as it gets, definitely an upgrade compared to kraftkit

craciunoiuc:
Have to rebase the code and to generate tests on Monday.

cmd/unikraft/instances_test.go
```diff
@@ -306,6 +325,142 @@ func instancesTests(t *testing.T, r *testRunner) {
...
+			"--set", "image=nginx:latest",
+			"--set", "autostart=true",
+			"--set", "resources.memory=128",
+			"--set", "resources.vcpus=1",
+		)
+		t.Cleanup(func() {
+			cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancel()
+			env.unikraftIgnoreErr(cleanupCtx, "instance", "delete", instanceName)
+		})
```
jedevc: We should use the same test style as in the rest of the tests, so we get a resulting golden file!

For managing a tunnel in the background - we should probably add some sort of way to run it in the background using the golden setup.

internal/cmd/instances.go
```diff
@@ -1459,3 +1462,119 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
 	})
 	return multimetro.Keys(suspended), err
 }
+
+type InstancesTunnelCmd struct {
+	Targets []string `arg:"" name:"target" min:"1" help:"Forwarding target(s) in the form [LOCAL_PORT:]<INSTANCE>:DEST_PORT[/TYPE]."`
```
jedevc: Use `placeholder` to show the form. See other examples of how this is done.

internal/cmd/instances.go
```diff
@@ -1459,3 +1462,119 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
...
+		{
+			Description: "Use a custom relay port to avoid collisions",
+			Commands:    []string{"unikraft instance tunnel -p 5500 my-instance:8080"},
+		},
+	}
+}
+
+func (cmd *InstancesTunnelCmd) Run(ctx context.Context, stdio config.Stdio) error {
+	if len(cmd.TunnelProxyPorts) == 0 {
+		cmd.TunnelProxyPorts = []string{"4444"}
```
jedevc: Can we not use the `default` tag?

internal/cmd/instances.go
```diff
@@ -1459,3 +1462,119 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
...
+	}
+	if len(namesToResolve) > 0 {
+		resolved, err := Instance{}.Get(ctx, namesToResolve)
+		if err != nil {
+			return nil, "", fmt.Errorf("could not resolve instances: %w", err)
+		}
+		for i, r := range resolved {
+			inst := r.(Instance)
+			if metro != "" && inst.key.Metro != "" && inst.key.Metro != metro {
+				return nil, "", fmt.Errorf("forwarding targets must be in the same metro: found %q and %q", metro, inst.key.Metro)
```
jedevc: 🤔 we could support instances across multiple metros. We would just need to create multiple tunnels. This is supported for the `logs` cmd, so makes sense to support it here.

Instead of resolving all the metros, we'd do a call more like `CollectRefs`, and use that to resolve all the ips, and metros, and then just do multiple tunnel creations!

internal/cmd/instances.go
```diff
@@ -1459,3 +1462,119 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
...
+	}()
+
+	return tun.StartRelay(ctx, sgFQDN, resolvedTargets, instUUID, cmd.ProxyControlPort, stdio.Stderr)
+}
+
+// tunnelLookupInstances resolves the instance names in the tunnel configuration
+// to their private IPs, ensuring that all instances are in the same metro and
+// returning the metro name for later use.
+// NOTE(craciunoiuc): We do this here to have access to the Instances API
+func (cmd *InstancesTunnelCmd) tunnelLookupInstances(ctx context.Context, t *tunnel.Tunnel) ([]string, string, error) {
```
jedevc: This function could very easily be added to the tunnel package actually.

Yes, we'd have to do the `GetInstances` request directly there, but then it's a reusable package, for anyone to use. Also, we already are doing `CreateInstance` requests in there, so.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+	"strings"
+
+	"unikraft.com/cli/internal/multimetro"
+	"unikraft.com/cloud/sdk/platform"
+	"unikraft.com/cloud/sdk/platform/group"
+	"unikraft.com/x/log"
+	"unikraft.com/x/ptr"
+)
+
+type Tunnel struct {
```
jedevc: `struct` here shouldn't be public - see other comment about being able to move the ip + metro resolution here.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+	// Targets holds the target instance identifiers; entries are replaced by
+	// their private IPs once resolved.
+	Targets []string
```
jedevc: I really want to avoid this pattern of "populate this opts struct" from lots of different places, or changing them once they're set.

1. We should do the actual parsing of the `targets` string in `internal/cmd`. This makes this package more reusable. It should ideally produce something like
	```go
	type Target struct {
		Host string
		Source uint16
		Dest uint16
		Network string // tcp/udp (like in net.Dial)
	}
	```
	We can then also add tests for parsing. This simplifies the `Tunnel` struct, so instead of needing to have lots of `[]string`s and `[]uint16`s, we just have *one* `[]Target` (which then has all that info).
2. Clean up the `portIterator`. Instead of determining them like this, we should be able to determine this based on the index of the instance in that metro + some base port.
3. Then on `New`, we can do any initialization from these targets - but we should do it all in `New`, and then we never write to to the Tunnel again.

I know this is just an import from kraftkit right now, but we should take the opportunity to tidy it up.

internal/cmd/instances.go
```diff
@@ -1459,3 +1462,119 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
...
+		return fmt.Errorf("could not start tunnel proxy: %w", err)
+	}
+
+	defer func() {
+		if err := tun.TerminateProxy(context.WithoutCancel(ctx), g, metro, instUUID); err != nil {
+			log.G(ctx).Error().Err(err).Msg("could not terminate tunnel proxy")
+		}
+	}()
+
+	return tun.StartRelay(ctx, sgFQDN, resolvedTargets, instUUID, cmd.ProxyControlPort, stdio.Stderr)
```
jedevc: These details should be entirely moved to the `tunnel` package as well. So you just call `tun.Start()`, and then call `tun.Stop()` when you want to close everything up (which is when `<-ctx.Done` is finished).

This keeps the package nice and reusable while reducing it's API surface.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+		Image:    &image,
+		MemoryMb: &memMB,
+		Args:     args,
+		ServiceGroup: &platform.CreateInstanceRequestServiceGroup{
+			Services: services,
+		},
+		Autostart: &autostart,
+		TimeoutS:  &timeoutS,
+		Features:  []platform.CreateInstanceRequestFeatures{platform.CreateInstanceRequestFeatures("delete-on-stop")},
+	}
```
jedevc: Use `new(int64(128))` to avoid needing to create `memMB` and similar.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+			stderr:   stderr,
+		}
+		go func() {
+			if err := pr.up(ctx); err != nil {
+				log.G(ctx).Error().Err(err).Msg("relay error")
+			}
+		}()
+	}
+
+	return r.up(ctx)
```
jedevc: 🤔 Confused by why we don't just have one loop here?

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+	// Start the control relay to keep the tunnel connection alive.
+	cr := tunnelRelay{
+		rAddr:  net.JoinHostPort(sgFQDN, strconv.FormatUint(uint64(proxyControlPort), 10)),
+		auth:   t.auth,
+		stderr: stderr,
+	}
+	ready := make(chan struct{}, 1)
+	go func() {
+		if err := cr.controlUp(ctx, ready); err != nil {
+			log.G(ctx).Error().Err(err).Msg("control relay error")
```
jedevc: If this fails, is this not fatal?

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+		if err != nil {
+			return fmt.Errorf("deleting proxy instance %q: %w", uuid, err)
+		}
+		return nil
+	})
+}
+
+// genRandAuth generates a 32-character random alphanumeric string used to
+// authenticate connections to the tunnel service.
+func genRandAuth() (string, error) {
```
jedevc: Oop, let's find a way to share this with the `volimport` package.

internal/tunnel/relay.go
```diff
@@ -0,0 +1,102 @@
...
+// NOTE(craciunoiuc): Protocol and heartbeat encoding can be found at:
+// https://github.com/unikraft-cloud/catalog/tree/prod-staging/library/official/utils/tunlr/1.0
+type tunnelRelay struct {
+	lAddr    string
+	rAddr    string
+	ctype    string
+	auth     string
+	name     string
+	nameAddr string
+	stderr   io.Writer
```
jedevc: nit: `lAddr`/`rAddr`/`ctype` are moderately unclear. Maybe some clearer names?

internal/tunnel/relay.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// up starts a local listener and relays accepted connections to the remote host.
+func (r *tunnelRelay) up(ctx context.Context) error {
+	l, err := r.listenLocal(ctx)
+	if err != nil {
+		return err
+	}
+	defer l.Close()
+	go func() { <-ctx.Done(); l.Close() }()
```
jedevc: `l.Close()` can get called twice?

internal/tunnel/relay.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (r *tunnelRelay) dialRemote(ctx context.Context) (net.Conn, error) {
+	var d tls.Dialer
+	return d.DialContext(ctx, "tcp4", r.rAddr)
+}
+
+func (r *tunnelRelay) listenLocal(ctx context.Context) (net.Listener, error) {
+	var lc net.ListenConfig
+	return lc.Listen(ctx, r.ctype+"4", r.lAddr)
+}
```
jedevc: We can inline this right?

internal/tunnel/relay.go
```diff
@@ -0,0 +1,102 @@
...
+
+// controlUp dials the remote control port, signals ready, then sends periodic
+// heartbeats to keep the tunnel service alive.
+func (r *tunnelRelay) controlUp(ctx context.Context, ready chan struct{}) error {
+	rc, err := r.dialRemote(ctx)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+	go func() { <-ctx.Done(); rc.Close() }()
```
jedevc: Again, closed twice?

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+	<-ready
+
+	r := tunnelRelay{
+		// TODO(antoineco): allow dual-stack by creating two separate listeners.
+		// Alternatively, we could default to "::" to create a tcp46 socket, but
+		// listening on all addresses is an insecure default.
+		lAddr: net.JoinHostPort("127.0.0.1", strconv.FormatUint(uint64(t.localPorts[0]), 10)),
+		rAddr: net.JoinHostPort(sgFQDN, strconv.FormatUint(uint64(t.exposedProxyPorts[0]), 10)),
+		// NOTE(craciunoiuc): Only TCP is supported at the moment. This refers to the
+		// local listener; the remote side always uses TLS-over-TCP.
```
jedevc: 🤔 not for now (probably), but don't we support UDP services? (like as of this week?)

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+
+// parsePorts parses a single forwarding target of the form
+// [LOCAL_PORT:]INSTANCE:DEST_PORT[/TYPE] and returns the components.
+func parsePorts(ctx context.Context, portsArg string) (instance string, lport, rport uint16, ctype string, err error) {
+	var rest string
+	parts := strings.SplitN(portsArg, "/", 3)
+	if len(parts) == 3 {
+		rest = parts[0] + "/" + parts[1]
+		ctype = parts[2]
+	} else if len(parts) == 2 {
```
jedevc: :eyes: what? What is this case for? What if there are 4 `/`s here?

We could simplify this by just splitting at the _last_ `/`.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,449 @@
...
+	case 2:
+		// INSTANCE:DEST_PORT — no local port override
+		if _, parseErr := strconv.ParseUint(segments[0], 10, 16); parseErr == nil {
+			return "", 0, 0, "", fmt.Errorf("%q is not a valid instance identifier", segments[0])
+		}
+		rport64, parseErr := strconv.ParseUint(segments[1], 10, 16)
+		if parseErr != nil {
+			return "", 0, 0, "", fmt.Errorf("%q is not a valid port number", segments[1])
+		}
+		return segments[0], uint16(rport64), uint16(rport64), ctype, nil
```
jedevc: If no local port - shouldn't we select it to a random available port? (by setting it to 0)

jedevc (commented):
Needs a rebase, and to get the tests passing :tada:

cmd/unikraft/instances_test.go
craciunoiuc: Hmm, I don't have much context on that but I can try

internal/cmd/instances.go
craciunoiuc: What are you referring to? you mean for the images? you can just set a different image

internal/cmd/instances.go
craciunoiuc: Sounds crazy, but doable, the command has support for multiple tunnel connections

internal/cmd/instances.go
craciunoiuc: Sure, was conflicted about that

internal/cmd/instances.go
craciunoiuc: hmm, I thought 3 exposed functions would make sense: Create/Start/Stop

so you can wait a bit and maybe do other stuff before connecting

internal/tunnel/relay.go
craciunoiuc: I could just call them `localAddr`, `remoteAddr`, and `connectionType`

internal/tunnel/relay.go
craciunoiuc: Whoops

but yeah, I think it's fine to call it twice

internal/tunnel/relay.go
craciunoiuc: Sure

internal/tunnel/relay.go
craciunoiuc: :sob: 

I though I caught all the places where this happened

internal/tunnel/tunnel.go
craciunoiuc: Not in the tunnel unfortunately, it need C code

internal/tunnel/tunnel.go
craciunoiuc: I believe the format is now possibly `metro/instance/connectionType`

Comment is outdated needs an update

internal/tunnel/tunnel.go
craciunoiuc: I think it sets it to the default value?

internal/tunnel/tunnel.go
craciunoiuc: True, but the instance needs cleaning afterwards

I can potentially move the cleanup also here, but it's extra code

internal/tunnel/tunnel.go
craciunoiuc: What do you mean by one loop?

It basically makes the main thread block and starts other tunnels in parallel

internal/tunnel/tunnel.go
craciunoiuc: 1. Sure
2. True
3. I guess

I'm okay with changes, I more or less just imported and made it work fine in the CLI so I'm open to suggestions

craciunoiuc:
I'll rebase and fix afterwards

internal/tunnel/tunnel.go
jedevc: It seems to be just using `rport64` for both?

internal/cmd/instances.go
jedevc: I mean in the kong command struct.

internal/cmd/instances.go
jedevc: That's fair, but let's just keep them simply named in that case, `Create`/`Start`/`Stop` as you suggest.

internal/tunnel/tunnel.go
jedevc: But why not use an errorgroup?

internal/cmd/instances.go
craciunoiuc: Ah sorry, probably some old kraftkit copy paste

internal/tunnel/tunnel.go
craciunoiuc: I guess that's one way to do it 😅 

Sure

cmd/unikraft/instances_test.go
craciunoiuc: Done, tidied up a bit, hopefully good

internal/cmd/instances.go
craciunoiuc: Resolving:

New comment below agreed on:
* New
* Run
* Terminate
* Start

internal/tunnel/tunnel.go
craciunoiuc: I think we can just import from `volimport`. If you find this heretical lemme know

craciunoiuc:
once you agree with the changes, I will fixup the `fixup`, and regenerate the tests once more

I am pretty sure I covered all your requests. It now even has multi-metro support

The API is now: `New -> Run -> Start -> Terminate` Lemme know if you want me to reword

I tested that the usual + new stuff still works

internal/cmd/instances.go
```diff
@@ -1625,3 +1628,82 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
 	})
 	return multimetro.Keys(suspended), err
 }
+
+type InstancesTunnelCmd struct {
+	Targets []string `arg:"" name:"target" min:"1" help:"Forwarding target(s) in the form [LOCAL_PORT:]<INSTANCE>:DEST_PORT[/TYPE]." placeholder:"<target>"`
+
+	TunnelProxyPorts   []string `short:"p" name:"tunnel-proxy-port"   help:"Remote port(s) exposed by the tunnel service. When a single value is given it is used as the starting port for multiple targets." default:"4444" hidden:""`
+	ProxyControlPort   uint     `short:"P" name:"tunnel-control-port" help:"Command-and-control port of the tunnel service." default:"4443" hidden:""`
+	TunnelServiceImage string   `          name:"tunnel-image"        help:"Image to use for the tunnel service." default:"official/utils/tunnel:1.0" hidden:""`
```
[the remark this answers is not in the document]
craciunoiuc: comments are wrong

internal/integration/env.go
```diff
@@ -121,3 +124,81 @@ func (env *TestEnv) Run(t *testing.T, args []string, opts ...CmdOption) string {
...
+		return strings.HasPrefix(s, "UNIKRAFT_")
+	})
+	c.Env = append(c.Env, "NO_COLOR=1")
+	c.Env = append(c.Env, "UNIKRAFT_CONFIG="+env.configPath)
+	c.Env = append(c.Env, "BUILDKIT_PROGRESS=quiet")
+	c.Env = append(c.Env, resource.UnikraftSandboxEnv+"="+env.sandboxPath)
+	c.Cancel = func() error {
+		return c.Process.Signal(os.Interrupt)
+	}
+	c.WaitDelay = 30 * time.Second
```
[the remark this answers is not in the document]
craciunoiuc: dump to buffer and log on error

cmd/unikraft/integration/instance_test.go
```diff
@@ -497,4 +497,49 @@ func TestInstances(t *testing.T) {
...
+
+		// Use metro/instance:port/tcp syntax for the tunnel target.
+		tunnel := r.StartBackground(t,
+			[]string{"unikraft", "instance", "tunnel", "18081:" + r.Config.MetroName + "/test-" + instName + ":8080/tcp"},
+			"127.0.0.1:18081",
+			0,
+		)
+
+		// Verify that we can reach the instance through the tunnel.
+		out := r.Run(t, []string{
```
jedevc, line 838: Use the `HTTPGet` helper.

cmd/unikraft/integration/instance_test.go
```diff
@@ -497,4 +497,49 @@ func TestInstances(t *testing.T) {
...
+			"--output", "/dev/null",
+			"--write-out", "HTTP %{http_code} OK\n",
+			"--retry", "5",
+			"--retry-delay", "2",
+			"--retry-all-errors",
+			"--connect-timeout", "5",
+			"--max-time", "10",
+			"http://127.0.0.1:18081",
+		})
+		assert.Regexp(t, `HTTP 200 OK`, out)
```
jedevc: Not just this, but we should make sure that the response is actually something from the instance.

Idea: host an inline ROM in the instance, then assert that that's the thing being served by nginx.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+	// Populated by Run, consumed by Start and Terminate.
+	proxies map[string]proxyInfo
+}
+
+// New creates a tunnel from the raw CLI target strings. It parses each target,
+// assigns proxy ports, and resolves instance names to private IPs via the
+// platform API. proxyPorts configures which ports the tunnel service exposes:
+// if a single port is given it is used as the starting port for sequential
+// assignment; otherwise there must be exactly one port per target.
+func New(ctx context.Context, rawTargets []string, proxyPorts []string) (*tunnel, error) {
```
jedevc: Lift the parsing out to `instances.go`. Or, have `internal/tunnel/parse.go` with parsing helpers. But regardless, we should be passing in fully parsed results into here.

Essentially: we should be passing in `targets`.

jedevc: We shouldn't return private types from a public function if we can avoid it.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+		return fmt.Errorf("could not resolve instances: %w", err)
+	}
+
+	// NOTE(craciunoiuc): if the same instance name appears in multiple metros
+	// (no metro prefix given), the last-seen result wins. Users should include
+	// a metro prefix to disambiguate (e.g. fra/my-instance:8080).
+	infoByDisplay := make(map[string]instanceInfo, len(resolved))
+	for _, r := range resolved {
+		infoByDisplay[r.ref.Display] = r.info
+	}
```
jedevc, line 144: This is fine for now, but maybe we should have a helper `CollectRefsMap` or something.

Then we could use the same logic for logs.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+// Start starts the tunnel relays. For each metro, a control relay is
+// established to the proxy in that metro, then data relays are started for
+// every target in that metro.
+func (t *tunnel) Start(ctx context.Context, proxyControlPort uint, isInsecure func(string) bool, stderr io.Writer) error {
+	// Start a control relay per metro proxy.
+	for metro, proxy := range t.proxies {
+		cr := tunnelRelay{
+			remoteAddr: net.JoinHostPort(proxy.fqdn, strconv.FormatUint(uint64(proxyControlPort), 10)),
+			auth:       t.auth,
+			insecure:   isInsecure(metro),
```
jedevc: Why can't we derive insecure based on the metro? We've resolved all the instances, so, we know the metros. I don't think it needs to be an arg...

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+			stderr:     stderr,
```
jedevc: Why do we need `stderr`?

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+	image := tunnelImage
+	req := platform.CreateInstanceRequest{
+		Image:    &image,
+		MemoryMb: new(int64(128)),
+		Args:     args,
+		ServiceGroup: &platform.CreateInstanceRequestServiceGroup{
+			Services: services,
+		},
+		Autostart: new(bool(true)),
+		TimeoutS:  new(int64(3)),
```
jedevc: Why not `-1`?

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+		eg.Go(func() error {
+			return r.up(ctx)
+		})
+	}
+
+	return eg.Wait()
+}
+
+// Terminate deletes all tunnel proxy instances across all metros.
+func (t *tunnel) Terminate(ctx context.Context, g *group.Group[multimetro.MetroClient]) error {
```
jedevc: Maybe `Close()`?

Also, note that we currently call this twice - once in `Run` and once in `InstancesTunnelCommand`.

Also! We should gracefully handle the instance already being deleted by delete-on-stop.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,524 @@
...
+			return fmt.Errorf("creating proxy in metro %q: %w", metro, err)
+		}
+		t.proxies[metro] = info
+	}
+	return nil
+}
+
+// createProxy creates a single tunnel proxy instance in the given metro for
+// the provided targets.
+func createProxy(ctx context.Context, g *group.Group[multimetro.MetroClient], metro string, targets []target, authStr string, proxyControlPort uint, tunnelImage string) (proxyInfo, error) {
```
jedevc: Move this and associated format functions into a separate file.

internal/cmd/instances.go
```diff
@@ -1625,3 +1628,82 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
...
+	}
+
+	g, err := multimetro.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	if err := tun.Run(ctx, g, cmd.ProxyControlPort, cmd.TunnelServiceImage); err != nil {
+		return fmt.Errorf("could not start tunnel proxy: %w", err)
+	}
```
jedevc: This API is still a little odd.

Because, what if I call `Start` without calling `Run`? Is there a reason to not collapse `Start` + `Run` together so that it's not possible to misuse like this?

cmd/unikraft/integration/instance_test.go
craciunoiuc: We can just check the default message, yeah

internal/tunnel/tunnel.go
craciunoiuc: Yeah, no idea why this was private

craciunoiuc: Did both 🙏

internal/tunnel/tunnel.go
craciunoiuc: Vestigial content, I'll remove

internal/tunnel/tunnel.go
craciunoiuc: Vestigial also

internal/tunnel/tunnel.go
craciunoiuc: I'll move all to `parse.go`

internal/tunnel/tunnel.go
craciunoiuc: Ok, I'll call it close

Also I think I'll add a `closed` per tunnel and we use that to make sure subsequent calls do not break

internal/cmd/instances.go
craciunoiuc: I think it made sense before but not so much now? I'll merge them definitely

internal/tunnel/tunnel.go
craciunoiuc: Ok, leaving like this. Feel free to open a linear issue and we can do in a separate PR 😳

craciunoiuc:
@khatibomar can you help review this to revive it?

Ideally we would get it in this cycle

khatibomar:
@craciunoiuc on it

khatibomar:
can you rebase it please @craciunoiuc

khatibomar (dismissed):
```
Reviewed-by: Omar El Khatib omar@unikraft.com
Approved-by: Omar El Khatib omar@unikraft.com
```

internal/integration/env.go
```diff
@@ -146,3 +148,90 @@ func (env *TestEnv) Run(t *testing.T, args []string, opts ...CmdOption) string {
...
+	return proc
+}
+
+// StopBackground terminates a background process immediately, without waiting
+// for the test to finish.
+func (env *TestEnv) StopBackground(t *testing.T, proc *BackgroundProcess) {
+	t.Helper()
+	t.Logf("stopping background process")
+	proc.stop()
+}
```
jedevc: Why not just `proc.Stop`.

internal/cmd/instances.go
```diff
@@ -1957,3 +1961,79 @@ func suspendInstances(ctx context.Context, g *group.Group[multimetro.MetroClient
...
+	g, err := multimetro.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		// Strip the parent context's cancellation (so shutdown isn't aborted
+		// by the same Ctrl-C that triggered it) but bound it, so a hung or
+		// unresponsive API call can't leave the process hanging forever with
+		// an orphaned, still-billed proxy instance.
```
jedevc: big comment.

internal/tunnel/parse.go
```diff
@@ -0,0 +1,226 @@
...
+	// closed indicates whether this proxy instance has been closed.
+	closed bool
+}
+
+// ParseTargets parses raw target strings and proxy port specifications into
+// structured target objects.
+// proxyPorts configures which ports the tunnel service exposes:
+// if a single port is given it is used as the starting port for sequential
+// assignment; otherwise there must be exactly one port per target.
+func ParseTargets(rawTargets []string, proxyPorts []string) ([]target, error) {
```
jedevc: Don't return private structs from a public method - it's just got an icky vibe to it, you can make `target` public, and just keep the fields private.

internal/tunnel/tunnel.go
jedevc: This still applies. Can you do a follow-up?

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,388 @@
...
+		infoByDisplay[r.ref.Display] = r.info
+	}
+
+	for i := range t.Targets {
+		info, ok := infoByDisplay[t.Targets[i].host]
+		if !ok {
+			return fmt.Errorf("could not determine metro for %q: include the metro prefix (e.g. fra/my-instance:8080)", t.Targets[i].host)
+		}
+		t.Targets[i].host = info.ip
+		t.Targets[i].metro = info.metro
```
jedevc: I *swear* I mentioned this before, but this is a really odd pattern. We're writing back to the args that were passed in and overwriting the inputs which isn't great.

I still don't love the separate `resolve` method, and this weird way in which this struct can be in two states, resolved and "not resolved". IMO, would be better to create a `resolvedTarget` if we are doing it this way.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,388 @@
...
+	network string
+	// exposedProxyPort is the port exposed by the tunnel service for this
+	// target. It is computed from the proxy port configuration.
+	exposedProxyPort uint16
+	// metro is the metro where this target's instance lives, set by resolve.
+	metro string
+}
+
+type Tunnel struct {
+	Targets []target
```
jedevc: Again, public property, private type, icky.

internal/tunnel/tunnel.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	return errors.Join(errs...)
+}
+
+// tunnelConnection represents an accepted local connection being relayed to a
+// remote host through the tunnel service.
+type tunnelConnection struct {
+	relay *tunnelRelay
+	conn  net.Conn
+}
```
jedevc: Move to `conn.go`, since relay is in a separate `relay.go`. It's weird for reading to jump from `tunnel.go` to `relay.go` back to `tunnel.go` for the connection info.

internal/integration/env.go
```diff
@@ -146,3 +148,88 @@ func (env *TestEnv) Run(t *testing.T, args []string, opts ...CmdOption) string {
...
+			p.t.Logf("background process output:\n%s", p.output.String())
+		}
+	})
+}
+
+// StartBackground starts a command in the background and registers cleanup to
+// stop it when the test ends. If waitAddr is non-empty, it polls that TCP
+// address until it is reachable or the timeout (default 90s) expires.
+// The returned handle's Stop method can be called for early termination.
+func (env *TestEnv) StartBackground(t *testing.T, args []string, waitAddr string, timeout time.Duration) *BackgroundProcess {
```
jedevc, line 180: Follow-up, not needed here: but we should split this out to a separate helper `env.WaitAddr`.

jedevc (approved):
```
```

PR #249, opened 2026-04-07
fix: Make install.sh compatible with sh

jedevc:
So that we don't need bash installed, which isn't available everywhere. Also it's just nice practice.

Additionally, run shellcheck and fix any resulting issues.


nderjung (approved):
Thanks!

PR #254, opened 2026-04-07
fix: Ensure that logs don't rewind back to the very beginning

jedevc:
Using `--tail` with logs could sometimes rewind and produce bizarre results.

Imagine a page size of 1000, and a log size of 2500. We would previously produce the follow sequence of requests:
- offset=-1000, limit=1000 (returns back start=1500)
- offset=500, limit=1000
- offset=-500, limit=1000 (returns back start=1500)
- weird errors, and strangeness occurs

The issue is that we need to recognize when we accidentally rewind behind the start, and distinguish that from when we don't *know* the start, and so need to use a negative offset.

Additionally, add a sanity check for the platform, that ensures that it definitely does return the expected amount of data!

### DIFF
--- a/internal/logs/logs.go
+++ b/internal/logs/logs.go
@@ -79,13 +79,17 @@ func (r *logsReader) rewind(tail int) (n int64, err error) {
 
 	total := 0
 
-	var chunkOffset int64 = 0
 	chunk := make([]byte, logPageSize)
+	chunkOffset := ptr.ZeroIfNil(r.end)
 	for {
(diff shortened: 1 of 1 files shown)


jedevc:
The original issue that I saw lol:

```
❯ unikraft instance logs cloud-console-pr-633-cad9a --tail 500
│  
│ error:
│   failed on "stable" client: could not rewind logs: incomplete chunk read at
│ offset 140757
│  
```

nderjung (approved):
Thanks!

PR #256, opened 2026-04-08
feat: Use new /v1/auth API to fetch organization details

jedevc:
Fixes https://linear.app/unikraft/issue/TOOL-716/populate-cli-organization-from-controlplane-if-not-specified-with.

If the user doesn't specify an `--organization` with their `--token`, we
would previously fail out. We do need to know the user for various
operations.

This patch communicates with the controlplane to determine the organization
from the token if it's not specified using the new `GET /v1/auth` API.


razvand:
I tested this, and it works :+1:

nderjung (approved):
Thanks!

PR #261, opened 2026-04-09
feat: Allow selecting via TUI between profiles without arg

nderjung:
I became frustrated at typing things

### DIFF
--- a/internal/cmd/profile.go
+++ b/internal/cmd/profile.go
@@ -8,12 +8,15 @@ package cmd
 import (
 	"cmp"
 	"context"
+	"errors"
+	"maps"
 	"slices"
 
 	jujuerrors "github.com/juju/errors"
 	"unikraft.com/cli/internal/config"
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
+	"unikraft.com/cli/internal/tui/selector"
 	"unikraft.com/x/kingkong"
 	"unikraft.com/x/log"
 )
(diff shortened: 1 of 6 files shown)


internal/cmd/profile.go
```diff
@@ -98,22 +100,42 @@ func (Profile) Examples() map[cmd.CmdType][]kingkong.Example {
...
-	if !ok {
-		return config.ErrProfileNotFound{Name: cmd.Name}
+	name := cmd.Name
+
+	if name == "" {
+		names := make([]profileName, 0, len(cfg.Profiles))
+		for k := range cfg.Profiles {
+			names = append(names, profileName(k))
+		}
+		sort.Slice(names, func(i, j int) bool { return names[i] < names[j] })
```
jedevc: Can also just use `maps.Keys` as well for a nice one liner.

internal/tui/selector/single.go
```diff
@@ -0,0 +1,141 @@
...
+			out += colors.InfoFg("]")
+		}
+
+		out += "\n"
+	}
+
+	out += colors.InfoFgBg(" ") + "\n"
+
+	if !m.quitting {
+		out += colors.InfoFgBg(" ") + " use arrow keys to navigate; enter to select; esc to cancel"
```
jedevc: :eyes: Can we use `charm.land/bubbles/v2/help` for consistency with the TUI. Instead of writing our own help format here.

internal/cmd/profile.go
```diff
@@ -98,22 +100,41 @@ func (Profile) Examples() map[cmd.CmdType][]kingkong.Example {
...
 func (cmd *UseCmd) Run(ctx context.Context, cfg *config.Config) error {
-	_, ok := cfg.Profiles[cmd.Name]
-	if !ok {
-		return config.ErrProfileNotFound{Name: cmd.Name}
+	name := cmd.Name
+
+	if name == "" {
+		names := make([]profileName, 0, len(cfg.Profiles))
+		for _, k := range slices.Sorted(maps.Keys(cfg.Profiles)) {
+			names = append(names, profileName(k))
```
jedevc: That's annoying that we need the wrapper for that.

Can we just make `SingleWithDefault` not generic and just directly take strings? Or at least change the constraint to be `[~string]`.

internal/tui/selector/single.go
```diff
@@ -0,0 +1,372 @@
...
+		case "down":
+			visible := m.filtered()
+			if m.cursor+1 < len(visible) {
+				m.cursor++
+			}
+			return nil
+		default:
+			// Append printable runes to the filter.
+			for _, r := range msg.String() {
+				if r >= 32 && r != 127 {
```
jedevc: `unicode.IsPrint` 🙏

jedevc: Actually, we could switch to using the bubbles https://github.com/charmbracelet/bubbles#text-input.

It has readline-like key combinations.

jedevc (approved):

PR #266, opened 2026-04-13
feat: Add instance templates

jedevc:
Implements https://linear.app/unikraft/issue/TOOL-544/implement-unikraft-instance-template.


internal/cmd/instances.go
```diff
@@ -860,9 +863,25 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 				req.AdditionalProperties = make(map[string]json.RawMessage)
 			}
 			req.AdditionalProperties["vsock"] = dt
+		case "template":
+			template := field.Create.Set.(string)
+			key := multimetro.ParseKey(template)
+			req.Template = &platform.CreateInstanceRequestTemplate{}
+			if key.UUID != "" {
```
craciunoiuc, line 871: Not sure where to write this, but should we validate inputs to make sure that someone doesn't specify an image (for example) when creating from a template? Same goes for ports, and I think there were others, can't remember.

Or just let the platform resolve those.

internal/cmd/instance_templates.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+func (c *InstanceTemplateEditCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *resource.Sandbox, kctx *kong.Context) error {
+	if err := cmd.ApplyShortcutFlags(&c.SetArgs, kctx.Flags()); err != nil {
+		return err
+	}
+	return c.ResourceEditCmd.Run(ctx, stdio, sandbox)
+}
+
+type InstanceTemplate struct {
```
craciunoiuc, line 76: Wondering if these fields are similar to the instance ones and could be reused, or are different?

craciunoiuc (commented):
Just some comments my side

internal/cmd/instances.go
jedevc: I've got a sort of ticket for this here: https://linear.app/unikraft/issue/TOOL-596/add-patch-validation.

But yes, I think it's better to defer to the platform for validation in pretty much all cases. We can add some in the future here, but this isn't the *only* place where we'd have this problem.

internal/cmd/instance_templates.go
jedevc: They are *very* similar. And cannot be reused :(

e.g. you can modify how much memory an instance is using. But not while it's a template.

There is a lot of duplication, I am working on cleaning it up so it's less miserable though. Shouldn't block this though.

jedevc:
Actually, got some bugs I found during manual testing. Don't merge just yet.

Depends on:
- https://github.com/unikraft-cloud/proto/pull/259

jedevc:
Should be good now!

craciunoiuc (approved):
All good here. Thanks!

PR #274, opened 2026-04-14
fix: Defer opening buildkit client until use

jedevc:
We should defer opening buildkit until right before we need to use it.

However, we add the context helpers so that we can still inject it for the purposes of testing.

e.g. with this code path https://github.com/unikraft/cli/pull/253, we don't need a buildkit, so we should avoid opening the client.


nderjung (approved):
Thanks!

PR #279, opened 2026-04-16
fix: Avoid useless patches

jedevc:
We actually weren't properly comparing image patches, so we were *always*
setting it on edit.

Before:

```
❯ unikraft instance edit nginx-fijep --cmd cat --dry-run
image := nginx
```

After:

```
❯ unikraft instance edit nginx-fijep --cmd cat --dry-run
```

The filtering change is so that the tests actually make sense :sweat:


craciunoiuc (approved):
All good here. Thanks!

PR #283, opened 2026-04-20
fix: Add instance delete fallback

jedevc:
On the old terrible public metros, `timeout_s` is not supported. However, there's no reliable way to know which features are supported on which metros, so we add a fallback, and if we fail, we just try again without the `timeout_s`.


craciunoiuc (approved):
Neat.

craciunoiuc:
I guess tests will pass after merging/rebasing?

PR #286, opened 2026-04-21
feat: Improved arg splitting

jedevc:
Fixes https://linear.app/unikraft/issue/TOOL-682/split-instance-args-in-cli.

When specified something like `--args "echo foo"`, this splits it into `["echo", "foo"]`. However, I noticed that this rendered weirdly, so updated the array-like rendering to use JSON-like formatting instead of CSV formatting, and updated the parser to support this as well, so we keep some level of consistency.

### DIFF
--- a/internal/resource/value/parse_test.go
+++ b/internal/resource/value/parse_test.go
@@ -115,6 +115,60 @@ func TestParseJSON(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, testStruct{Name: "hello", Value: 42}, got)
 	})
+
+	t.Run("slice with escaped quotes", func(t *testing.T) {
+		got, err := Parse[[]string]([]string{`["say \"hello\"", "world"]`})
+		require.NoError(t, err)
+		assert.Equal(t, []string{`say "hello"`, "world"}, got)
+	})
+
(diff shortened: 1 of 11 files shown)


internal/resource/value/parse_test.go
```diff
@@ -0,0 +1,186 @@
...
+		assert.Empty(t, got)
+	})
+
+	t.Run("map", func(t *testing.T) {
+		got, err := Parse[map[string]string]([]string{`{"a": "1", "b": "2"}`})
+		require.NoError(t, err)
+		assert.Equal(t, map[string]string{"a": "1", "b": "2"}, got)
+	})
+
+	t.Run("map with commas in values", func(t *testing.T) {
```
craciunoiuc, line 95: Can you add some cursed tests also?

Aka, some that have escaped characters, maaybe some escaped quotes/double quotes also.

craciunoiuc (commented):
Just one request my side 🥹

internal/resource/value/parse_test.go
jedevc: Done.

craciunoiuc (approved):
All good here. Thanks!

PR #290, opened 2026-04-23
feat: Use explicit kraftfile type if specified

jedevc:
Follow-up to https://github.com/unikraft/cli/pull/245#discussion_r3110958924

No need to do source type detection if the kraftfile has an explicit `type` - we should just use that.

Additionally, this adds some more tests to check that all the rootfs builds produce reasonable results.


internal/builder/rootfs.go
```diff
@@ -93,30 +93,28 @@ func DetectRootfsType(path string) (RootfsType, error) {
...
 	switch opts.Rootfs.Type {
-	case RootfsTypeCpio, RootfsTypeErofs:
+	case kraftfile.SourceTypeCpio, kraftfile.SourceTypeErofs:
+		if opts.Rootfs.Format != "" {
+			expected := map[kraftfile.SourceType]kraftfile.FsType{
+				kraftfile.SourceTypeCpio:  kraftfile.FsTypeCpio,
+				kraftfile.SourceTypeErofs: kraftfile.FsTypeErofs,
+			}
+			if exp, ok := expected[opts.Rootfs.Type]; ok && opts.Rootfs.Format != exp {
+				// TODO: maybe we could be smart and convert this
```
craciunoiuc, line 110: Theoretically we have readers/writes for them, so this means we could covert them. right? 👀

craciunoiuc (approved):
All good here. Thanks!

craciunoiuc:
Static check failed :(((

PR #292, opened 2026-04-23
fix: Use tarfs for dockerfile builds

jedevc:
Fixes https://linear.app/unikraft/issue/TOOL-791 and https://linear.app/unikraft/issue/TOOL-724.

Instead of using directories - directories are lossy: e.g. Mac OS uses a case-insensitive filesystem by default, we can't create device file types without root permissions, etc.

So, instead of using the filesystem as an intermediary, we use a tarball which represents all this much more accurately.

To do this, we use archivefs's tarfs package, which uses an fs.FS interface, so we can actually just abstract the packaging to consume that, allowing us to package up any input fs, without needing future changes. We consolidate the old erofs and cpio packages together, since they now share a lot more in common.

The updated archivefs also includes some other fixes, like allowing iteration over device file types, etc.


internal/builder/buildfs/cpio.go
```diff
@@ -3,37 +3,21 @@
...
-func WithAllRoot(allRoot bool) CpioCreateOption {
-	return func(co *createOptions) {
-		co.allRoot = allRoot
-	}
-}
-
-// CreateFSFromDirectory creates a CPIO filesystem from an existing directory.
-func CreateFSFromDirectory(ctx context.Context, w io.Writer, source string, opts ...CpioCreateOption) error {
+// CreateCPIO creates a CPIO filesystem from the given fs.FS.
+func CreateCPIO(ctx context.Context, w io.Writer, fsys fs.FS, opts ...CreateOption) error {
```
jedevc, line 20: suggestion: move this to `archivefs`, or `go-cpio`.

internal/builder/buildfs/tarball.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, err
+	}
+	defer gz.Close()
+
+	data, err := io.ReadAll(gz)
+	if err != nil {
+		return nil, err
+	}
+	return bytes.NewReader(data), nil
+}
```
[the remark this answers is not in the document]
jedevc: Indeed the robot is correct.

internal/builder/buildfs/erofs.go
```diff
@@ -0,0 +1,26 @@
...
+
+// CreateEROFS creates an EroFS filesystem image from the given fs.FS.
+func CreateEROFS(w io.WriterAt, fsys fs.FS, opts ...CreateOption) error {
+	c := createOptions{}
+	for _, opt := range opts {
+		opt(&c)
+	}
+	var erofsOpts []erofs.ErofsCreateOption
+	if c.allRoot {
+		erofsOpts = append(erofsOpts, erofs.WithAllRoot(true))
```
craciunoiuc, line 23: we should readd the option to keep file owners

craciunoiuc (approved):
All good here. Thanks!

PR #298, opened 2026-04-27
fix(build): Always use cpio as default

jedevc:
Don't use erofs as the default in any circumstances. It should be applied globally, for all the build types, not just for directory sources.

Technically, this was already happening, since the parsed kraftfile always has a format set, but the code should be clearer here.


craciunoiuc (approved):
All good here. Thanks!

PR #303, opened 2026-04-28
feat: Add rom support

jedevc:
Allowing both building and creating instances with mounted ROMs.

Fixes https://linear.app/unikraft/issue/TOOL-695/rom-support-in-cli.

TODO:
- [x] Fix inline XXX
- [x] Add `file=./path/to/file` or something for mounting in files.
- [x] Needs https://github.com/unikraft-cloud/x/pull/243.


jedevc:
Grrr, copilot is right, I will look at this later :angry:

internal/builder/build.go
```diff
@@ -27,109 +28,146 @@ type BuildOpts struct {
...
-	// Buildkit params
-	// Dockerfile string
-	BuildArg []string
-	Target   string
-	Secrets  []*Secret
-	SSH      []*SSH
-
-	NoCache bool
+var DefaultPlatform = ocispec.Platform{
+	Architecture: "x86_64", OS: "kraftcloud",
```
craciunoiuc, line 55: Some years ago we were talking about phasing out the `OS: "kraftcloud"` and replacing with `ukc` or `unikraftcloud`.

I guess there's no easy way to set both and then quietly phase out the other right?

internal/builder/build.go
```diff
@@ -27,109 +28,146 @@ type BuildOpts struct {
...
+		if err != nil {
+			return nil, err
 		}
-
-		images := make([]*imagespec.Image, 0, len(kernels))
+		defer func() {
+			for _, kernel := range kernels {
+				kernel.Close()
+			}
+		}()
```
craciunoiuc, line 81: Can this be moved in the for below and then you would just add `defer kernel.Close()`

internal/builder/build.go
```diff
@@ -27,109 +28,146 @@ type BuildOpts struct {
...
+				root.Close()
+			}
+		}()
+	}
 
-	if len(kernels) != len(roots) {
+	// When we have both kernels and rootfs, they must match 1:1 by platform.
+	if len(kernels) > 0 && len(roots) > 0 && len(kernels) != len(roots) {
 		panic(fmt.Sprintf("internal error: number of kernels (%d) does not match number of root filesystems (%d)", len(kernels), len(roots)))
 	}
```
craciunoiuc, line 121: Why panic and not just log the error and exit?

internal/builder/build.go
```diff
@@ -27,109 +28,146 @@ type BuildOpts struct {
...
 
-		if platforms.Format(kernel.Image.Platform) != platforms.Format(root.Image.Platform) {
-			panic(fmt.Sprintf("internal error: kernel platform (%s) does not match rootfs platform (%s)",
-				platforms.Format(kernel.Image.Platform),
-				platforms.Format(root.Image.Platform)),
-			)
+		// Attach kernel if available.
+		if i < len(kernels) {
+			if platforms.Format(kernels[i].Image.Platform) != platforms.Format(p) {
+				panic(fmt.Sprintf("internal error: kernel platform (%s) does not match expected platform (%s)",
```
craciunoiuc: Same here

Is it a coding pattern to use `panic` in the cli?

internal/builder/kraftfile.go
```diff
@@ -53,6 +54,27 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
 		})
 	}
 
+	for _, rom := range kf.Roms {
+		romPath := filepath.Join(dir, rom.Source)
+		romFormat := cmp.Or(rom.Format, kraftfile.FsTypeErofs)
+		if romFormat != kraftfile.FsTypeErofs {
+			return BuildOpts{}, fmt.Errorf("rom format must be %s, not %s", kraftfile.FsTypeErofs, romFormat)
+		}
```
craciunoiuc: Roms can be also CPIO so this is not a mandatory check

Actually, they can be anything 💫 but only CPIO/EROFS are automounted I think

craciunoiuc: so in this case the `file` or `unknown` type would come in handy

I once mounted some tar archives

internal/builder/kraftfile.go
```diff
@@ -53,6 +54,27 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
...
+		romOpt := FSOpts{
+			Path:   romPath,
+			Format: romFormat,
+			Type:   rom.Type,
+		}
+		if romOpt.Type == "" {
+			typ, err := DetectSourceType(romPath)
+			if err != nil {
+				return BuildOpts{}, fmt.Errorf("detecting rom type for %q: %w", rom.Source, err)
```
craciunoiuc, line 71: Should we error out if we can't detect the type? Or should we just mount as a generic file? See above

internal/cmd/build.go
```diff
@@ -87,24 +87,24 @@ func (c *BuildCmd) Run(ctx context.Context, cfg *config.Config) error {
 	}
 
 	if len(c.BuildArg) > 0 {
-		buildOpts.Rootfs.BuildArg = append(buildOpts.Rootfs.BuildArg, c.BuildArg...)
+		buildOpts.BuildArg = append(buildOpts.BuildArg, c.BuildArg...)
 	}
 	if c.NoCache {
-		buildOpts.Rootfs.NoCache = true
+		buildOpts.NoCache = true
 	}
```
craciunoiuc: Time to flex my programming skills:

buildOpts.NoCache = c.NoCache

internal/cmd/instance_templates.go
```diff
@@ -292,24 +295,24 @@ func (InstanceTemplate) Edit(ctx context.Context, target resource.Resource, fiel
 	return results[0], nil
 }
 
-func instanceTemplatePatchSpec(path string, op patchOp, value any) (platform.UpdateTemplateInstancesRequestItemProp, any) {
+func instanceTemplatePatchSpec(path string, op patchOp, value any) (platform.UpdateTemplateInstancesRequestItemProp, any, error) {
```
craciunoiuc, line 298: Wait, you added an error return but all there are no errors and everything is nil?

internal/cmd/instances.go
```diff
@@ -330,6 +339,102 @@ func (v *InstanceVolume) UnmarshalText(data []byte) error {
...
+	parsed, err := value.Parse[alias]([]string{string(data)})
+	if err != nil {
+		return err
+	}
+	*r = InstanceRom(parsed)
+	return nil
+}
+
+// romNameFromAt derives a ROM name from a mount path by stripping leading
+// slashes and replacing path separators with hyphens.
```
craciunoiuc: This needs a debate? I would either:
* Use the `at` like here but only the `filepath.Base` + the hyphens trick for not accepted characters
* Use the source directory/file name + the project/instance name or smth

The current format might generate lengthy names that are not very relevant no? `myuser/var-lib-smth-myrom0:latest`

internal/cmd/instances.go
```diff
@@ -330,6 +339,102 @@ func (v *InstanceVolume) UnmarshalText(data []byte) error {
...
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+		if d.Type()&os.ModeSymlink != 0 {
+			return fmt.Errorf("symlinks are not supported in ROM directories: %s", path)
+		}
```
craciunoiuc: Why not? If after mounting the path is valid then who are we to judge?

craciunoiuc: But we should make sure not to follow the symlink

internal/cmd/instances.go
```diff
@@ -330,6 +339,102 @@ func (v *InstanceVolume) UnmarshalText(data []byte) error {
...
+	if len(files) == 0 {
+		return nil, fmt.Errorf("directory %q is empty", dir)
+	}
+	return files, nil
+}
+
+// inlineFileFromPath reads a single local file and returns it as a
+// base64-encoded InlineFile with the given destination path inside the ROM.
+func inlineFileFromPath(src string, destPath string) (platform.InlineFile, error) {
+	data, err := os.ReadFile(src)
```
craciunoiuc, line 419: What if the file is a symlink, this will follow it and read something else no? 🤔

internal/cmd/services.go
```diff
@@ -516,13 +519,13 @@ func (ServiceGroup) Examples() map[cmd.CmdType][]kingkong.Example {
 	}
 }
 
-func serviceGroupPatchSpec(path string, _ patchOp, value any) (platform.UpdateServiceGroupsRequestItemProp, any) {
+func serviceGroupPatchSpec(path string, _ patchOp, value any) (platform.UpdateServiceGroupsRequestItemProp, any, error) {
```
craciunoiuc, line 522: Same here + multiple places

craciunoiuc (changes requested):
Some comments my side 😳

internal/cmd/instance_templates.go
jedevc: Yes. This is because I changed the callback! This is used as a callback, so needs to be updated with the new func signature.

internal/builder/build.go
jedevc: We can change this, but this is such an unrecoverable error - it means that there is some horrible issue in the implementation of the Build function that these don't match. I found this was easier during debugging than getting an index panic instead.

internal/cmd/instance_templates.go
craciunoiuc: Oh, ok, sure, no problem

internal/builder/build.go
craciunoiuc: True, it's a design preference after all, but if this is rare to happen then I guess it's fine?

internal/builder/build.go
jedevc: Urgh. Not super easily, doing this requires agent changes. IMO, we should do all our image format updates at once, *once* the agent is consuming `image-spec`.

internal/builder/kraftfile.go
jedevc: `DetectSourceType` should do the handling for this. But I'll do this when we add a `file`/`unknown` type/etc as a valid `FsType` (which is out of scope for here imo).

internal/cmd/instances.go
jedevc: Well, they're going to lose their symlinkiness. I guess we shouldn't exclude symlinks specifically, but we should remove all non-regular files.

The inline file API *only* allows setting file contents. No other mode bits, etc.

internal/builder/kraftfile.go
jedevc: So I think this is a lie - if I try this with `CPIO`, it breaks:

```
$ unikraft instance logs test-$UNIQ_INST

stdout:
	test-<INST> ┏ [    0.000000] RDSEED32 is broken. Disabling the corresponding CPUID bit.
	test-<INST> │ symlink /dev/ukp_rom1 -> /dev/ukp_rom_myrom
	test-<INST> │ mount /dev/ukp_rom1 -> /rom
	test-<INST> │ [    0.000000] erofs: (device ukp_rom1): erofs_read_superblock: cannot find valid erofs superblock
	test-<INST> │ /rom: Failed to mount erofs: -22
	test-<INST> │ Unable to mount ukp_rom1 as EROFS: -22
	test-<INST> │ cat: can't open '/rom/hello.txt': No such file or directory
	test-<INST> │ Application exited with 0x100 (exit code: 1)
	test-<INST> │ [    0.000000] reboot: Restarting system
```

EROFS is the only one that's supported. How did you add a tar as a ROM?

internal/cmd/instances.go
jedevc: Okay, the problem is that each ROM name needs to be unique *within* the instance (they can clash outside the instance). So if we just do `filepath.Base`, `/foo/a/` and `/foo/b/` would collide. So the full path is the only way to do it.

internal/builder/build.go
jedevc: Each defer has a cost though. It's better to just have one defer instead of multiple.

internal/cmd/instances.go
craciunoiuc: Not necessarily, like in rootfses where we treat symlinks as files and read them as is without following

One example would be a ROM that contains symlinks to wellknown files like `/dev/stdout`. They would work on all our instances but we would ignore the file assuming it would break

craciunoiuc: > The inline file API only allows setting file contents. No other mode bits, etc.

Ooooh, so you mean you literally can't give it anything other than regular files

ok then, in that case it makes sense. I can't remember if it's different in regular roms, but I guess I trust you

internal/cmd/instances.go
craciunoiuc: sorry, I think I don't know my `filepath` functions

In your example I meant to use only `a` and `b` as names there, but yeah, even that is not safe I guess

I'm fine either way then

internal/builder/kraftfile.go
craciunoiuc: Sounds like something is set in the sdk/cli or the docs are broken :D

https://unikraft.com/docs/tutorials/rootfs-volumes-roms#roms
> they're attached on stopped instances, but aren't mounted automatically on boot as they can have any format (ext4, cpio, erofs, xfs, ntfs, etc)

I think that the cli is setting automount or something like that

craciunoiuc: Yes, it tries automounting if you set `at`

https://github.com/unikraft-cloud/go-sdk/blob/368d282a4a6dbef6a0971b9e319f45d489aa572d/platform/model_instance_rom.gen.go#L19-L24

So the check should be conditioned on that being set

craciunoiuc: Now the real question is why does the `mount` command assume erofs right from the start

craciunoiuc: It seems that I was wrong, you can't mount `CPIO` files at all. Probably I just remembered when I added it and unpacked in the instance the content

jedevc: Yeah the platform via firecracker rejects them.

jedevc: Are the docs wrong here?

craciunoiuc: Does it also reject them if you do not use `at`?

alternatively: does it work with an ext4 file?

jedevc: Aha, if `at` isn't specified it works. But, apparently all ROMs need to be padded to the page size, so I've had to add that in for CPIO, since they're not auto padded to that (erofs already is).

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #309, opened 2026-05-07
feat(cmd): Add rollout mechanics when creating

craciunoiuc:
~~Open to UI/UX suggestions, right now it prints every new instance that was rolled~~
Completely reworked, now, as discussed below, added extra:
* Two modes, `rolling/replace`, one starting + deleting, one stopping + starting + deleting
* A configurable wait time to recheck that state is ok `--rollout-healthy-after`
* Made only for instances in services

Notable ickies still open to suggestions:
* Code needed to be added in the cmd resource to make sure that fields are accessed only once instead of twice
* Not quite so plug-and-play: I split things in a rollout package, but still a lot of helpers were needed in the `instance.go` itself

Closes: TOOL-794

### DIFF
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -41,6 +41,7 @@ import (
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
 	"unikraft.com/cli/internal/resource/value"
+	"unikraft.com/cli/internal/rollout"
 	"unikraft.com/cli/internal/timeouts"
 	"unikraft.com/cli/internal/tunnel"
 	"unikraft.com/cli/internal/types"
@@ -78,7 +79,8 @@ type InstancesCmd struct {
 type InstanceCreateCmd struct {
 	cmd.ResourceCreateCmd[Instance]
 
-	DeleteOnStop bool `group:"flag-create" name:"rm" help:"Automatically delete the instance when it stops."`
+	DeleteOnStop   bool `group:"flag-create" name:"rm" help:"Automatically delete the instance when it stops."`
+	ServiceRollout bool `group:"flag-create" name:"service-rollout" help:"Replace every instance in the service group with a new one."`
 }
 
 func (c *InstanceCreateCmd) Run(ctx context.Context, stdio config.Stdio, partition *resource.Partition) error {
@@ -94,7 +96,120 @@ func (c *InstanceCreateCmd) Run(ctx context.Context, stdio config.Stdio, partiti
 			return fmt.Errorf("--domain cannot be used with --service")
 		}
 	}
-	return c.ResourceCreateCmd.Run(ctx, stdio, partition)
+	_, err := c.RunResources(ctx, stdio, partition)
+	return err
+}
+
+// RunResources creates the instances and, when asked for a rollout, replaces
+// the instances already in the service group with them.
+func (c *InstanceCreateCmd) RunResources(ctx context.Context, stdio config.Stdio, partition *resource.Partition) ([]resource.Resource, error) {
+	if !c.ServiceRollout {
+		return c.ResourceCreateCmd.RunResources(ctx, stdio, partition)
+	}
+	if c.Save != "" {
+		// --save writes the spec the flags hold and creates nothing, so it
+		// keeps the values the user gave and the rollout stays out of it.
+		return c.ResourceCreateCmd.RunResources(ctx, stdio, partition)
+	}
+	return c.runServiceRollout(ctx, stdio, partition)
+}
+
+// runServiceRollout replaces every instance in the target service group. It
+// creates as many instances as the group already holds and waits for all of
+// them to run before it deletes any of the old ones.
+func (c *InstanceCreateCmd) runServiceRollout(ctx context.Context, stdio config.Stdio, partition *resource.Partition) ([]resource.Resource, error) {
+	fields, err := c.CreateFields(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if _, ok := rollout.FieldValue[int64](fields, "replicas"); ok {
+		return nil, fmt.Errorf("--replicas cannot be used with --service-rollout: the rollout creates one instance for every instance already in the service group")
+	}
+	if autostart, _ := rollout.FieldValue[bool](fields, "autostart"); !autostart {
+		return nil, fmt.Errorf("--service-rollout requires --autostart: the old instances are only deleted once the new ones run")
+	}
+	svc, ok := rollout.FieldValue[*InstanceService](fields, "service")
+	if !ok || svc == nil || (svc.Name == "" && svc.UUID == "") {
+		return nil, fmt.Errorf("--service-rollout requires an existing service group")
+	}
+
+	metro, _ := rollout.FieldValue[LinkName[Metro]](fields, "metro")
+	key := multimetro.Key{Metro: cmp.Or(svc.Metro, string(metro)), UUID: svc.UUID}
+	if svc.UUID == "" {
+		key.Name = svc.Name
+	}
+	svcResults, err := (ServiceGroup{}).Get(ctx, []string{key.Canonical()})
+	if err != nil {
+		return nil, fmt.Errorf("looking up service group %q: %w", key.Canonical(), err)
+	}
+	if len(svcResults) == 0 {
+		return nil, fmt.Errorf("service group %q not found", key.Canonical())
+	}
+	svcGroup := svcResults[0].(ServiceGroup)
+
+	oldKeys := make([]string, 0, len(svcGroup.Instances))
+	for _, inst := range svcGroup.Instances {
+		key := multimetro.Key{Metro: cmp.Or(inst.Metro, string(svcGroup.Metro)), UUID: inst.UUID}
+		if inst.UUID == "" {
+			key.Name = inst.Name
+		}
+		if key.Name == "" && key.UUID == "" {
+			continue
+		}
+		oldKeys = append(oldKeys, key.Canonical())
+	}
+	if len(oldKeys) == 0 {
+		log.G(ctx).Info().
+			Str("service", svcGroup.Name).
+			Msg("service group is empty, creating without a rollout")
+		return c.ResourceCreateCmd.RunResources(ctx, stdio, partition)
+	}
+
+	count := int64(len(oldKeys))
+	if err := rollout.ValidateCapacity(ctx, string(svcGroup.Metro), svcGroup.Autoscale, count, fields); err != nil {
+		return nil, err
+	}
+
+	if count > 1 {
+		if err := c.SetCreateFields(ctx, append(slices.Clone(fields), resource.Field{
+			Name:   "replicas",
+			Create: &resource.Patch{Set: count - 1},
+		})); err != nil {
+			return nil, err
+		}
+	}
+	log.G(ctx).Info().
+		Str("service", svcGroup.Name).
+		Int64("instances", count).
+		Msg("rolling out")
+
+	created, err := c.ResourceCreateCmd.RunResources(ctx, stdio, partition)
+	if err == nil && int64(len(created)) < count {
+		err = fmt.Errorf("rollout created %d of %d instances", len(created), count)
+	}
+	// A create reports success for an instance that exists but never reached
+	// the running state, so the states are checked as well as the error.
+	for _, r := range created {
+		inst := r.(Instance)
+		if state := platform.InstanceState(inst.State); state != platform.InstanceStateRunning &&
+			state != platform.InstanceStateStandby {
+			err = errors.Join(err, fmt.Errorf("instance %s is %s, not running", cmp.Or(inst.Name, inst.UUID), state))
+		}
+	}
+	if err != nil {
+		return created, errors.Join(err, rollout.Undo(ctx, partition, Instance{}, created))
+	}
+
+	if err := partition.WrapDeletable(Instance{}).Delete(ctx, oldKeys); err != nil {
+		// The new instances are the wanted state by now, so they stay and the
+		// group holds both sets until the old ones are dealt with.
+		return created, fmt.Errorf("the new instances are up, but deleting the replaced ones failed, so %s are still in the service group: %w",
+			strings.Join(oldKeys, " "), err)
+	}
+	log.G(ctx).Info().
+		Strs("instances", oldKeys).
+		Msg("replaced instances deleted")
+	return created, nil
 }
 
 type Instance struct {
@@ -1621,6 +1736,17 @@ func (Instance) Examples() map[cmd.CmdType][]kingkong.Example {
 	  --metro fra \
 	  --image nginx:latest \
 	  --plugin 'name=sandbox,image=plugins/sandbox:latest,config={"persist_path":"/data"}'`,
+				},
+			},
+			{
+				Description: "Replace every instance in a service group with a new image",
+				Commands: []string{
+					`unikraft instance create \
+	  --metro fra \
+	  --image my-app:v2 \
+	  --service my-service \
+	  --autostart \
+	  --service-rollout`,
 				},
 			},
 		},
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -40,6 +40,7 @@ import (
 	"unikraft.com/cli/internal/muxreader"
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/cmd"
+	"unikraft.com/cli/internal/resource/patch"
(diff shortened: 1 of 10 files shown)


internal/cmd/instances.go
```diff
@@ -88,12 +89,17 @@ type InstanceCreateCmd struct {
 	Replicas  int64    `group:"flag-create" shortcut:"replicas" help:"Number of replicas." placeholder:"n" example:"1,3"`
 	Features  []string `group:"flag-create" shortcut:"features" help:"Instance features." placeholder:"feature"`
 	Template  string   `group:"flag-create" shortcut:"template" help:"Create from instance template." placeholder:"name"`
+
+	Rollout bool `group:"flag-create" shortcut:"rollout" help:"Replace old instances with new ones matching image name"`
```
jedevc: This should map to a field (see the shortcut). `--set rollout=...` should work as well.

internal/cmd/instances.go
```diff
@@ -217,7 +223,11 @@ func (i *InstanceService) UnmarshalText(data []byte) error {
 		*i = InstanceService(parsed)
 		return nil
 	}
-	return i.Link.UnmarshalText([]byte(str))
+	key := multimetro.ParseKey(str)
+	i.Metro = key.Metro
+	i.UUID = key.UUID
+	i.Name = key.Name
+	return nil
```
jedevc: Why? We should just be able to use link unmarshalling here.

internal/cmd/instances.go
```diff
@@ -914,6 +921,131 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 	return results, nil
 }
 
+// runRollout performs a sequential rollout: for each old instance in the same
+// service group whose image base name matches, it creates a new instance,
+// waits for it to reach running state (up to 10 seconds), then deletes the
+// old one.
+func (c *InstanceCreateCmd) runRollout(ctx context.Context, stdio config.Stdio, sandbox *resource.Sandbox) error {
```
jedevc: We need some sort of a way to keep this in sync with `Run`.

jedevc: Maybe we could have some generic way of linking new resources to old resources - we create the news ones, then delete the old ones if that was successful - since that's what makes the rollout.

craciunoiuc: I only kind of remember the discussion we had so I will try to write it down to remember:

You were saying I think that we should generically support the `rollout` component on any type of `resource` that implements it

To start with we would have only `instances` that support rollout.

Later we would maybe support rollout on `certificates` or `volumes` or others.

craciunoiuc:
I reworked the implementation as you asked @jedevc 

Right now only instance rollout is implemented, but we can do other also I think

craciunoiuc:
something funky going on with tests, probably server is misbehaving. Will need to wait till I can make sure tests pass

internal/cmd/instances.go
```diff
@@ -1050,9 +1054,24 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+			rollout = field.Create.Set.(bool)
 		}
 	}
 
+	if rollout && req.Replicas != nil && *req.Replicas > 0 {
+		return nil, fmt.Errorf("--replicas cannot be used with --rollout")
+	}
+	if rollout && req.Autostart != nil && !*req.Autostart {
+		return nil, fmt.Errorf("--rollout cannot be used with --autostart=false")
+	}
```
jedevc: Can't it?

internal/cmd/instances.go
```diff
@@ -1050,9 +1054,24 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+	if rollout && req.Image == nil {
+		return nil, fmt.Errorf("--rollout requires an image (use --image)")
+	}
```
jedevc: Why can't it use a template?

craciunoiuc: Ah, you are right, code I imported was before templates so this got transported in, will check for templates also

internal/cmd/instances.go
craciunoiuc: Would make no sense to rollout without starting. In general you are using rollout to ensure no outage (create instance and start -> remove old instance -> repeat)

If you create new instances stopped you are probably using rollout wrongly 😅 

Still, up to you, I can remove the check (but then I probably also need to remove some other checks)

craciunoiuc: I'm saying this because right now we're loop-waiting for the new instance to be in `running` before deleting the old instance

(I can just put that whole check in an if-block and make it optional though)

craciunoiuc: I will do that and revert depending on what you prefer

craciunoiuc:
One thing left to discuss here is if we're going for the loop-wait-retry logic

If yes, then feel free to merge, if not, then we got to wait for the platform changes to go in

I would say we leave as is because a rollout can take significantly more than a simple start or restart, si it's fine to have a more complex logic in place (at least conceptually)

internal/cmd/instances.go
craciunoiuc: Moved to platform-timeout only

jedevc:
Before merging the timeouts stuff here, could we get https://github.com/unikraft-cloud/cli/pull/223 in first?

internal/cmd/instances.go
```diff
@@ -1195,9 +1198,21 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+			rollout = field.Create.Set.(bool)
 		}
 	}
 
+	if rollout && req.Replicas != nil && *req.Replicas > 0 {
+		return nil, fmt.Errorf("--replicas cannot be used with --rollout")
+	}
+	if rollout && req.Image == nil && req.Template == nil {
+		return nil, fmt.Errorf("--rollout requires an image or template (use --image or --template)")
+	}
```
jedevc: Unnecessary, duplicate check below.

internal/cmd/instances.go
```diff
@@ -1256,6 +1271,150 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 	return results, nil
 }
 
+func (Instance) Rollout(ctx context.Context, created []resource.Resource, fields []resource.Field) error {
```
jedevc: Which of this logic is *actually* specific to just instances. I was imagining that we would just be able to use the existing `delete` methods.

If it's this specialized, there's no point in it being a special case like this, your choice to go back, or to make it more generic.

internal/cmd/instances.go
```diff
@@ -1256,6 +1271,150 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+		return fmt.Errorf("fetching existing instances: %w", err)
+	}
+
+	var toRemove []Instance
+	for _, r := range existing {
+		inst := r.(Instance)
+		if inst.Image.Reference == nil {
+			continue
+		}
+		if inst.Image.Reference.Name() == newImageName {
```
jedevc: I thought we didn't want to do rollouts based on image name - we moved over from doing that in kraftlet.

I think we should just do it based on the service.

internal/cmd/instances.go
```diff
@@ -1256,6 +1271,150 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+	}
+
+	for i, old := range toRemove {
+		var newInst Instance
+		if i < len(created) {
+			newInst = created[i].(Instance)
+		} else {
+			// Create an additional replacement instance with a modified name.
+			rolloutFields := slices.Clone(fields)
+			for j := range rolloutFields {
```
jedevc: Why are we doing *more* creation after our existing creation? We should do all the creation in one step, then do all our deletions.

Also, surely we could do bulk creation using `replicas`.

internal/cmd/instances.go
```diff
@@ -1256,6 +1271,150 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+		}
+
+		if waitErr := group.DoMetro(ctx, g, metro, func(ctx context.Context, mc multimetro.MetroClient) error {
+			_, err := mc.WaitInstanceByUUID(ctx, newInst.UUID, platform.WaitInstanceByUUIDRequestBody{
+				State: platform.InstanceStateRunning,
+			})
+			return err
+		}); waitErr != nil {
+			return fmt.Errorf("waiting for new instance to be running: %w", waitErr)
+		}
```
jedevc: I think this should be configurable - we should be able to configure some timeout, in which we check that *after* the instance has started running that it is still running after `t` time.

We don't really have a good way to check the health of an instance, so we need some way to make sure that we don't start up an instance, it immediately crashes, and then we delete the old one.

internal/cmd/instances.go
```diff
@@ -1256,6 +1271,150 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+	}
+	if len(toRemove) == 0 {
+		return nil
+	}
+
+	if extra := len(toRemove) - len(created); extra > 0 {
+		var baseName string
+		for key, field := range resource.IterFields(fields) {
+			if key.String() == "name" && field.Create != nil && field.Create.Set != nil {
```
jedevc: `resource.GetFieldByPath` instead of iteration.

internal/cmd/instances.go
```diff
@@ -99,6 +99,7 @@ type InstanceCreateCmd struct {
 	Features     []string `group:"flag-create" shortcut:"features" help:"Instance features." placeholder:"feature"`
 	Template     string   `group:"flag-create" shortcut:"template" help:"Create from instance template." placeholder:"name"`
 	DeleteOnStop bool     `group:"flag-create" name:"rm" help:"Automatically delete the instance when it stops."`
+	Rollout      bool     `group:"flag-create" shortcut:"rollout" help:"Replace old instances with new ones matching image name"`
```
jedevc: I wonder if we should rename this to be more specific `--service-rollout`. We rollout on a service.

internal/cmd/instances.go
```diff
@@ -1195,9 +1198,21 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
 				req.Template.Name = &template
 			}
+		case "rollout":
+			rollout = field.Create.Set.(bool)
 		}
 	}
 
+	if rollout && req.Replicas != nil && *req.Replicas > 0 {
+		return nil, fmt.Errorf("--replicas cannot be used with --rollout")
+	}
```
jedevc: Why not?

internal/resource/resource.go
```diff
@@ -62,6 +62,13 @@ type DeletableResource interface {
 	Delete(ctx context.Context, keys []string) error
 }
 
+type RolloutableResource interface {
```
jedevc: In hindsight, maybe this abstraction isn't right. Apologies.

jedevc:
A couple thoughts. There are two modes of rollout I think:
- "Rolling" rollout. As implemented, create new instances, wait for healthy, delete old ones (or delete new ones if never healthy). Very nice, everyone is happy.
- "Replace" rollout. Not implemented, but creates new instances, stops old instances, starts new ones, wait for healthy, delete old ones (or delete new ones, and restore old ones).
  - Imagine a database. Each running database *needs* exclusive access to it's volume, so we should only actually be running one at a time. We'd still like to be able to do rollout, but we can't start any in parallel, so we should try and stop the old one and restart it later if the rollout fails.

Does that make sense? I don't really know what to call these "modes", but I think we should have both of them.

Also, as far as I can tell, the current implementation doesn't have the failure mode, where we restore back to the "known" good state. It assumes that every rollout is perfectly successful, which it may not be.

internal/cmd/instances.go
craciunoiuc: At some point we wanted to do rollout for all resources, not just instances

But each wait and creation is unique so you got to implement this in every resource right?

internal/cmd/instances.go
craciunoiuc: Hmm, I didn't know about this?

So just rollout everything in a service?

internal/cmd/instances.go
craciunoiuc: Well this is the exact opposite of what we want :sweat: 

If you have a service of max size 16 (default), and you have 9 instances in it, it means you can't rollout

If you do it sequentially, it means it will actually fit even if you are at `15/16` used

Alternatively we can add a `--sequential` flag to rollout

internal/cmd/instances.go
craciunoiuc: How would that flag look like like `--recheck 5s`

Aka, wait 5s and check if the instance is still running? I can add that, but one could argue that it's on them if they deploy a service that crashes after X seconds

If it were to crash at boot it would do `starting -> stopped` without going into `running` I think

internal/cmd/instances.go
craciunoiuc: The original intent/discussion was to move this to be generic and implementable for all resources

Not sure if we still want that?

internal/resource/resource.go
craciunoiuc: :sob:

internal/cmd/instances.go
craciunoiuc: It would make the math complicated and confusing? What would replicas mean in this context?

If I'm rolling ~6 instances in a service, and I specify 2 replicas, does it mean I want 2 instances in the end or 6*2==12?

We would need to be very clear in this case

craciunoiuc:
for `rolling` -- true, there is no rolling back, but you can't really do that if you're deleting instances while you're going :sweat_smile: 

for `replace` -- yes, we could do that, and I guess that should also allow for doing the health check after some time, and would only use the same amount of space in the service group.

I think we can do both, with `rolling` being the fast "destructive" one that just waits for instances to start, and `replace` being he "safe" one

With the small note that running a database in a service group is conceptually wrong so if we want to truly support database rollout we need to support non-servicegroup based rollout

jedevc:
> With the small note that running a database in a service group is conceptually wrong so if we want to truly support database rollout we need to support non-servicegroup based rollout

Hm, so how would you imagine database rollout working? I think service group still makes sense here - it's about the *database service* that you care about, but I'm curious why this is wrong.

> for `rolling` -- true, there is no rolling back, but you can't really do that if you're deleting instances while you're going 😅

Indeed, that's why I was suggesting creating them all, and only deleting once you're convinced of the health of the others. Similar to how rolling deployments work in k8s.

> I think we can do both, with `rolling` being the fast "destructive" one that just waits for instances to start, and `replace` being he "safe" one

I think the safety of them is the same no? `replace` results in downtime, but ensures that we never have more than `n` instances actively running, while `rolling` potentially allows up to `2n` instances running, but never dipping below `n`.

I don't actually know actually if the names `rolling`/`replace` are right here. Maybe there's some better name here, or some better way of expressing the constraint (like `max_running = n`).

jedevc:
Note, the platform doesn't really have a concept of instance health… this is sad.

IMO, we should *currently* think of healthy as:

* Instance starts and enters a running state
* We monitor the instance for "time"
* We check that the instance is still running
* Note instance as healthy

However, as we build the rollout flag, etc, we should consider that at some point, we would add a "real" healthcheck system (or add the ability to monitor health in new ways, like by checking logs, etc)

internal/cmd/instances.go
craciunoiuc: Ok, let's focus on all-at-once for now

internal/cmd/instances.go
craciunoiuc: Ok, so everything in a service, if you have multiple images doing different things in a service you are doing smth wrong

internal/cmd/instances.go
craciunoiuc: Commands now block till in running

If they rollout with a running instance that crashes 5m later that's on them

internal/resource/resource.go
craciunoiuc: Ok, removing the abstraction and focusing only on rolling instances in services

internal/cmd/instances.go
craciunoiuc: I'm keeping this if check, I still find it confusing

internal/cmd/instances.go
craciunoiuc: We no longer want that, only instances in services

internal/cmd/instances.go
craciunoiuc: Only instances in services for now

craciunoiuc:
Ok so I think the summary is:
* Two modes, `rolling`/`replace` doing what we said above
* A rollout healthcheck flag to optionally recheck health after X seconds
* Implemented just for instances with services for now, in the interest of time

jedevc:
Yup. For the last one, eventually we could do things like support rollout over tags. Might be worth planning what the CLI interface looks like? So it's nice and consistent for when we do.

PR #312, opened 2026-05-08
feat(cmd): Add missing examples and slightly reword

craciunoiuc:
Closes: TOOL-846

### DIFF
--- a/internal/cmd/certificates.go
+++ b/internal/cmd/certificates.go
@@ -278,30 +278,40 @@ func (Certificate) Examples() map[cmd.CmdType][]kingkong.Example {
 	return map[cmd.CmdType][]kingkong.Example{
 		cmd.CmdTypeGet: {
 			{
-				Description: "Get a certificate by name or UUID",
+				Description: "Inspect a certificate by name",
 				Commands:    []string{"unikraft certificate get demo-cert"},
+			},
+			{
+				Description: "Inspect a certificate by UUID",
+				Commands:    []string{"unikraft certificate get d4f0c132-a70e-4777-9631-7dfe08a1c15d"},
+			},
+			{
+				Description: "Inspect a certificate in a specific metro",
+				Commands:    []string{"unikraft certificate get fra/demo-cert"},
+			},
+			{
+				Description: "Show certificate details in JSON format",
+				Commands:    []string{"unikraft certificate get demo-cert -o json"},
 			},
 		},
 		cmd.CmdTypeList: {
 			{
-				Description: "List all certificates",
+				Description: "List all certificates across metros",
 				Commands:    []string{"unikraft certificate list"},
+			},
+			{
+				Description: "List certificates and watch for changes",
+				Commands:    []string{"unikraft certificate list -w"},
 			},
 		},
 		cmd.CmdTypeCreate: {
 			{
-				Description: "Create a new certificate",
+				Description: "Create a self-signed certificate and upload it",
 				Commands: []string{
 					`openssl req -x509 -newkey rsa:2048 -sha256 -days 365 -nodes \
   -subj "/CN=demo.unikraft.dev" \
   -keyout cert.key \
   -out cert.pem`,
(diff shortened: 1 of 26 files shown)


internal/cmd/certificates.go
```diff
@@ -310,12 +316,27 @@ func (Certificate) Examples() map[cmd.CmdType][]kingkong.Example {
...
+	  --pkey cert.key \
+	  --metro sfo \
+	  --dry-run`,
+				},
+			},
 		},
 		cmd.CmdTypeDelete: {
 			{
-				Description: "Delete a certificate by name or UUID",
+				Description: "Delete a certificate by name",
```
nurof3n, line 337: cannot delete resources by uuid??

internal/cmd/instance_templates.go
```diff
@@ -412,27 +412,45 @@ func (InstanceTemplate) Examples() map[cmd.CmdType][]kingkong.Example {
 	return map[cmd.CmdType][]kingkong.Example{
 		cmd.CmdTypeGet: {
 			{
-				Description: "Inspect an instance template by name or UUID",
+				Description: "Inspect an instance template by name",
```
nurof3n, line 415: bruski

internal/cmd/certificates.go
craciunoiuc: added

internal/cmd/certificates.go
```diff
@@ -278,30 +278,40 @@ func (Certificate) Examples() map[cmd.CmdType][]kingkong.Example {
...
+			},
+			{
+				Description: "Show certificate details in JSON format",
+				Commands:    []string{"unikraft certificate get demo-cert -o json"},
+			},
 		},
 		cmd.CmdTypeList: {
 			{
-				Description: "List all certificates",
+				Description: "List all certificates across metros",
```
jedevc, line 299: Does including `across metros` here add anything? All commands work across all metros.

internal/cmd/instances.go
```diff
@@ -918,34 +918,60 @@ func (Instance) Examples() map[cmd.CmdType][]kingkong.Example {
...
+				Commands:    []string{"unikraft instance list --sort -timestamps.created"},
+			},
+			{
+				Description: "Watch instances and refresh every 5 seconds",
+				Commands:    []string{"unikraft instance list -w 5s"},
+			},
+			{
+				Description: "List instances as a table with additional fields",
+				Commands:    []string{"unikraft instance list -f +resources -o table"},
+			},
```
jedevc, line 961: Ah. I'm not entirely sure about some of these.

The issue is that we should use the examples to give info *specific* to this command. But things like `--field`, `--sort`, `--filter`, `--watch`, `--dry-run`, etc, apply to everything. I don't really know where these need to appear in the help hierarchy to make them better, but I don't want to add them to every single place - it's a lot of repetition (and e.g. suppose we update the `--watch` flag - then we have to update it *everywhere*).

jedevc (changes requested):

internal/cmd/certificates.go
craciunoiuc: True

Probably only makes sense if there's also an example that has `--metro=smth` and that says `on a single metro"`

internal/cmd/instances.go
craciunoiuc: also true

If I remember correctly in kraftkit we were adding them ocassionally to some commands to show/remind people how to use them

Of course there should be a better way to do this

Maybe if we would show them only in the top level help?

jedevc: I think we should show common functionality for each command - but we shouldn't have examples that explain the flags like `--sort` unless that really is a super used operation for exactly that command.

Instead, we should expand the help for `--sort` and such for those cases.

craciunoiuc: Hmm, okay, makes sense, I'll do a pass through it and hand-curate

PR #315, opened 2026-05-11
fix: Remove rom file helper

jedevc:
This feature has super weird behavior:
- If I mount a `file` ROM to /etc/foo.txt, it'll replace the existing contents of /etc.
- If I mount two separate ROMs, one as /mnt/a.txt and another as /mnt/b.txt, this tries to create two mounts at /mnt, which conflict.

This feature really just needs platform support, so we should remove it from the CLI for now, and not try and hack it in.

### DIFF
--- a/cmd/unikraft/build_test.go
+++ b/cmd/unikraft/build_test.go
@@ -130,39 +130,6 @@ cmd: ["cat", "/rom/hello.txt"]
 			})
 	})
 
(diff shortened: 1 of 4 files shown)


jedevc:
FYI @nurof3n

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #317, opened 2026-05-11
fix: Use colorprofile detection on stdout/stderr

jedevc:
So that then we automatically downsample for all our normal outputs.

e.g. by setting `TERM=xterm COLORTERM=`, you can mimic a 16 color terminal, and get automatic downsampling.
image


nderjung (approved):
That's nice!

PR #319, opened 2026-05-12
refactor: Rework integration testing framework and usage

craciunoiuc:
Closes: TOOL-850


Taskfile.yml
```diff
@@ -72,12 +72,32 @@ tasks:
       vars:
         go_tags: "integration"
         package: ./cmd/unikraft ./internal/builder
-        CLI_ARGS: '{{if .TEST}}-run "TestGolden/{{.TEST}}"{{end}}'
+        CLI_ARGS: '{{if .TEST}}-run "TestIntegration/{{.TEST}}"{{end}}'
 
   integration-update:
-    desc: Update integration test golden files. Use TEST=<pattern> to filter (e.g., TEST=help).
+    desc: Re-record expected outputs for integration tests. Use TEST=<pattern> to filter (e.g., TEST=help).
```
jedevc: From the `Taskfile` perspective, we should have the following types of tests:
- `unit` (same as today)
- `integration` (all the behavior tests). No more `integration-update`.
- `golden` (all the output + help matching ones). `golden-update` should update all these.

cmd/unikraft/instance_templates_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"testing"
 )
 
-func instanceTemplatesTests(t *testing.T, r *testRunner) {
```
jedevc, line 12: Let's clean up this pattern while we're here as well. This should just be `TestInstances`, instead of running at  `TestIntegration/instances`.

cmd/unikraft/instance_templates_test.go
```diff
@@ -61,24 +58,20 @@ func instanceTemplatesTests(t *testing.T, r *testRunner) {
...
 				{args: []string{
 					unikraftCmd, "instance", "create",
 					"--set", "name=test-from-template-$UNIQ_INST",
 					"--set", "metro=" + metroName,
 					"--set", "template=$TEMPLATE_NAME",
 				}},
-				// Verify the new instance was created
-				{args: []string{unikraftCmd, "instance", "inspect", "test-from-template-$UNIQ_INST"}},
-				// Clean up template
+				{args: []string{unikraftCmd, "instance", "inspect", "test-from-template-$UNIQ_INST"}, match: []string{`image:\s+nginx`, `memory:\s+128`}},
```
jedevc: In my original message, I actually suggested we change the way these integration tests are written completely.

https://discord.com/channels/879723098881028167/1502228369859678319/1502248500904198224

```go
    t.Run("create", func(t *testing.T) {
        r := runner(t, true)  // true indicates this test is online, and should be skipped if no config is available
        out := r.cli(t.Context(), t, []string{
            "unikraft", "instance", "create",
            "--set", "name=test-$UNIQ_INST",
            "--set", "metro=" + r.cfg.MetroName,
            "--set", "image=nginx:latest",
            "--set", "runtime.env=A=1,B=2,C=3",
            "--set", "autostart=true",
            "--set", "resources.memory=128",
            "--set", "resources.vcpus=1",
        })
        // TODO: verify that out contains various properties - like env, it starts
        // running. use regexp! maybe have a cute little helper? like `verifyContains(out, "env:.*\nA=1\nB=@", "image:.*nginx:latest")`

        out = r.cli(t.Context(), t, []string{"unikraft", "instance", "inspect"})
        // TODO: verify that inspect contains the same stuff as create

        out = r.cli(t.Context(), t, []string{"unikraft", "instance", "list"})
        // TODO: verify that it's in the list

        out = r.cli(t.Context(), t, []string{"unikraft", "instance", "delete-"})
        // TODO: verify the output of delete
    })
```

Instead of doing this `r.online().run()` mess with the command struct, we should just make this normal code - then we can insert arbitrary assertions at any point, which makes it more powerful.

This applies for all the integration tests. This is definitely invasive, but it will make the tests better, and flow easier, instead of relying on the magic helper. It's just a sketch, and up to you a bit, but I do want to change all the tests to use this kind of flow.

cmd/unikraft/instances_test.go
```diff
@@ -456,56 +399,110 @@ func instancesTests(t *testing.T, r *testRunner) {
...
+		Name:      "my-instance",
+		UUID:      "7b79b250-0658-4d10-8dc0-d854399d7e74",
+		Tags:      []string{"env-prod", "team-core"},
+		State:     types.InstanceState(platform.InstanceStateRunning),
+	}
+	require.NoError(t, sample.Image.UnmarshalText([]byte("nginx:latest")))
+	sample.Runtime.Args = cmd.InstanceArgs{"arg1", "arg2"}
+	sample.Runtime.Env = map[string]string{"KEY1": "val1", "KEY2": "val2"}
+	sample.Resources.Memory = 256
+	sample.Resources.VCPUs = 2
```
jedevc: Ideally, this should look more like *one* assignment.

```go
samples := cmd.Instance{
...
}
```

This will keep it shorter, and easier to follow.

cmd/unikraft/integration/instance_test.go
jedevc, line 1: These files are now pretty mixed up. I would suggest a slightly different file structure.

- `cmd/unikraft/help_test.go` contains *all* the help gilded tests. It's in `cmd/unikraft` since it calls the cli.
- `internal/cmd/output_test.go` contains *all* the gilded output tests (yeah it's gonna be a long file, but it'll have lots of boilerplate). It can be in `internal/cmd`, since it doesn't rely on any kong stuff, just printing.
- `cmd/unikraft/integration/instance_test.go` contains instance tests, etc. So all the integration tests go sit in a folder together, inside `cmd/unikraft` since it tests the whole cli.

cmd/unikraft/testdata/TestOutput/instances
```diff
@@ -0,0 +1,700 @@
+=== kv ===
```
jedevc: We should also have a "kv-all" section? That shows the equivalent of `-fall` maybe?

cmd/unikraft/testdata/TestOutput/instances
```diff
@@ -0,0 +1,700 @@
...
+networks:
+- uuid:        net-uuid-1234
+  private-ip:  192.168.1.10
+  mac:         aa:bb:cc:dd:ee:ff
+scale-to-zero: policy=on,stateful=true,cooldown-time=500ms,notify-time=100ms
+stop:
+  reason:      crashed
+  exit-code:   1
+
+=== table ===
```
jedevc: nit: Make these headers hit the 80 char limit, so they're more distinct.

.github/workflows/test.yaml
```diff
@@ -54,9 +54,19 @@ jobs:
...
 
+      - name: Run help tests
+        shell: bash
+        run: |
+          task --yes help
+
+      - name: Run output tests
+        shell: bash
+        run: |
+          task --yes output
```
jedevc: These should run even if the unit tests fail, so we get as much feedback at once as possible.

We should run them all, but if any of them fail, then the whole job should fail.

cmd/unikraft/main_test.go
```diff
@@ -42,12 +42,13 @@ import (
 	"unikraft.com/cli/internal/config"
 	"unikraft.com/cli/internal/integration"
 	"unikraft.com/cli/internal/resource"
+	resourcecmd "unikraft.com/cli/internal/resource/cmd"
 )
 
 const unikraftCmd = "unikraft"
```
jedevc: nit: Remove this while we're here, I added it, but don't like it anymore 😆

cmd/unikraft/main_test.go
```diff
@@ -217,7 +210,7 @@ func (b *testBuilder) run(t *testing.T, commands []command) {
 		cmd.Env = slices.DeleteFunc(cmd.Env, func(s string) bool {
 			return strings.HasPrefix(s, "UNIKRAFT_")
 		})
-		cmd.Env = append(cmd.Env, "NO_COLOR=1") // color makes golden files harder to read
+		cmd.Env = append(cmd.Env, "NO_COLOR=1")
```
jedevc: comment gone?

cmd/unikraft/main_test.go
```diff
@@ -639,3 +418,206 @@ func buildUnikraftBinary(t *testing.T) string {
...
+	}
+	for _, c := range offlineCleaners {
+		s = c.pattern.ReplaceAllString(s, c.repl)
+	}
+	return s
+}
+
+// offlineCleaners normalize build-environment values that differ between
+// machines.
+var offlineCleaners = []cleaner{
```
jedevc: 🤔 Where are these still used?

cmd/unikraft/main_test.go
```diff
@@ -639,3 +418,206 @@ func buildUnikraftBinary(t *testing.T) string {
...
+	}
+
+	t.Parallel()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			tt.fn(t)
+		})
+	}
+}
```
jedevc: Instead of doing a loop, we could just have a little local helper, and then it would be:

```go
run("instances", instancesOutputTests)
```

cmd/unikraft/main_test.go
```diff
@@ -639,3 +418,206 @@ func buildUnikraftBinary(t *testing.T) string {
...
+func TestOutput(t *testing.T) {
+	tests := []struct {
+		name string
+		fn   func(*testing.T)
+	}{
+		{"instances", instancesOutputTests},
+		{"volumes", volumesOutputTests},
+		{"services", servicesOutputTests},
+		{"certificates", certificatesOutputTests},
+		{"images", imagesOutputTests},
```
jedevc: :eyes: this is missing resources. `instance-templates`, etc.

cmd/unikraft/main_test.go
craciunoiuc: In the output tests, because some outputs have timestamps etc.

cmd/unikraft/main_test.go
craciunoiuc: ah, I thought it looked weird 😅 

I can add it back

cmd/unikraft/main_test.go
craciunoiuc: sounds good

cmd/unikraft/testdata/TestOutput/instances
craciunoiuc: you mean adding `=` chars to make it long? ok

cmd/unikraft/main_test.go
craciunoiuc: Whoops, my bad

cmd/unikraft/main_test.go
jedevc: But the output tests should be based on deterministic inputs?

cmd/unikraft/instance_templates_test.go
craciunoiuc: Hmm, ok

Taskfile.yml
craciunoiuc: Right, forgot to remove `integration-update`

will do + the rest

cmd/unikraft/instances_test.go
craciunoiuc: Right, so a big `cmd.Instance` block, got it.

cmd/unikraft/integration/instance_test.go
craciunoiuc: Your wish is my command

cmd/unikraft/instance_templates_test.go
craciunoiuc: I will leave this for last actually

cmd/unikraft/main_test.go
craciunoiuc: Sure, but if you have a different go version for example, you compile, and run `unikraft version` these will fail without cleaners.

check the list it's:
- go versions
- go OS/ARCH
- tmp paths
- timestamps

I think it's good enough no?

jedevc: Okay, then if this is the case, then we shouldn't have `unikraft version` as a gilded test. Just make it a separate test, where we can use regexp, etc.

cmd/unikraft/main_test.go
craciunoiuc: I think I got all

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+		e.uniq = make(map[string]string)
+	}
+	if e.certs == nil {
+		e.certs = make(map[string]*generatedCert)
+	}
+	if e.env == nil {
+		e.env = make(map[string]string)
+	}
+	expanded := make([]string, 0, len(args))
+	for _, arg := range args {
```
jedevc: All of the expander logic should go as well. The only reason this is here, is again, so we could actually have all of this written in a table. Now that we aren't using the tables, this can all be removed as well.

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+	c.Env = append(c.Env, "NO_COLOR=1") // color makes golden files harder to read
+	c.Env = append(c.Env, "UNIKRAFT_CONFIG="+r.configPath)
+	c.Env = append(c.Env, "BUILDKIT_PROGRESS=quiet")
+	c.Env = append(c.Env, resource.UnikraftSandboxEnv+"="+r.sandboxPath)
+
+	if r.cmdCount > 0 {
+		// HACK: this is awful, but the platform can take a moment for things to
+		// get ready :(
+		time.Sleep(500 * time.Millisecond)
+	}
```
jedevc: We should remove this.

cmd/unikraft/integration/instance_test.go
```diff
@@ -0,0 +1,445 @@
...
+			"--max-time", "10",
+			"https://$FQDN",
+		})
+		assert.Regexp(t, `HTTP 200 OK`, out)
+
+		r.cli(t.Context(), t, []string{"unikraft", "instance", "delete", "test-$UNIQ_INST"})
+	})
+
+	t.Run("start-stop", func(t *testing.T) {
+		t.Skip("start doesn't actually wait to start")
```
jedevc: Aha, we can remove this too. We just need to comment out the check that checks the output of `start`.

cmd/unikraft/integration/instance_test.go
```diff
@@ -0,0 +1,445 @@
...
+			"--show-error",
+			"--output", "/dev/null",
+			"--write-out", `HTTP %{http_code} OK\n%header{server}\n`,
+			"--retry", "10",
+			"--retry-delay", "2",
+			"--retry-all-errors",
+			"--connect-timeout", "5",
+			"--max-time", "10",
+			"https://$FQDN",
+		})
```
jedevc: nit: can we write a little http helper? like `cli`, but that makes requests. we shouldn't need to shell out to curl, we can use the golang http client.

cmd/unikraft/integration/instance_test.go
```diff
@@ -0,0 +1,445 @@
...
+			"--set", "resources.vcpus=1",
+			"--set", "service.services=443:8080/tls+http",
+			"--set", "service.domains=name=$UNIQ_DOMAIN",
+		})
+
+		out := r.cli(t.Context(), t, []string{
+			"unikraft", "instance", "inspect", "test-$UNIQ_INST",
+			"--output", "template=" + `{{ (index .service.domains 0).fqdn }}`,
+		})
+		r.setEnv("FQDN", strings.TrimSpace(out))
```
jedevc: Instead of doing this, we can just store `fqdn` as a variable. No need to mess with the env.

cmd/unikraft/integration/instance_test.go
```diff
@@ -0,0 +1,445 @@
...
+			"--set", "runtime.env=A=1,B=2,C=3",
+			"--set", "autostart=true",
+			"--set", "resources.memory=128",
+			"--set", "resources.vcpus=1",
+			"--set", "service.services=443:8080/tls+http",
+			"--set", "service.domains=name=$UNIQ_DOMAIN",
+		})
+
+		out := r.cli(t.Context(), t, []string{
+			"unikraft", "instance", "inspect", "test-$UNIQ_INST",
```
jedevc: We can remove `UNIQ_INST` - instead we can generate a unique instance name at the top of the test, and just use that.

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+// setEnv sets an environment variable for future command expansion.
+func (r *testRunner) setEnv(key, value string) {
+	if r.expander.env == nil {
+		r.expander.env = make(map[string]string)
+	}
+	r.expander.env[key] = value
+}
+
+// runCmd executes a command and returns the combined output and any error.
+func (r *testRunner) runCmd(ctx context.Context, t *testing.T, args []string) (string, error) {
```
jedevc: Since we pass in `t.Context()`, we can just calculate that here.

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+		})
+		if err != nil {
+			panic(err)
+		}
+		expanded = append(expanded, arg)
+	}
+	return expanded
+}
+
+func generateCert() *generatedCert {
```
jedevc: Could maybe add this helper to `internal/integration`.

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	r.cmdCount++
+
+	err := c.Run()
+	combined := ansi.Strip(stdout.String() + stderr.String())
+	return combined, err
+}
+
+// cli runs a CLI command and returns the combined output. The command must succeed.
+func (r *testRunner) cli(ctx context.Context, t *testing.T, args []string) string {
```
jedevc: Instead of having `cliFail` and `cliMaybe` variants, maybe we should use a `...opts` pattern here.

cmd/unikraft/main_test.go
jedevc, line 1: This now just contains helpers :tada:

I think we can remove this file and move stuff to elsewhere.

- Noticed that `testEnv` here is very similar to `testRunner` in `integration/`? We shouldn't have that duplication, we could put some of those helpers in `internal/integration`.
- Likewise with `buildUnikraftBinary`.
- We can move `gild` into `internal/integration`.
- We can also move `TestHelp` into `help_test.go`, so it's all in one place.

cmd/unikraft/main_test.go
```diff
@@ -26,347 +18,99 @@ import (
...
-	s = strings.ReplaceAll(s, "\r", "\n")
+	var result strings.Builder
+	result.WriteString("$ " + strings.Join(formatArgs(args), " ") + "\n")
 
-	// remove ANSI escape sequences
-	s = ansi.Strip(s)
-	s = strings.TrimRightFunc(s, unicode.IsSpace)
-	if s == "" {
-		return ""
+	out := normalizeOutput(output.String())
```
jedevc: As mentioned previously - we shouldn't be normalizing the output. Any commands that have different content depending on these kind of things should not be tested via a gild approach.

We should still do ansi trimming though.

internal/cmd/output_test.go
```diff
@@ -0,0 +1,332 @@
...
+		output.WriteString(rendered)
+		output.WriteString("\n\n")
+	}
+
+	return strings.TrimRightFunc(output.String(), unicode.IsSpace) + "\n"
+}
+
+// gild runs a callback for each arg, concatenates the outputs, and asserts
+// against the golden file for the current test.
+func gild[Arg any](ctx context.Context, t *testing.T, callback func(context.Context, *testing.T, Arg) string, args ...Arg) {
```
jedevc: Duplication.

internal/cmd/output_test.go
```diff
@@ -0,0 +1,332 @@
...
+		UUID:      "7b79b250-0658-4d10-8dc0-d854399d7e74",
+		Tags:      []string{"env-prod", "team-core"},
+		State:     types.InstanceState(platform.InstanceStateRunning),
+		Runtime: struct {
+			Args cmd.InstanceArgs  `mirror:"instance.args" field:",short" create:"set" edit:"set"`
+			Env  map[string]string `mirror:"instance.env" field:",long" create:"set" edit:"set,add,del=keys"`
+		}{
+			Args: cmd.InstanceArgs{"arg1", "arg2"},
+			Env:  map[string]string{"KEY1": "val1", "KEY2": "val2"},
+		},
```
jedevc: Ohhh. I hadn't realized it would do this 😢 That's why it was all split out before 😢 

Kinda wondering if instead of this we could simplify it by writing a big `json` document inline instead...

cmd/unikraft/integration/build_test.go
```diff
@@ -0,0 +1,201 @@
...
+			imageTag := uniq()
+			instName := uniq()
+			romImage := romImagePrefix + ":" + imageTag
+			baseImage := baseImagePrefix + ":" + imageTag
+			romFlag := "image=" + romImage + ",name=myrom"
+			if romFormat == "erofs" {
+				romFlag += ",at=/rom"
+			}
+
+			r.writeFiles(t, map[string]string{
```
jedevc: :eyes: can you use `fstest` here instead? Something like https://github.com/unikraft-cloud/agent/blob/ff31983b553658ef6766a7060e055fa3bebacf14/internal/repositories/image/store/file/file_test.go#L46-L49. Then we can remove the `writeFiles` helper.

cmd/unikraft/integration/build_test.go
```diff
@@ -0,0 +1,201 @@
...
+			{
+				name:        "registry",
+				imagePrefix: ir.cfg.Profile.Organization + "/busybox-e2e",
+			},
+			{
+				name:        "direct-push",
+				imagePrefix: ir.cfg.Metro.Index().Host + "/" + ir.cfg.Profile.Organization + "/busybox-e2e",
+			},
+		}
+	}
```
jedevc: nit: move closer to use.

cmd/unikraft/integration/instance_test.go
```diff
@@ -0,0 +1,452 @@
...
+			"--output", "quiet",
+			"--set", "name=test-" + instName,
+			"--set", "metro=" + metroName,
+			"--set", "image=nginx:latest",
+			"--set", "autostart=true",
+			"--set", "resources.memory=16Mib",
+			"--set", "resources.vcpus=1",
+		})
+
+		r.cli(t, []string{"unikraft", "instance", "wait", "--until", "state==stopped", "--timeout", "10s", "test-" + instName})
```
jedevc: Make sure that the tests are still testing the right behavior. This test previously made sure that the stop codes + explanation appeared and made sense. Now we aren't checking that.

cmd/unikraft/integration/instance_test.go
```diff
@@ -0,0 +1,452 @@
...
+		out := r.cli(t, []string{
+			"unikraft", "instance", "inspect", "test-" + instName,
+			"--output", "template=" + `{{ (index .service.domains 0).fqdn }}`,
+		})
+		fqdn := strings.TrimSpace(out)
+
+		r.cli(t, []string{"unikraft", "instance", "wait", "--until", "state==running", "--timeout", "10s", "test-" + instName})
+
+		body := r.httpGet(t, "https://"+fqdn)
+		assert.NotEmpty(t, body)
```
jedevc: Check some feature of the output for these.

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,259 @@
...
+	"unikraft.com/x/log"
+
+	"unikraft.com/cli/internal/cmd"
+	"unikraft.com/cli/internal/config"
+	integ "unikraft.com/cli/internal/integration"
+	"unikraft.com/cli/internal/resource"
+)
+
+// uniq generates a unique 12-character hex string for use in test resource names.
+func uniq() string {
```
jedevc: nit: make sure we don't generate duplicates.

cmd/unikraft/integration/main_test.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// testRunner holds per-test state for running CLI commands in integration tests.
+type testRunner struct {
+	cfg          *integ.Config
+	unikraftPath string
+	configPath   string
+	sandboxPath  string
+	dir          string
+}
```
jedevc: Merge these structs together. No reason to have them be separate.

cmd/unikraft/integration/certificate_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	integ "unikraft.com/cli/internal/integration"
+)
+
+func TestCertificates(t *testing.T) {
+	ir := newIntegrationRunner(t)
+	metroName := ""
+	if ir.cfg != nil {
+		metroName = ir.cfg.MetroName
+	}
```
jedevc: I think we can remove the snippets that look like this, and just access them. This was only to support the table tests before, we can just do all this computation inside the actual test itself.

cmd/unikraft/main_test.go
craciunoiuc: Sounds good

cmd/unikraft/main_test.go
craciunoiuc: Hmm, sure, though there might be a lot of them that don't fit this 😭 

Will check

internal/cmd/output_test.go
craciunoiuc: I mean, it's not that bad is it? It's definitely better than an inline json lol

If you ask me:
split assign > big messy block > inline json

cmd/unikraft/integration/instance_test.go
craciunoiuc: Betrayal upon betrayal, I will do a pass yeah

cmd/unikraft/integration/build_test.go
craciunoiuc: Interesting, I will leave this for last as an additional pass.

cmd/unikraft/integration/instance_test.go
craciunoiuc: I don't understand your sentence 😔 

You mean we should check also the actual output generated by this when we use it?

---

I guess in this case we can check `Welcome to Nginx!` or whatever the default prompt is

internal/cmd/output_test.go
jedevc: It just means that if we update the Runtime struct, we have to go and update this here, or the code won't even compile. I would argue this is worse than the split assign.

cmd/unikraft/integration/instance_test.go
jedevc: > I guess in this case we can check `Welcome to Nginx!` or whatever the default prompt is

Exactly.

internal/cmd/output_test.go
craciunoiuc: Ok, removed the weird structs, let me know if it will look fine now

cmd/unikraft/integration/build_test.go
craciunoiuc: Wasn't that bad

cmd/unikraft/integration/instance_test.go
craciunoiuc: Apparently only one was missing 😋

cmd/unikraft/integration/main_test.go
craciunoiuc: New take: we generate a random number and then increase it one by one on subsequent calls, like in kraft

cmd/unikraft/integration/build_test.go
craciunoiuc: Happened in just one place, fixed.

cmd/unikraft/integration/instance_test.go
craciunoiuc: Okay it found several things that it fixed. Complexity is huge though considering how many lines are changed

craciunoiuc:
server failed again when test ran

cmd/unikraft/integration/helpers_test.go
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
```
craciunoiuc: `2026`

Feel free to ignore hehe

jedevc (approved):

PR #322, opened 2026-05-12
fix(builder): Allow custom dockerfile names when building

craciunoiuc:
Previously, the name was cut from the path, so it used the default 'Dockerfile' name breaking functionality.

~~Depends on: https://github.com/unikraft-cloud/x/pull/253~~
Merged

Closes: TOOL-855

~~TODOs left:~~
~~* Bump the library after merging~~
~~* Update the golden files for tests when they start behaving~~

### DIFF
--- a/internal/builder/build_test.go
+++ b/internal/builder/build_test.go
@@ -87,18 +87,18 @@ EOF
 `
 	rootfsPath := writeDockerfile(t, dockerfile)
 	opts := BuildOpts{
-		Runtime: "unikraft.io/unikraft.org/base",
+		Runtime: "unikraft.io/official/base-compat",
 		Rootfs: FSOpts{
 			Format: kraftfile.FsTypeCpio,
 			Type:   kraftfile.SourceTypeDockerfile,
 			Path:   rootfsPath,
 		},
-		Platform: []ocispec.Platform{{OS: "fc", Architecture: "x86_64"}},
+		Platform: []ocispec.Platform{{OS: "kraftcloud", Architecture: "x86_64"}},
 	}
 
 	imgs := runBuild(t, ctx, opts)
 	require.Len(t, imgs, 1)
-	assertPlatforms(t, imgs, []string{"fc/x86_64"})
+	assertPlatforms(t, imgs, []string{"kraftcloud/x86_64"})
 }
 
 func TestBuildSinglePlatformErofsIntegration(t *testing.T) {
@@ -112,18 +112,18 @@ EOF
 `
 	rootfsPath := writeDockerfile(t, dockerfile)
 	opts := BuildOpts{
-		Runtime: "unikraft.io/unikraft.org/base",
+		Runtime: "unikraft.io/official/base-compat",
 		Rootfs: FSOpts{
 			Format: kraftfile.FsTypeErofs,
 			Type:   kraftfile.SourceTypeDockerfile,
 			Path:   rootfsPath,
 		},
-		Platform: []ocispec.Platform{{OS: "fc", Architecture: "x86_64"}},
+		Platform: []ocispec.Platform{{OS: "kraftcloud", Architecture: "x86_64"}},
 	}
 
 	imgs := runBuild(t, ctx, opts)
 	require.Len(t, imgs, 1)
-	assertPlatforms(t, imgs, []string{"fc/x86_64"})
+	assertPlatforms(t, imgs, []string{"kraftcloud/x86_64"})
 }
 
 func TestBuildDefaultErofsFormatBaseCompatIntegration(t *testing.T) {
@@ -169,18 +169,18 @@ RUN ln -s /etc/passwd /another-test-link
 `
 	rootfsPath := writeDockerfile(t, dockerfile)
 	opts := BuildOpts{
-		Runtime: "unikraft.io/unikraft.org/base",
+		Runtime: "unikraft.io/official/base-compat",
 		Rootfs: FSOpts{
 			Format: kraftfile.FsTypeCpio,
 			Type:   kraftfile.SourceTypeDockerfile,
 			Path:   rootfsPath,
 		},
-		Platform: []ocispec.Platform{{OS: "fc", Architecture: "x86_64"}},
+		Platform: []ocispec.Platform{{OS: "kraftcloud", Architecture: "x86_64"}},
 	}
 
 	imgs := runBuild(t, ctx, opts)
 	require.Len(t, imgs, 1)
-	assertPlatforms(t, imgs, []string{"fc/x86_64"})
+	assertPlatforms(t, imgs, []string{"kraftcloud/x86_64"})
 
 	// Inspect the CPIO archive to verify symlink targets
 	initrd := imgs[0].Initrd
@@ -270,18 +270,18 @@ RUN --mount=type=secret,id=api_key cat /run/secrets/api_key | grep -q s3cr3t
 	require.NoError(t, err)
 
 	opts := BuildOpts{
-		Runtime: "unikraft.io/unikraft.org/base",
+		Runtime: "unikraft.io/official/base-compat",
 		Rootfs: FSOpts{
 			Format: kraftfile.FsTypeCpio,
 			Type:   kraftfile.SourceTypeDockerfile,
 			Path:   rootfsPath,
 		},
 		Secrets:  secrets,
-		Platform: []ocispec.Platform{{OS: "fc", Architecture: "x86_64"}},
+		Platform: []ocispec.Platform{{OS: "kraftcloud", Architecture: "x86_64"}},
 	}
 	imgs := runBuild(t, ctx, opts)
 	require.Len(t, imgs, 1)
-	assertPlatforms(t, imgs, []string{"fc/x86_64"})
+	assertPlatforms(t, imgs, []string{"kraftcloud/x86_64"})
 }
 
 func TestBuildCmdEnvLabelsIntegration(t *testing.T) {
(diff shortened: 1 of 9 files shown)


jedevc:
This shouldn't be merged as is, it should use the kraftfile `dockerfile` property.

craciunoiuc:
correct

internal/builder/rootfs.go
```diff
@@ -524,8 +522,14 @@ func applyBuildOpts(attrs map[string]string, localDirs map[string]string, sessio
...
 
-	localDirs["context"] = opts.Rootfs.Path
-	localDirs["dockerfile"] = opts.Rootfs.Path
+	if opts.Rootfs.Dockerfile != "" {
+		localDirs["context"] = opts.Rootfs.Path
+		localDirs["dockerfile"] = filepath.Join(opts.Rootfs.Path, filepath.Dir(opts.Rootfs.Dockerfile))
+		attrs["filename"] = filepath.Base(opts.Rootfs.Dockerfile)
+	} else {
+		localDirs["context"] = filepath.Dir(opts.Rootfs.Path)
+		localDirs["dockerfile"] = filepath.Dir(opts.Rootfs.Path)
```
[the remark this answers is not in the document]
craciunoiuc: Validation done in library

internal/builder/rootfs.go
```diff
@@ -524,8 +522,14 @@ func applyBuildOpts(attrs map[string]string, localDirs map[string]string, sessio
 		attrs["platform"] = strings.Join(ps, ",")
 	}
 
-	localDirs["context"] = opts.Rootfs.Path
-	localDirs["dockerfile"] = opts.Rootfs.Path
+	if opts.Rootfs.Dockerfile != "" {
+		localDirs["context"] = opts.Rootfs.Path
+		localDirs["dockerfile"] = filepath.Join(opts.Rootfs.Path, filepath.Dir(opts.Rootfs.Dockerfile))
+		attrs["filename"] = filepath.Base(opts.Rootfs.Dockerfile)
+	} else {
```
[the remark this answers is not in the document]
craciunoiuc: I don't see why this would be bad

internal/builder/kraftfile_test.go
```diff
@@ -93,6 +99,100 @@ func TestKraftfileToBuildOptsRootfsPathJoined(t *testing.T) {
 		"rootfs path must be joined with the kraftfile directory")
 }
 
+func TestKraftfileToBuildOptsDockerfileWithType(t *testing.T) {
+	rootfsDir := t.TempDir()
+
+	runtime := kraftfile.Runtime("unikraft.io/unikraft.org/base")
```
jedevc, line 105: what's this image :eyes:

internal/builder/rootfs.go
```diff
@@ -179,13 +179,20 @@ func BuildRootfs(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image, rer
 		return buildRootfsTarball(ctx, opts)
 	case kraftfile.SourceTypeDockerfile:
 		opts.Rootfs.Format = cmp.Or(opts.Rootfs.Format, kraftfile.FsTypeCpio)
```
jedevc: Pre-existing issue: why is this here? Don't we have the `DefaultRootfsFormat` above?

internal/builder/rootfs.go
```diff
@@ -524,8 +531,14 @@ func applyBuildOpts(attrs map[string]string, localDirs map[string]string, sessio
...
 	}
 
-	localDirs["context"] = opts.Rootfs.Path
-	localDirs["dockerfile"] = opts.Rootfs.Path
+	if opts.Rootfs.Dockerfile != "" {
+		localDirs["context"] = opts.Rootfs.Path
+		localDirs["dockerfile"] = filepath.Join(opts.Rootfs.Path, filepath.Dir(opts.Rootfs.Dockerfile))
+		attrs["filename"] = filepath.Base(opts.Rootfs.Dockerfile)
+	} else {
+		localDirs["context"] = filepath.Dir(opts.Rootfs.Path)
```
jedevc, line 541: :eyes: Wait why are we changing the `context` to be the dir here? That's a change from before.

IMO, the `source` field in the file should now ideally never contain a path to a `Dockerfile`. We could have a check *somewhere* that if the path passed is a file, then take the parent directory as the directory, and take the filename as the "filename".

Essentially `source: app/foo.Dockerfile` (with no `dockerfile` set) would set `context` + `dockerfile` to `app/` and would set `filename` to `foo.Dockerfile`.

This is worth a little test.

internal/builder/kraftfile_test.go
```diff
@@ -93,6 +99,100 @@ func TestKraftfileToBuildOptsRootfsPathJoined(t *testing.T) {
...
+	opts, err := KraftfileToBuildOpts(rootfsDir, kf)
+	require.NoError(t, err)
+	require.Equal(t, rootfsDir+"/context", opts.Rootfs.Path)
+	require.Equal(t, "MyDockerfile", opts.Rootfs.Dockerfile)
+	require.Equal(t, kraftfile.SourceTypeDockerfile, opts.Rootfs.Type,
+		"type must be inferred as dockerfile when dockerfile field is set")
+	require.Equal(t, kraftfile.FsTypeErofs, opts.Rootfs.Format)
+}
+
+func TestKraftfileToBuildOptsAllFields(t *testing.T) {
```
jedevc: What's this test doing?

internal/builder/kraftfile_test.go
craciunoiuc: The image is correct, but it's definitely a weird choice, probably a hallucination 😔

internal/builder/kraftfile_test.go
craciunoiuc: Same as the one above, I'll remove

internal/builder/rootfs.go
craciunoiuc: Good question, will check

internal/builder/rootfs.go
craciunoiuc: Hmm, I think it's doing just that no? Your comment is on the `else` case so where `dockerfile` is empty, but we're in the dockerfile rootfs build.

This means that `source` points to a Dockerfile and was set to into the `path` by the library.

This means that the context and the `dockerfile` fields are correctly extracted.

What is missing is the `filename` being set to `filepath.Base(opts.Rootfs.Path)`. I will edit.

Now for the tests: basically all the existing tests that build dockerfiles fall in this branch so we have tests. I'll add one more that sets a different Dockerfile name to test that also.

craciunoiuc: Lemme know if fine now

craciunoiuc:
Test node dead, needs rerunning

internal/builder/rootfs.go
```diff
@@ -172,20 +172,24 @@ func BuildRootfs(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image, rer
...
+			if _, err := os.Stat(dockerfilePath); err != nil {
+				if os.IsNotExist(err) {
+					return nil, fmt.Errorf("dockerfile %q does not exist in context %q", opts.Rootfs.Dockerfile, opts.Rootfs.Path)
+				}
+				return nil, fmt.Errorf("checking dockerfile path %q: %w", dockerfilePath, err)
 			}
-			return nil, fmt.Errorf("checking dockerfile path %q: %w", opts.Rootfs.Path, err)
-		} else if !f.IsDir() {
-			opts.Rootfs.Path = filepath.Dir(opts.Rootfs.Path)
 		}
```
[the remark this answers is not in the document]
craciunoiuc: This is intended, user is responsible

internal/builder/rootfs.go
```diff
@@ -172,20 +172,24 @@ func BuildRootfs(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image, rer
...
 	case kraftfile.SourceTypeDockerfile:
-		opts.Rootfs.Format = cmp.Or(opts.Rootfs.Format, kraftfile.FsTypeCpio)
-		if f, err := os.Stat(opts.Rootfs.Path); err != nil {
+		if _, err := os.Stat(opts.Rootfs.Path); err != nil {
 			if os.IsNotExist(err) {
-				return nil, fmt.Errorf("dockerfile does not exist")
+				return nil, fmt.Errorf("dockerfile context does not exist")
+			}
+			return nil, fmt.Errorf("checking dockerfile context path %q: %w", opts.Rootfs.Path, err)
+		}
```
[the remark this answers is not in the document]
craciunoiuc: If the target is not a valid dockerfile, it will be resolved as a different builder

internal/builder/kraftfile.go
```diff
@@ -76,13 +76,16 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
...
-		opts.Rootfs.Path = filepath.Join(dir, kf.Rootfs.Source)
+		opts.Rootfs.Path = filepath.Join(dir, kf.Rootfs.Source.Path)
 		opts.Rootfs.Format = kf.Rootfs.Format
-		opts.Rootfs.Type = kf.Rootfs.Type
-		if opts.Rootfs.Type == "" {
+		opts.Rootfs.Type = kf.Rootfs.Source.Type
+		opts.Rootfs.Dockerfile = kf.Rootfs.Source.Dockerfile
+		if opts.Rootfs.Dockerfile != "" && opts.Rootfs.Type == "" {
+			opts.Rootfs.Type = kraftfile.SourceTypeDockerfile
+		} else if opts.Rootfs.Type == "" {
```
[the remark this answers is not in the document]
craciunoiuc: not needed, but will do

internal/builder/kraftfile.go
```diff
@@ -55,12 +55,12 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
...
 
 	for _, rom := range kf.Roms {
-		romPath := filepath.Join(dir, rom.Source)
+		romPath := filepath.Join(dir, rom.Source.Path)
 		romFormat := cmp.Or(rom.Format, kraftfile.FsTypeErofs)
 		romOpt := FSOpts{
 			Path:   romPath,
 			Format: romFormat,
-			Type:   rom.Type,
+			Type:   rom.Source.Type,
```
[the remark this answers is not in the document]
craciunoiuc: not needed, but will do

internal/builder/build_test.go
```diff
@@ -231,34 +230,6 @@ RUN ln -s /etc/passwd /another-test-link
 	}
 }
```
jedevc, line 233: Test gone???

jedevc (commented):
Little test comment, otherwise LGTM. Please test on the examples!

Also failing integration.

craciunoiuc:
integration node is again borked, I'm discussing to fix it

internal/builder/build_test.go
craciunoiuc: Readded

jedevc (approved):

PR #328, opened 2026-05-15
feat(cmd): Add '--follow' flag for start/restart

craciunoiuc:
I wanted to do this for instance create --set autostart=true also, but I was unsure about that (was a bit controversial in a way), and it was also messing up the creation code, so I took it out.

For the tests I added a timeout field so we can force-quit out of following logs after some time.

Depends on: https://github.com/unikraft-cloud/go-sdk/pull/187

Closes: TOOL-680


internal/cmd/run.go
```diff
@@ -128,10 +130,30 @@ func (c *RunCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *resource.
...
+		Offset: &zero,
+		Limit:  &one,
+	}})
+	if err != nil {
+		return 0, err
+	}
+	if len(resp.Data.Instances) == 0 {
+		return 0, nil
+	}
+	return resp.Data.Instances[0].Available.End, nil
```
craciunoiuc: I am not 100% sure if we should fetch the end and start from there or fetch it before starting/restarting and print from there?

For example for the unikraft kernel, which is very fast, it missed the whole boot prints.
Do we want this?

internal/cmd/instances.go
```diff
@@ -1434,7 +1457,14 @@ func (c *InstancesRestartCmd) Run(ctx context.Context, stdio config.Stdio) error
...
 	diffErr := cmd.Diff(ctx, stdio.Stdout, c.FormatOpts, Instance{}, before, updated)
-	return errors.Join(opErr, diffErr)
+	if err := errors.Join(opErr, diffErr); err != nil {
+		return err
+	}
+	if !c.Follow || len(started) == 0 {
+		return nil
+	}
+	fmt.Fprintln(stdio.Stdout)
+	return streamInstanceLogs(ctx, stdio, started, 0, true, true)
```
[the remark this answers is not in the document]
craciunoiuc: Copilot also said this. I wonder what we want here?

internal/cmd/run.go
craciunoiuc: I decided we should actually print everything:
for start -> fetch the current log end -> start -> print from there
for restart -> stop -> fetch the current log end -> start -> print from there

internal/cmd/run.go
```diff
@@ -165,3 +175,41 @@ func streamInstanceLogs(ctx context.Context, stdio config.Stdio, keys multimetro
...
+	offsets := make(map[string]int64, len(keys))
+	err := group.DoRefs(ctx, g, keys.Refs(), func(_ context.Context, c multimetro.MetroClient, refs group.Refs) (group.Refs, error) {
+		for _, ref := range refs {
+			offset, loopErr := instanceLogEnd(ctx, c, ref.NameOrUUID())
+			if loopErr != nil {
+				return nil, loopErr
+			}
+			// ref.Metro is empty when input keys have no metro prefix
+			key := multimetro.Key{Metro: c.Metro.Name, Name: ref.Name, UUID: ref.UUID}
+			offsets[key.Canonical()] = offset
```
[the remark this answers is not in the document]
craciunoiuc: Not needed, as long as the instance is referenced correctly

craciunoiuc:
As good as it can get

internal/cmd/run.go
```diff
@@ -165,3 +175,48 @@ func streamInstanceLogs(ctx context.Context, stdio config.Stdio, keys multimetro
...
+	if len(offsets) == 0 {
+		return nil, fmt.Errorf("no valid instances found for log streaming")
+	}
+	return offsets, nil
+}
+
+func instanceLogEnd(ctx context.Context, c multimetro.MetroClient, id platform.NameOrUUID) (int64, error) {
+	zero := int64(0)
+	one := int64(1)
+	resp, err := c.GetInstanceLogs(ctx, []platform.GetInstancesLogsRequestItem{{
```
jedevc: This should be a helper in `x/logs`. Or we should use something there. I think this would be similar to the logic that we currently have for `--tail` there, where it tries rewinding - but I guess we'd just rewind "0".

internal/cmd/run.go
craciunoiuc: Discussed, it's ok

internal/cmd/run.go
craciunoiuc: Open in https://github.com/unikraft-cloud/go-sdk/pull/187 now

craciunoiuc: Ok, your rewind suggestion and PR made more sense, I'll use that

cmd/unikraft/integration/instance_test.go
```diff
@@ -146,6 +146,59 @@ func TestInstances(t *testing.T) {
...
+
+		// Start it and follow logs; killed after 5 s is treated as success.
+		r.Run(t, []string{
+			"unikraft", "instance", "start",
+			"--follow",
+			"test-" + instName,
+		}, integ.WithTimeout(5*time.Second))
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
+	})
```
jedevc, line 216: Hm, these tests are not sufficient.

Ideally, the image your print an incrementing string on boot like `starting 1`, `starting 2`, etc (maybe using the `busybox` image).

Then we should check that when using `--follow` that the output contains `starting 2`, but doesn't contain `starting 1`.

cmd/unikraft/integration/instance_test.go
jedevc, line 1: @craciunoiuc another test idea.

It's super useful to have a `busybox` image for writing tests - it's got a bunch of utilities. But we wastefully build it again and again and again.

We should have some sort of a `busyboxImage(testing.T) string` that uses `sync.Once` to only build the image once, and return it once. It should also use `t.Cleanup` to tear the image down. That way, we only build busybox once.

The tests that builds actually work - those should really be in the `internal/builder` tests (and maybe also `build_test.go`).

Could you do that refactoring and open a PR?

internal/cmd/run.go
```diff
@@ -128,40 +128,53 @@ func (c *RunCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *resource.
...
+		mux.Close()
+		return nil, nil, err
 	}
 	mux.Seal()
 
-	_, err = io.Copy(stdio.Stdout, mux)
+	return mux, cancel, nil
+}
+
+func drainInstanceLogs(stdio config.Stdio, mux *muxreader.Mux, follow bool) error {
```
jedevc: I would just move this inline to all the places that need it. It reads easier to see `io.Copy` in those places.

internal/cmd/run.go
```diff
@@ -128,40 +128,53 @@ func (c *RunCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *resource.
...
+	_, err := io.Copy(stdio.Stdout, mux)
 	if follow && errors.Is(err, context.Canceled) {
 		return nil
 	}
 	return err
 }
+
+func streamInstanceLogs(ctx context.Context, stdio config.Stdio, keys multimetro.Keys, tail *int, follow bool) error {
```
jedevc: Also just move this one inline, there's only one caller.

go.mod
```diff
@@ -57,10 +57,12 @@ require (
 	unikraft.com/x/joinerrgroup v0.0.0-20260304162956-523940cab1de
 	unikraft.com/x/kingkong v0.0.0-20260331102539-2c733927b46f
 	unikraft.com/x/kraftfile v0.0.0-20260501125416-502eb6c9c39d
-	unikraft.com/x/log v0.0.0-20260126171022-af62c17fcdf7
+	unikraft.com/x/log v0.0.0-20260527070319-f904d21cd6ea
 	unikraft.com/x/ptr v0.0.0-20260126094137-ab6e717e5679
 )
 
+require go.opentelemetry.io/otel/log v0.19.0 // indirect
```
jedevc: :eyes: this should be in the block below.

internal/integration/env.go
```diff
@@ -101,6 +118,12 @@ func (env *TestEnv) RunRaw(t *testing.T, args []string, opts ...CmdOption) (stri
 	c.Env = append(c.Env, resource.UnikraftSandboxEnv+"="+env.sandboxPath)
 
 	err := c.Run()
+	// A command killed by its own timeout is treated as success.
+	var killedErr *exec.ExitError
+	if cfg.timeout > 0 && errors.Is(cmdCtx.Err(), context.DeadlineExceeded) &&
+		errors.As(err, &killedErr) && killedErr.ExitCode() == -1 {
+		err = nil
+	}
```
jedevc: :eyes: why? If a command timed out, then we shouldn't just succeed.

It should be up to the caller to handle this.

jedevc (commented):
Looks good, some comments :tada: Thanks for taking a stab at this!

internal/integration/env.go
craciunoiuc: I think I was gaslit by copilot reviews if I remember correctly 💫

cmd/unikraft/integration/instance_test.go
craciunoiuc: Ok so a separate PR, got it.

cmd/unikraft/integration/instance_test.go
craciunoiuc: so you mean:
* `print starting 1`
* `follow logs`
* `stop the instance`
* `start the instance`
* `print starting 2`

Ok will do

craciunoiuc: Added, I think it works well now. Had to use a volume to survive data after boot (basically instance increments a counter in the logs on every boot and prints it)

If things work correctly, on the second boot you should see only `starting 2`

cmd/unikraft/integration/instance_test.go
craciunoiuc: Work for this is here now https://github.com/unikraft/cli/pull/350

Closing comment here

internal/integration/env.go
```diff
@@ -101,6 +119,9 @@ func (env *TestEnv) RunRaw(t *testing.T, args []string, opts ...CmdOption) (stri
 	c.Env = append(c.Env, resource.UnikraftSandboxEnv+"="+env.sandboxPath)
 
 	err := c.Run()
+	if cfg.timeout > 0 && errors.Is(cmdCtx.Err(), context.DeadlineExceeded) {
+		err = context.DeadlineExceeded
+	}
```
jedevc: Why? 🤔 Why strip error info?

internal/cmd/instances.go
```diff
@@ -1283,6 +1291,19 @@ func (c *InstancesStartCmd) Run(ctx context.Context, stdio config.Stdio) error {
...
+		mux    *muxreader.Mux
+		cancel context.CancelFunc
+	)
+	if c.Follow {
+		mux, cancel, err = newInstanceLogMux(ctx, keys, new(0), true)
+		if err != nil {
+			return err
+		}
+		defer cancel()
+		defer mux.Close()
```
jedevc: nit: maybe reverse these? Since closing conceptually might use the context, or try and leave things in a good state.

internal/cmd/instances.go
```diff
@@ -1283,6 +1291,19 @@ func (c *InstancesStartCmd) Run(ctx context.Context, stdio config.Stdio) error {
 		return err
 	}
 
+	var (
+		mux    *muxreader.Mux
+		cancel context.CancelFunc
```
jedevc: `cancel` can appear in the `if c.Follow` block.

internal/cmd/instances.go
```diff
@@ -1309,7 +1330,19 @@ func (c *InstancesStartCmd) Run(ctx context.Context, stdio config.Stdio) error {
...
+	if !c.Follow || len(started) == 0 {
+		return nil
+	}
+	fmt.Fprintln(stdio.Stdout)
+
+	_, err = io.Copy(stdio.Stdout, mux)
+	if c.Follow && errors.Is(err, context.Canceled) {
+		return nil
+	}
+	return err
```
jedevc, line 1350: Alternatively.

Just do the `io.Copy` conditional on `mux != nil`. Then we only check `c.Follow` in one place, which is neater.

jedevc (commented):
Nice! Only some small nits.

internal/integration/env.go
craciunoiuc: typo

I guess I'll just remove that check if it's just keeping the error

jedevc (approved):
```
```

PR #329, opened 2026-05-19
feat(internal): Show boot time field by default

craciunoiuc:
Shows boot time by default.

Shows at the end right now. Maybe we should move?

Bonus thing: I find it ugly that values are not rounded to two decimals. Especially since you use the nearest measurement unit, you can get things like `1.0858s` boot time. Should we change this? 
**I added it as a separate commit, let me know.**

~~TODO: Integration update once it starts behaving. Will mark as ready afterwards~~

Closes: TOOL-853
Supersedes: #250 
Closes: #250 

It's basically the same thing as #250 but that was not merged for some reason.
I will adopt and adapt it here.

### DIFF
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -173,7 +173,7 @@ type Instance struct {
 
 	Timing struct {
 		Uptime   types.DurationMS `mirror:"instance.uptime_ms"`
-		BootTime types.DurationUS `mirror:"instance.boot_time_us"`
+		BootTime types.DurationUS `mirror:"instance.boot_time_us" field:",long"`
 		NetTime  types.DurationUS `mirror:"instance.net_time_us"`
 	}
 
--- a/internal/types/duration.go
+++ b/internal/types/duration.go
@@ -94,7 +94,11 @@ func (d DurationMS) MarshalJSON() ([]byte, error) {
 }
 
 func (d DurationMS) String() string {
-	return (time.Duration(d) * time.Millisecond).String()
+	dur := time.Duration(d) * time.Millisecond
+	if d >= 1000 {
+		dur = dur.Round(10 * time.Millisecond)
+	}
+	return dur.String()
 }
 
 // DurationUS is a time wrapper that represents a duration in microseconds.
@@ -137,5 +141,12 @@ func (d DurationUS) MarshalJSON() ([]byte, error) {
 }
 
 func (d DurationUS) String() string {
-	return (time.Duration(d) * time.Microsecond).String()
+	dur := time.Duration(d) * time.Microsecond
+	switch {
+	case d >= 1_000_000: // >= 1s: round to 10 milliseconds
+		dur = dur.Round(10 * time.Millisecond)
+	case d >= 1000: // >= 1ms: round to 10 microseconds
+		dur = dur.Round(10 * time.Microsecond)
+	}
+	return dur.String()
 }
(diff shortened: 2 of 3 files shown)


internal/types/duration.go
```diff
@@ -94,7 +94,11 @@ func (d DurationMS) MarshalJSON() ([]byte, error) {
 }
 
 func (d DurationMS) String() string {
-	return (time.Duration(d) * time.Millisecond).String()
+	dur := time.Duration(d) * time.Millisecond
+	if d >= 1000 {
+		dur = dur.Round(10 * time.Millisecond)
+	}
+	return dur.String()
```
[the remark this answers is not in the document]
craciunoiuc: Intended

internal/cmd/instances.go
```diff
@@ -175,7 +175,7 @@ type Instance struct {
 
 	Timing struct {
 		Uptime   types.DurationMS `mirror:"instance.uptime_ms"`
-		BootTime types.DurationUS `mirror:"instance.boot_time_us"`
+		BootTime types.DurationUS `mirror:"instance.boot_time_us" field:",short"`
```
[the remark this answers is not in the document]
craciunoiuc: Yes this needs updating once the tests work

craciunoiuc:
Tests need to be rerun because the test machine cannot be reached

internal/cmd/instances.go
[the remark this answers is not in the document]
nderjung: I originally put it in the long output, not short, because the real estate on the terminal is limited.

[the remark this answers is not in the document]
craciunoiuc: fine by me

internal/cmd/instances.go
```diff
@@ -171,7 +171,7 @@ type Instance struct {
 
 	Timing struct {
 		Uptime   types.DurationMS `mirror:"instance.uptime_ms"`
-		BootTime types.DurationUS `mirror:"instance.boot_time_us"`
```
jedevc, line 176: Hm, this should result in a change to the `table` output right?

internal/cmd/instances.go
jedevc, line 1: Unrelated nit: https://github.com/unikraft/cli/blob/bd7caab08597b7ebd53a9bd15724e0362c19d0fe/internal/cmd/testdata/TestOutput/instances#L101-L103

Why is `example.unikraft.app` repeated twice?

That's one of the reasons this line is super long.

Also `CREATED` is `never` which is fairly unhelpful.

Looks like this test needs some fixes anyways.

jedevc (commented):
I'm happy with this, but, the tests don't seem right.

internal/cmd/instances.go
craciunoiuc: Alex suggested that this should be moved to `long`, hence why it doesn't show up in the short table

So which one should it be? I personally find boot time better to show in the short output than `ARGS` for example
But maybe we could just show both?

internal/cmd/instances.go
craciunoiuc: Weird, I can check

craciunoiuc: found it, really weird, I'll add as a separate commit

internal/cmd/instances.go
jedevc: I think we should show in just long output, until we have a better support for https://github.com/unikraft/cli/issues/372.

Since without this, everything will really start overflowing and looking weird.

jedevc (approved):
```
```

PR #333, opened 2026-05-21
fix(cmd): Fallback to old API on listing

craciunoiuc:
This is to accommodate old nodes with the CLI.
Ideally we can remove this code in the near future.
For now add it back as fallback.

### DIFF
--- a/internal/cmd/images.go
+++ b/internal/cmd/images.go
@@ -353,8 +353,13 @@ func listPlatformImages(ctx context.Context) ([]resource.Resource, error) {
 
 		resp, err := c.GetImageStore(ctx, nil, platform.GetImageStoreOpts{})
 		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/cmd/images.go
```diff
@@ -353,8 +353,13 @@ func listPlatformImages(ctx context.Context) ([]resource.Resource, error) {
 
 		resp, err := c.GetImageStore(ctx, nil, platform.GetImageStoreOpts{})
 		if err != nil {
-			log.G(ctx).Trace().Err(err).Msg("skipping image-store listing")
-			return nil, nil
+			log.G(ctx).Debug().Err(err).Msg("image-store listing failed, falling back to /v1/images")
```
jedevc: `Warn()`.

jedevc (approved):

PR #336, opened 2026-05-22
feat: Add api subcommand

jedevc:
Kinda like `gh api`. It's a nice half-way point between accessing the API raw and using the nice functionality in the rest of the CLI.

It uses the metros and config of the CLI, sets headers correctly, etc, and so should be a definite improvement over `curl` based commands.

### DIFF
--- a/internal/cmd/api.go
+++ b/internal/cmd/api.go
@@ -1,0 +1,242 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+
+	jujuerrors "github.com/juju/errors"
+	"unikraft.com/x/kingkong"
+	"unikraft.com/x/log"
+	"unikraft.com/x/ptr"
+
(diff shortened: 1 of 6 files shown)


internal/cmd/api.go
```diff
@@ -0,0 +1,212 @@
...
+	// Resolve the request body.
+	var body io.Reader
+	if c.Data != "" {
+		switch {
+		case c.Data == "@-":
+			body = stdio.Stdin
+		case strings.HasPrefix(c.Data, "@"):
+			f, err := os.Open(c.Data[1:])
+			if err != nil {
+				return jujuerrors.Annotate(err, "reading request body")
```
craciunoiuc, line 109: We're using juju?

craciunoiuc (commented):
Looks good, only one comment + the copilot ones

internal/cmd/api.go
jedevc: Always have been :eyes:

https://github.com/unikraft/cli/blame/93cc71532b12d80e71fef353a4eb4def5a4d6d29/cmd/unikraft/main.go#L19-L19

Alex set it up like this originally.

internal/cmd/api.go
```diff
@@ -0,0 +1,218 @@
...
+		} else {
+			if _, err := stdio.Stdout.Write(raw); err != nil {
+				return err
+			}
+			if !bytes.HasSuffix(raw, []byte("\n")) {
+				fmt.Fprintln(stdio.Stdout)
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
jedevc: Gonna ignore this, the results are small.

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #337, opened 2026-05-22
feat(cmd): Add quotas command

craciunoiuc:
Interactive terminal that shows per-node quotas.

~~The controlplane is used for fetching image data, which does not depend on a node for fetching.~~

Depends on: https://github.com/unikraft-cloud/console/pull/775
Depends on: https://github.com/unikraft-cloud/proto/pull/280

TODOs left:
- go sdk bump after https://github.com/unikraft-cloud/proto/pull/280
- Add image quota bar after https://github.com/unikraft-cloud/console/pull/775
- Revisit renderer after https://github.com/unikraft/cli/pull/334

Closes: TOOL-541

### DIFF
--- a/internal/cmd/metro_quotas.go
+++ b/internal/cmd/metro_quotas.go
@@ -1,0 +1,556 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package cmd
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"io"
+	"math"
+	"slices"
+	"strings"
+	"time"
+
+	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
+	"charm.land/lipgloss/v2/compat"
+	"github.com/docker/go-units"
+	"golang.org/x/sync/errgroup"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/kvwriter"
+	"unikraft.com/cli/internal/multimetro"
+	uitui "unikraft.com/cli/internal/tui/uitui"
+	xio "unikraft.com/cli/internal/x/io"
+	"unikraft.com/cloud/sdk/platform"
+	"unikraft.com/cloud/sdk/platform/group"
+	"unikraft.com/x/colors"
+	"unikraft.com/x/log"
+)
+
+// MetroQuotasCmd displays quota usage for metros in an interactive TUI.
+// Tabs allow switching between individual metros.
+type MetroQuotasCmd struct {
+	Metro []string       `short:"m" help:"Show quotas for specific metros only." placeholder:"metro"`
+	Watch *time.Duration `short:"w" help:"Watch for changes and refresh output. Defaults to 2s." type:"optional" placeholder:"duration"`
+}
+
+func (cmd *MetroQuotasCmd) Run(ctx context.Context, stdio config.Stdio) error {
+	profile, err := config.G(ctx).CurrentProfile()
+	if err != nil {
+		return fmt.Errorf("failed to load profile: %w", err)
+	}
+
+	if _, err := selectedMetros(profile, cmd.Metro); err != nil {
+		return err
+	}
+
+	var watch *time.Duration
+	if cmd.Watch != nil {
+		w := cmp.Or(*cmd.Watch, 2*time.Second)
+		watch = &w
+	}
+
+	if xio.IsTTY(stdio.Stdout) {
+		m := newQuotasModel(ctx, cmd.Metro, watch)
+		p := tea.NewProgram(m, tea.WithInput(stdio.Stdin), tea.WithOutput(xio.Unwrap(stdio.Stdout)))
+		_, err := p.Run()
+		return err
+	}
+	if cmd.Watch != nil {
+		return cmd.watchRenderLoop(ctx, stdio.Stdout, *watch)
+	}
+	return cmd.renderOnce(ctx, stdio.Stdout)
+}
+
+func (cmd *MetroQuotasCmd) watchRenderLoop(ctx context.Context, out io.Writer, interval time.Duration) error {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		if err := cmd.renderOnce(ctx, out); err != nil {
+			return err
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			_, _ = fmt.Fprint(out, "\n")
+		}
+	}
+}
+
+func (cmd *MetroQuotasCmd) renderOnce(ctx context.Context, out io.Writer) error {
+	result := fetchAllMetros(ctx, cmd.Metro)
+	if result.err != nil {
+		return result.err
+	}
+	if len(result.data) == 0 {
+		return fmt.Errorf("no quota data available")
+	}
+	_, err := fmt.Fprint(out, renderQuotaSections(result.data, result.errors, result.tabs, result.userName))
+	return err
+}
+
+func renderQuotaSections(data map[string]*platform.Quotas, errors map[string]error, tabs []string, userName string) string {
+	showMetroHeader := len(tabs) > 1
+	var b strings.Builder
+	for i, tab := range tabs {
+		q := data[tab]
+		if showMetroHeader {
+			fmt.Fprintf(&b, "metro: %s\n\n", tab)
+		}
+		if q != nil {
+			b.WriteString(renderQuotaView(q, userName, defaultQuotaBarWidth))
+		} else if err := errors[tab]; err != nil {
+			b.WriteString(uitui.ErrorStyle.Render(err.Error()))
+			b.WriteString("\n")
+		} else {
+			b.WriteString("no quota data available\n")
+		}
+		if i < len(tabs)-1 {
+			b.WriteString("\n\n")
+		}
+	}
+	return b.String()
+}
+
+// multiMetroResult holds results from fetching multiple metros.
+type multiMetroResult struct {
+	data     map[string]*platform.Quotas
+	errors   map[string]error
+	tabs     []string
+	userName string
+	err      error
+}
+
+type quotasModel struct {
+	ctx           context.Context
+	metros        []string
+	watchInterval *time.Duration
+	lastRefresh   time.Time
+	tabs          []string
+	activeTab     int
+	termWidth     int
+	termHgt       int
+	data          map[string]*platform.Quotas
+	errors        map[string]error
+	userName      string
+	loading       bool
+	err           error
+}
+
+type (
+	quotasLoadedMsg = multiMetroResult
+	quotasTickMsg   struct{}
+	quotasStatusMsg struct{}
+)
+
+const (
+	defaultQuotaBarWidth = 36
+	minQuotaBarWidth     = 10
+)
+
+func computeQuotaBarWidth(termWidth int) int {
+	if termWidth <= 0 {
+		return defaultQuotaBarWidth
+	}
+	const reserved = 42
+	w := termWidth - reserved
+	return max(minQuotaBarWidth, min(w, defaultQuotaBarWidth))
+}
+
+func newQuotasModel(ctx context.Context, metros []string, watchInterval *time.Duration) quotasModel {
+	var tabs []string
+	if len(metros) == 1 {
+		tabs = []string{metros[0]}
+	}
+	return quotasModel{
+		ctx:           ctx,
+		metros:        metros,
+		watchInterval: watchInterval,
+		tabs:          tabs,
+		data:          make(map[string]*platform.Quotas),
+		errors:        make(map[string]error),
+		loading:       true,
+	}
+}
+
+func (m quotasModel) Init() tea.Cmd {
+	if m.watchInterval != nil {
+		return tea.Batch(m.fetchCmd(), m.watchTickCmd(), m.watchStatusTickCmd())
+	}
+	return m.fetchCmd()
+}
+
+func (m quotasModel) fetchCmd() tea.Cmd {
+	ctx := m.ctx
+	metros := slices.Clone(m.metros)
+	return func() tea.Msg {
+		return fetchAllMetros(ctx, metros)
+	}
+}
+
+func (m quotasModel) watchTickCmd() tea.Cmd {
+	if m.watchInterval == nil {
+		return nil
+	}
+	return tea.Tick(*m.watchInterval, func(time.Time) tea.Msg {
+		return quotasTickMsg{}
+	})
+}
+
+func (m quotasModel) watchStatusTickCmd() tea.Cmd {
+	if m.watchInterval == nil {
+		return nil
+	}
+	return tea.Tick(50*time.Millisecond, func(time.Time) tea.Msg {
+		return quotasStatusMsg{}
+	})
+}
+
+func (m quotasModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg := msg.(type) {
+	case quotasLoadedMsg:
+		m.loading = false
+		if msg.err != nil {
+			m.err = msg.err
+			return m, nil
+		}
+		m.lastRefresh = time.Now()
+		m.userName = msg.userName
+		m.data = msg.data
+		m.errors = msg.errors
+		m.tabs = msg.tabs
+
+		if len(m.tabs) == 0 && len(m.metros) == 1 {
+			m.tabs = []string{m.metros[0]}
+		}
+		if m.activeTab >= len(m.tabs) {
+			m.activeTab = 0
+		}
+		return m, nil
+
+	case quotasTickMsg:
+		m.loading = true
+		return m, tea.Batch(m.fetchCmd(), m.watchTickCmd())
+
+	case quotasStatusMsg:
+		return m, m.watchStatusTickCmd()
+
+	case tea.KeyPressMsg:
+		switch msg.Keystroke() {
+		case "ctrl+c", "q", "escape":
+			return m, tea.Quit
+		case "tab", "right", "l":
+			if len(m.tabs) > 1 {
+				m.activeTab = (m.activeTab + 1) % len(m.tabs)
+			}
+		case "shift+tab", "left", "h":
+			if len(m.tabs) > 1 {
+				m.activeTab = (m.activeTab - 1 + len(m.tabs)) % len(m.tabs)
+			}
+		}
+
+	case tea.WindowSizeMsg:
+		m.termWidth = msg.Width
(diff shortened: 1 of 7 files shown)


jedevc:
If you want to split this up, you might find it easier to just have the local quotas in each tab? We can work on the global tab later (and then work on console+proto changes separately)

craciunoiuc:
That's how it is now

So you want me to open a separate PR for the global stats?

Should we also only show the  image quota in the global tab?

jedevc:
Yes image quota would be in global tab.

Just thinking of what we need to do to unblock this PR and get it out of draft.

craciunoiuc:
Hmm, theoretically the PR is not blocked on this. We could merge it and then slightly change the view depending on what you want to see @jedevc 

Right now `All` adds the current and the max everywhere

If that's too controversial we can just remove it completely and rethink :D

image
image

jedevc:
Ah I had assumed the PR was blocked because it was in draft :sweat_smile:

craciunoiuc:
ah no, I just wanted to have a final version before putting it out there

\+ I'd rather we review/merge the ones that are in `Ready` but up to you of course :D

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+				entries[i] = metroQuotaResult{metro: m.Name, err: fmt.Errorf("%s: %w", m.Name, err)}
+				return
+			}
+			if resp.Data == nil || len(resp.Data.Quotas) == 0 {
+				entries[i] = metroQuotaResult{metro: m.Name, err: fmt.Errorf("%s: no quota data", m.Name)}
+				return
+			}
+			entries[i] = metroQuotaResult{metro: m.Name, quotas: &resp.Data.Quotas[0]}
+		})
+	}
```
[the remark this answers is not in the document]
craciunoiuc: Not needed

internal/cmd/metros.go
```diff
@@ -28,6 +28,8 @@ type MetrosCmd struct {
 	cmd.ResourceCmd[Metro]
 	cmd.GettableResourceCmd[Metro]
 	cmd.ListableResourceCmd[Metro]
+
+	Quotas MetroQuotasCmd `cmd:"" help:"Display quota usage for metros."`
```
jedevc: IMO, this should be a top-level command. `unikraft quota` / `unikraft quotas` (allow an `alias`).

Then it matches what's expected by users.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+)
+
+// MetroQuotasCmd displays quota usage for metros in an interactive TUI.
+// Tabs allow switching between individual metros. Use --watch for periodic refresh.
+// Note: non-interactive mode requires a specific metro (--metro flag).
+type MetroQuotasCmd struct {
+	Watch         *time.Duration `short:"w" help:"Watch for changes and refresh output. Defaults to 2s." type:"optional" placeholder:"duration"`
+	Metro         string         `short:"m" help:"Show quotas for a specific metro only." placeholder:"metro"`
+	NoInteractive bool           `name:"no-interactive" short:"n" help:"Print stats once without entering the interactive TUI."`
+}
```
jedevc, line 40: :eyes: I don't love all these flags. I think we can have *just* `--metro` (without the integration with interactivity).

No reason to allow `--non-interactive`, we should always be interactive (even with `--metro`, for consistency - other commands don't have this non-interactive property). Though, note, we should always degrade elegantly when stdout is not a TTY, so `unikraft metro quotas | cat` should work (if there are multiple metros, just print them out in a row).

I think having a separate `--watch` is good though.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	if len(metros) == 0 {
+		return multiMetroResult{
+			err: fmt.Errorf("no metros configured in profile %q", profile.Name),
+		}
+	}
+
+	entries := make([]metroQuotaResult, len(metros))
+	var wg sync.WaitGroup
+	for i := range metros {
+		m := metros[i] // capture loop variable for goroutine safety
```
jedevc: Nope, modern go doesn't need this!

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+		wg.Go(func() {
```
jedevc: `errgroup` please!

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	loading   bool
+	err       error
+	watch     *time.Duration
+}
+
+type quotasLoadedMsg = multiMetroResult
+
+type quotasTickMsg struct{}
+
+func newQuotasModel(ctx context.Context, watch *time.Duration, metro string) quotasModel {
```
jedevc: IMO, we should make this a more generic model (like the watcher already is).

It should take just tabs of strings, be populated, and then just run. Then we can reuse the tabs model later for other data, I can see it being useful in other commands as well.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	if cmd.Watch != nil {
+		interval := cmp.Or(*cmd.Watch, 2*time.Second)
+		return watcher.WatchOutput(ctx, interval, stdio.Stdout, render)
+	}
+	return render(stdio.Stdout)
+}
+
+// preCheck validates configuration before entering the TUI. Returns an error
+// if the setup is invalid so we can fail fast instead of hanging in a broken TUI.
+func (cmd *MetroQuotasCmd) preCheck(ctx context.Context) error {
```
jedevc: Why not inline this? There's just one caller, having it split out just makes the code harder to follow.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	"unikraft.com/x/log"
+	"unikraft.com/x/ptr"
+)
+
+// MetroQuotasCmd displays quota usage for metros in an interactive TUI.
+// Tabs allow switching between individual metros. Use --watch for periodic refresh.
+// Note: non-interactive mode requires a specific metro (--metro flag).
+type MetroQuotasCmd struct {
+	Watch         *time.Duration `short:"w" help:"Watch for changes and refresh output. Defaults to 2s." type:"optional" placeholder:"duration"`
+	Metro         string         `short:"m" help:"Show quotas for a specific metro only." placeholder:"metro"`
```
jedevc: `--metro` should be a `[]string`, so you can select a subset of metros.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+var (
+	quotaTabActive = lipgloss.NewStyle().
+			Bold(true).
+			Foreground(lipgloss.Color("#ffffff")).
+			Background(lipgloss.Color("#7c3aed")).
+			Padding(0, 1)
+	quotaTabInactive = lipgloss.NewStyle().
+				Faint(true).
+				Padding(0, 1)
+)
```
jedevc: Use `x/colors` for consistency.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	if limit <= 0 {
+		return quotaBarEmptyStyle.Render(strings.Repeat(" ", quotaBarWidth))
+	}
+	filled := int(math.Floor(float64(used) / float64(limit) * float64(quotaBarWidth)))
+	filled = max(0, min(filled, quotaBarWidth))
+
+	var c color.Color
+	pct := float64(used) / float64(limit)
+	switch {
+	case pct > 0.83:
```
jedevc: Magic number.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+		// fuchsia: violet — darker on light terminals for contrast
+		c = compat.AdaptiveColor{Light: lipgloss.Color("#a21caf"), Dark: lipgloss.Color("#e879f9")}
```
jedevc: `x/colors`

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+			b.WriteString(quotaTabInactive.Render(tab))
+		}
+		if i < len(m.tabs)-1 {
+			b.WriteString(" ")
+		}
+	}
+	return b.String()
+}
+
+const quotaBarWidth = 36
```
jedevc: What if this is longer than the screen width?

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	return units.BytesSize(float64(mb) * units.MiB)
+}
+
+type quotaRow struct {
+	label string
+	bar   bool
+	used  int64
+	limit int64
+	value string
+	isMB  bool
```
jedevc: Just pass a function pointer here instead.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	return bar + empty
+}
+
+func quotaFormatSizeMB(mb int64) string {
+	return units.BytesSize(float64(mb) * units.MiB)
+}
+
+type quotaRow struct {
+	label string
+	bar   bool
```
jedevc: We can derive this depending on whether `used` + `limit` are used. We don't need a separate `bar` right?

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+			{label: "active used memory", bar: true, used: q.Used.LiveMemoryMb, limit: q.Hard.LiveMemoryMb, isMB: true},
+			{label: "memory size limits", value: fmt.Sprintf("%s-%s", quotaFormatSizeMB(q.Limits.MinMemoryMb), quotaFormatSizeMB(q.Limits.MaxMemoryMb))},
+		},
+		{
+			{label: "exposed services", bar: true, used: q.Used.Services, limit: q.Hard.Services},
+			{label: "services", bar: true, used: q.Used.ServiceGroups, limit: q.Hard.ServiceGroups},
+		},
+	}
+
+	enabledStyle := lipgloss.NewStyle().Foreground(compat.AdaptiveColor{Light: lipgloss.Color("#4338ca"), Dark: lipgloss.Color("#818cf8")})
```
jedevc: `x/colors`. Might be worth adding this to `AGENTS.md` (along with some of the other comments I've made, I keep having to repeat some of them).

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+			err: fmt.Errorf("metro %q not found in profile", metro),
+		}
+	}
+
+	httpClient := httpclient.GetClient(ptr.ZeroIfNil(selected.Insecure))
+	client := platform.NewClient(
+		platform.WithHTTPClient(httpClient),
+		platform.WithToken(profile.Token),
+		platform.WithDefaultMetro(selected.Endpoint),
+	)
```
jedevc: Can't we use the `multimetro` here? With `group.DoMetro`? Like we do everywhere else, instead of manually constructing all the clients?

That handles the concurrency + error resolution.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+	var wg sync.WaitGroup
+	for i := range metros {
+		m := metros[i] // capture loop variable for goroutine safety
+		wg.Go(func() {
+			httpClient := httpclient.GetClient(ptr.ZeroIfNil(m.Insecure))
+			client := platform.NewClient(
+				platform.WithHTTPClient(httpClient),
+				platform.WithToken(profile.Token),
+				platform.WithDefaultMetro(m.Endpoint),
+			)
```
jedevc: Code duplication with single metro.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,559 @@
...
+		return multiMetroResult{err: firstErr}
+	}
+
+	return multiMetroResult{
+		data:     data,
+		tabs:     tabs,
+		userName: userName,
+		// Keep partial-success behavior: show successful metros even when
+		// some metros fail to fetch.
+		err: nil,
```
jedevc: Not needed. Go defaults to zero values.

internal/cmd/metros.go
craciunoiuc: Hmm, sure thing, I have no preference on this 😅

internal/cmd/metro_quotas.go
craciunoiuc: Hmm ok, yeah `--non-intereactive` was already redundant so yeah

ok we remove `--watch`

internal/cmd/metro_quotas.go
craciunoiuc: I always forget about this

internal/cmd/metro_quotas.go
craciunoiuc: True, will do

internal/cmd/metro_quotas.go
craciunoiuc: For some reason I thought that is only for running commands on metros lol

internal/cmd/metro_quotas.go
craciunoiuc: It was a magic kraft number, that's the value we had it at before

I'll set some consts at the top

internal/cmd/metro_quotas.go
craciunoiuc: sure thing

internal/cmd/metro_quotas.go
craciunoiuc: It was fixed size in kraft also. I can make it refuse to run or make it change length on small screens

What do you think?

Kraft:
image

internal/cmd/metro_quotas.go
craciunoiuc: This I need to check

internal/cmd/metro_quotas.go
craciunoiuc: Hmm, this i need to investigate

internal/cmd/metro_quotas.go
jedevc: > I think having a separate `--watch` is good though.

:eyes: We can have `--watch`!

internal/cmd/metro_quotas.go
jedevc: Let's change lengths on small screens. We don't need to inherit all kraftkit bugs.

internal/cmd/metro_quotas.go
craciunoiuc: I pray the agent understood what you meant cause I didn't :(

internal/cmd/metro_quotas.go
craciunoiuc: I will make it behave like `btop` hehe

jedevc: Ideally, it should also grow and shrink as we resize :)))

internal/cmd/metro_quotas.go
craciunoiuc: Oh no, I understood a different watch, rip, readding

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,551 @@
...
+
+type (
+	quotasLoadedMsg = multiMetroResult
+	quotasTickMsg   struct{}
+	quotasStatusMsg struct{}
+)
+
+const quotaBarWidth = 36
+
+func newQuotasModel(ctx context.Context, metros []string, watchInterval time.Duration, watchEnabled bool) quotasModel {
```
jedevc: Why two `watch` parameters? Why not `*time.Duration`?

internal/cmd/metro_quotas.go
jedevc, line 1:
```
❯ unikraft quotas
  Terminal too small for quota view. Increase terminal size.

  Current: width=80 height=24
  Needed:  width=69 height=25
```

What? :eyes: that's the default terminal size.

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,551 @@
...
+	results := make([]struct {
+		metro  string
+		quotas *platform.Quotas
+		err    error
+	}, len(metros))
+
+	var eg errgroup.Group
+	for i, m := range metros {
+		eg.Go(func() error {
+			err := group.DoMetro(ctx, g, m.Name, func(ctx context.Context, mc multimetro.MetroClient) error {
```
jedevc, line 513: Hm, this is fine for now. But this would be simplified by updating the `group` code to support a `DoMetros` and `CollectMetros`.

If you could add those, upstream them, then we can fix this up (either before or after merge, no strong preference there).

internal/cmd/metro_quotas.go
craciunoiuc: makes sense, whoops

jedevc (commented):
ffs, i reviewed a week ago and forgot to post comments.

internal/cmd/metro_quotas.go
craciunoiuc: hmm, sure, I just need to make the requirement 1-2 smaller in height

internal/cmd/metro_quotas.go
craciunoiuc: After we merge  then :flushed:

craciunoiuc: This is actually used in a single place in the whole code. So let's wait till there's more usage

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,546 @@
...
+	err           error
+}
+
+type (
+	quotasLoadedMsg = multiMetroResult
+	quotasTickMsg   struct{}
+	quotasStatusMsg struct{}
+)
+
+const quotaBarWidth = 36
```
jedevc: non-blocking, let's get this in, but I commented this before:

> Ideally, it should also grow and shrink as we resize :)))

internal/cmd/metro_quotas.go
```diff
@@ -0,0 +1,546 @@
...
+
+func newQuotasModel(ctx context.Context, metros []string, watchInterval *time.Duration) quotasModel {
+	var tabs []string
+	if len(metros) == 1 {
+		tabs = []string{metros[0]}
+	} else {
+		// Will be populated after first fetch.
+		tabs = nil
```
jedevc: Why does this need special casing?

jedevc (commented):
Why does this contain https://github.com/unikraft/cli/pull/376?

Otherwise good to go, some small non-blocking comments (which you can ignore, or do later).

internal/cmd/metro_quotas.go
craciunoiuc: Ah, missed that. Agree

Though I wouldn't make it smaller than 10 blocks :eyes:

internal/cmd/metro_quotas.go
craciunoiuc: can't remember, will have to check

internal/cmd/metro_quotas.go
craciunoiuc: you added this, so we can do it before now

jedevc (approved):
```
```

PR #341, opened 2026-05-25
feat(volumes): Add volume 'access_mode' flag

craciunoiuc:
This allows for mounting volumes in multiple instances.
The access mode can't be edited after creation, as such
there is no support for patching.

Closes: TOOL-826


internal/cmd/testdata/TestOutput/volumes
```diff
@@ -18,6 +18,7 @@ size:         50MiB
 filesystem:   ext4
 quota-policy: hard
 persistent:   true
+access-mode:
```
jedevc: Hm, why empty? Shouldn't the struct be updated so the gen of this is correct?

internal/cmd/testdata/TestOutput/volumes
jedevc, line 1: nit: just noticed that this golden file (and maybe others) don't have the right `linguist-generated` gitattributes.

internal/cmd/volumes.go
```diff
@@ -540,7 +548,8 @@ func (Volume) Examples() map[cmd.CmdType][]kingkong.Example {
 					`unikraft volume create \
 	  --name demo-volume \
 	  --size 10 \
-	  --metro fra`,
+	  --metro fra \
+	  --access-mode rwo`,
```
jedevc: No need to add the access-mode to this specific example, IMO, it should just be a very basic example.

jedevc (commented):
Looks mostly good, left some comments. Once fixed, we can merge after the feature freeze is over.

internal/cmd/testdata/TestOutput/volumes
craciunoiuc: what the hyok is that? :eyes:

internal/cmd/testdata/TestOutput/volumes
craciunoiuc: Good question, need to recheck, I think I'm setting it to empty in the request to make the platform decide

craciunoiuc: Image

weird that this is a thing, I'm adding a fix now

internal/cmd/testdata/TestOutput/volumes
craciunoiuc: added

jedevc (approved):

PR #342, opened 2026-05-26
chore(integration): Import old integration tests

craciunoiuc:


cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,213 @@ func TestInstances(t *testing.T) {
 
 		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
 	})
+
+	t.Run("create-stopped", func(t *testing.T) {
```
jedevc: nit: test ordering. put this one up by `create` at the top. Make sure it shares some of the same assertions as well.

You could also make them all subtests of `create`, like `TestInstances/create/stopped`.

cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,213 @@ func TestInstances(t *testing.T) {
...
+			"--set", "autostart=false",
+			"--set", "resources.memory=128",
+			"--set", "resources.vcpus=1",
+		})
+		assert.Regexp(t, `state:\s+stopped`, out)
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
+	})
+
+	t.Run("create-name-too-long", func(t *testing.T) {
```
jedevc: Some of these tests test platform behavior instead of the CLI. I don't *think* we should test this here... WDYT? I'd rather see these as integration tests.

cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,213 @@ func TestInstances(t *testing.T) {
...
+		r := runner(t, true)
+		instName := uniq()
+
+		r.Run(t, []string{
+			"unikraft", "instance", "create",
+			"--output", "quiet",
+			"--set", "name=test-" + instName,
+			"--set", "metro=" + r.Config.MetroName,
+			"--set", "image=nginx:latest",
+			"--set", "runtime.env=HÉLLO=wörld",
```
jedevc: Good idea for a test!

cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,213 @@ func TestInstances(t *testing.T) {
...
+		})
+
+		r.Run(t, []string{"unikraft", "instance", "wait", "--until", "state==running", "--timeout", "30s", "test-" + instName})
+
+		out := r.Run(t, []string{"unikraft", "instance", "logs", "test-" + instName})
+		assert.NotEmpty(t, out)
+
+		out = r.Run(t, []string{"unikraft", "instance", "logs", "--tail", "1", "test-" + instName})
+		lines := strings.Split(strings.TrimSpace(out), "\n")
+		assert.LessOrEqual(t, len(lines), 1)
```
jedevc: Let's test for some actual content in here (e.g. busybox + echo).

cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,213 @@ func TestInstances(t *testing.T) {
...
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
+	})
+
+	t.Run("restart", func(t *testing.T) {
```
jedevc: nit: place after `start-stop`.

cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,213 @@ func TestInstances(t *testing.T) {
...
+			"--set", "resources.vcpus=1",
+		})
+
+		out := r.Run(t, []string{"unikraft", "instance", "inspect", "test-" + instName})
+		assert.Regexp(t, `HÉLLO:\s+wörld`, out)
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
+	})
+
+	t.Run("logs", func(t *testing.T) {
```
jedevc: nit: make `start-follow` and `stop-follow` as subtests. For example:
- `TestInstances/logs/tail`
- `TestInstances/logs/follow`
- `TestInstances/logs/follow-restart`

cmd/unikraft/integration/instance_test.go
craciunoiuc: True, though, if we have them it's good anyway, we just need to make sure to not write more :flushed: 

If you want I can tag someone to add them as integration tests for the platform first?

cmd/unikraft/integration/instance_test.go
craciunoiuc: This is useless, we already have a log test no?

I'm deleting

cmd/unikraft/integration/instance_test.go
craciunoiuc: I removed the `logs` function so now we don't need to do this :flushed:

PR #345, opened 2026-05-27
fix: Use default http client for secure image operations

jedevc:
We were overriding to use our custom http client for insecure image operations - but defaulting to containerd's http client for secure operations.

This was inconsistent, and actually resulted in us using a 30s timeout, instead of waiting indefinitely (which is technically what we want to be doing here, at least until we introduce an explicit `--timeout`).


craciunoiuc (approved):
All good here. Thanks!

```
```

PR #346, opened 2026-05-27
feat(cmd)!: Add global timeout flag

craciunoiuc:
This allows cancelling any ongoing command by cancelling the running context for everything.

This also removes existing `--timeout` flags in login, wait, and volume import.
These flags had duplicate effects and are now centralised.

Closes: TOOL-818

### DIFF
--- a/cmd/unikraft/integration/instance_test.go
+++ b/cmd/unikraft/integration/instance_test.go
@@ -71,7 +71,7 @@ func TestInstances(t *testing.T) {
 			"--set", "resources.vcpus=1",
 		})
 
-		out := r.Run(t, []string{"unikraft", "instance", "wait", "--until", "state==stopped", "--timeout", "10s", "test-" + instName})
+		out := r.Run(t, []string{"unikraft", "--timeout", "10s", "instance", "wait", "--until", "state==stopped", "test-" + instName})
 		assert.Regexp(t, `state:\s+stopped`, out)
 		assert.Regexp(t, `stop:`, out)
 		assert.Regexp(t, `reason:.*(page fault|out of memory)`, out)
@@ -102,7 +102,7 @@ func TestInstances(t *testing.T) {
 		})
 		fqdn := strings.TrimSpace(out)
 
-		r.Run(t, []string{"unikraft", "instance", "wait", "--until", "state==running", "--timeout", "10s", "test-" + instName})
+		r.Run(t, []string{"unikraft", "--timeout", "10s", "instance", "wait", "--until", "state==running", "test-" + instName})
 
 		body := integ.HTTPGet(t, "https://"+fqdn)
 		assert.Contains(t, body, "Welcome to nginx!")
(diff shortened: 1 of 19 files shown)


craciunoiuc:
Need to fix these

craciunoiuc:
test machine was broken, will need to recheck tests again after

internal/cmd/volumes.go
```diff
@@ -705,7 +706,12 @@ func (c *VolumeImportCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *
 	var instUUID, instFQDN string
 	if err := group.DoMetro(ctx, g, vol.Metro.Name, func(ctx context.Context, mc multimetro.MetroClient) error {
 		var merr error
-		instUUID, instFQDN, merr = volimport.Start(ctx, mc, c.Image, vol.UUID, authStr, c.Timeout, c.Port)
+		volimportTimeout := uint64(10)
+		if deadline, ok := ctx.Deadline(); ok {
+			t := max(10*time.Second, time.Until(deadline)+2*time.Second)
```
jedevc: Comments for this pls.

```suggestion
			t := max(
				// don't run past deadline (+2s for tolerance)
				time.Until(deadline)+2*time.Second,
				// set reasonable max timeout
				10*time.Second,
			)
```

internal/cmd/volumes.go
```diff
@@ -705,7 +706,12 @@ func (c *VolumeImportCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *
 	var instUUID, instFQDN string
 	if err := group.DoMetro(ctx, g, vol.Metro.Name, func(ctx context.Context, mc multimetro.MetroClient) error {
 		var merr error
-		instUUID, instFQDN, merr = volimport.Start(ctx, mc, c.Image, vol.UUID, authStr, c.Timeout, c.Port)
+		volimportTimeout := uint64(10)
```
jedevc, line 709: bug: If there is no deadline, then `volImportTimeout` stays zero.

internal/resource/cmd/cmd_test.go
```diff
@@ -1019,11 +1018,12 @@ func TestWait(t *testing.T) {
 	t.Run("timeout", func(t *testing.T) {
 		env := setupTestEnv()
 		ctx := resourcet.WithTestEnv(context.Background(), env)
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
```
jedevc: clarity nit:

```suggestion
		ctx, cancel := context.WithTimeout(ctx, 1 * time.Second)
```

cmd/unikraft/integration/instance_test.go
```diff
@@ -71,7 +71,7 @@ func TestInstances(t *testing.T) {
 			"--set", "resources.vcpus=1",
 		})
 
-		out := r.Run(t, []string{"unikraft", "instance", "wait", "--until", "state==stopped", "--timeout", "10s", "test-" + instName})
+		out := r.Run(t, []string{"unikraft", "--timeout", "10s", "instance", "wait", "--until", "state==stopped", "test-" + instName})
```
jedevc, line 74: :eyes: why do all the tests change? Can `--timeout` not be set at the end? It should be a global flag, and should be able to appear everywhere.

Test change is fine, but just want to check that this isn't hiding unintended behavior.

internal/cmd/root_test.go
jedevc, line 1: Hmmmm. I'm not convinced these tests are actually testing anything super useful. We shouldn't be testing the context that's returned by `NewRootCmd`.

Instead of doing this, I'd suggest adding some behavioral tests that timeouts are followed, for commands like `watch` (e.g. that no `--timeout` doesn't exit by itself, and that `--timeout` is obeyed).

cmd/unikraft/integration/instance_test.go
craciunoiuc: Ah, this is not a problem, it can appear everywhere

I thought it would be better to move it in front to make it clear that it's a global flag

I can revert if you want, but from the message I understand it's fine to leave as is :D

internal/cmd/root_test.go
craciunoiuc: sure thing, should be easy to add

(I actually used `instance ls` to test manually)

internal/cmd/volumes.go
craciunoiuc: Wait, what do you mean, the default value is 10 seconds

That is outside of the deadline check so all good?

jedevc (approved):
```
```

PR #348, opened 2026-05-29
feat(cmd): Add 'sched-priority' instance field

craciunoiuc:
Also bump the sdk to the latest version to make the field available to the cli for using.
I bumped to `prod-stable` as the `prod-staging` branch messed up go versions after it was removed.
We'll go back to that one afterwards.

Closes: TOOL-926


internal/cmd/testdata/TestOutput/instances
```diff
@@ -92,6 +92,7 @@ restart:
   policy:        always
   start-count:   3
   restart-count: 1
+sched-priority:
```
jedevc: Make sure to populate the test struct, so this actually has a value!

jedevc (approved):
One nit, please fix and regen golden files, then merge at will!

PR #350, opened 2026-06-02
refactor(integration): Build shared images only once

craciunoiuc:
Like this we build the instance only once instead of multiple times cutting down significantly on builds.

Three commits:
1. Add the shared mechanism and use it for busybox images
2. Use the shared image also for images that change only the command
3. Use another shared image for some python tests

### DIFF
--- a/internal/integration/image.go
+++ b/internal/integration/image.go
@@ -1,0 +1,177 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package integration
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	imagespec "unikraft.com/x/image-spec"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/images"
+)
+
+// Busybox is a busybox rootfs with no application. Tests give their own
+// command with --args or runtime.args, which replaces the cmd of the image.
+var Busybox = &SharedImage{
+	Image: Image{
+		Name: "busybox-e2e",
+		Files: map[string]string{
+			"Dockerfile": "FROM busybox:latest",
+			"Kraftfile": `spec: v0.7
+name: busybox-e2e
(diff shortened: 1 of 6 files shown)


craciunoiuc:
prod fail is unrelated to this change (does not touch the same tests)

internal/integration/image.go
```diff
@@ -0,0 +1,170 @@
...
+type SharedImage struct {
+	// Name is the image name, without the organization and the tag.
+	Name string
+	// Each key is a file name in the context root.
+	Files map[string]string
+
+	once sync.Once
+	ref  string
+	err  error
+}
```
jedevc: idea: what if we had this as a `type Image` and `type SharedImage` (which then extends the `Image`?)

Using the bare `Image` would be super useful for places where we aren't using a shared image.

internal/integration/image.go
```diff
@@ -0,0 +1,170 @@
...
+	"sync"
+	"testing"
+
+	"github.com/containerd/continuity/fs/fstest"
+	"github.com/stretchr/testify/require"
+)
+
+// SharedBusybox is a busybox rootfs with no application. Tests give their own
+// command with --args or runtime.args, which replaces the cmd of the image.
+var SharedBusybox = &SharedImage{
```
jedevc: Just call this `Busybox`.

```suggestion
var Busybox = &SharedImage{
```

internal/integration/image.go
```diff
@@ -0,0 +1,170 @@
...
+	refs := slices.Clone(sharedImageRefs)
+	sharedImageRefs = nil
+	sharedImageMu.Unlock()
+
+	if cleanup == nil || len(refs) == 0 {
+		return
+	}
+	defer os.RemoveAll(cleanup.dir)
+
+	c := exec.Command(cleanup.unikraftPath, append([]string{"image", "delete"}, refs...)...)
```
jedevc: Hm, can we instead just call the actual functions this relies on? We don't need the full shell out here.

internal/integration/image.go
craciunoiuc: Hmm, and I assume the base image would have only the Name and Files

I guess it could help yeah

internal/integration/image.go
craciunoiuc: Hmm yeah, it's just one place where this happens, I think it should work, checking

craciunoiuc: Ok, this helped remove a lot of clutter, should be better now

internal/integration/image.go
craciunoiuc: Done

jedevc (approved):
Neat :tada:

```
```

PR #351, opened 2026-06-03
feat(workflows): Add go-sdk auto bump workflow

craciunoiuc:
If we add this, we will get PRs for sdk bumps instantly, instead of waiting 1 week for dependabot to kick in.
Also, dependabot doesn't work with branch bumps I believe.

We need to merge this before we merge the one below to be clean:

Depends on: https://github.com/unikraft-cloud/go-sdk/pull/196


.github/workflows/sync-sdk.yaml
```diff
@@ -0,0 +1,116 @@
+name: sync sdk
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
craciunoiuc: It's fine, we can use it like this for manual bumps and auto-bumps also

.github/workflows/sync-sdk.yaml
```diff
@@ -0,0 +1,117 @@
...
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Bump SDK version
+        id: bump-sdk
+        shell: bash
+        run: |
+          echo "Bumping Go SDK to latest from ${{ inputs.channel }}..."
+          go get unikraft.com/cloud/sdk@${{ inputs.channel }}
```
jedevc, line 41: I *think* we probably want to make sure that this is set using a commit hash. I know that `go get` on a branch can sometimes annoyingly cache.

.github/workflows/sync-sdk.yaml
```diff
@@ -0,0 +1,117 @@
...
+          if git diff --quiet -- go.mod go.sum; then
+            echo "No changes detected. Skipping PR creation."
+            echo "changes=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "changes=true" >> "$GITHUB_OUTPUT"
+            echo "Changes detected in go.mod/go.sum:"
+            git diff --stat go.mod go.sum || true
+          fi
+
+      - name: Close existing related unmerged bump PRs
```
jedevc, line 57: No need to do this. `peter-evans/create-pull-request` will update an existing branch if it's already there :tada:

.github/workflows/sync-sdk.yaml
```diff
@@ -0,0 +1,117 @@
+name: sync sdk
```
nderjung, line 1:
```suggestion
name: sync-sdk
```

PR #354, opened 2026-06-05
fix: Ensure certificate linking to services works

jedevc:
We were missing the connection.

Fix + add a test.

### DIFF
--- a/internal/integration/http.go
+++ b/internal/integration/http.go
@@ -7,9 +7,12 @@ package integration
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/url"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


craciunoiuc:
seems ok, I'll approve after copilot + the test passes

internal/integration/http.go
```diff
@@ -50,3 +53,47 @@ func HTTPGet(t *testing.T, url string) string {
...
+// HTTPGetTLSCerts dials the HTTPS URL and returns the peer's certificate chain.
+// TLS verification is intentionally skipped so self-signed certificates work.
+// It retries up to 10 times with a 2-second sleep between attempts.
+//
+// dialAddr overrides the TCP address to connect to (host:port). When empty the
+// address is derived from the URL itself. This is useful when DNS for the URL
+// hostname does not resolve but the load balancer is reachable at a known
+// address (e.g. resolved from an internal FQDN) and SNI routing is used to
+// select the right certificate.
+func HTTPGetTLSCerts(t *testing.T, rawURL string, dialAddr string) []*x509.Certificate {
```
craciunoiuc, line 66: Why can't we just check if a `:<number>` is set at the end?

basically if `:` is used and is not in the preamble, and it's followed by a number, then do the thing you want to do

If not, just follow the usual flow

jedevc: So the address here is used to override.

E.g. the server is serving on "foo.example.com" (an external service). But we want to still make the connection against "foo.unikraft.app". We need this to simulate an external service, so we can attach a certificate.

craciunoiuc: hmm, ok

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #355, opened 2026-06-05
feat: Add autokill

jedevc:
Adds `--autokill` everywhere the platform accepts it: instances (`time` + `num-requests`), and service groups, instance templates and checkpoints (`time` only — those endpoints have no request-count rigger). Create and edit both, on all four.

Templates read back from `instance.template_autokill` and checkpoints from `instance.checkpoint_autokill` — the platform stores one timer but emits it under different names depending on the kind of instance.

Checkpoint create needed [https://github.com/unikraft-cloud/proto/pull/343](<https://github.com/unikraft-cloud/proto/pull/343>) first, since the spec never described `autokill` on that endpoint; the SDK bump here picks it up.

The one remaining autokill surface is image pin, which the CLI has no command for at all (`PinImages`/`UnpinImages` are unused) — that is its own feature, not an autokill gap. The same proto change also added `tags` to checkpoint create, which is still unwired here; happy to fold that in if you would rather not have a second PR.

### DIFF
--- a/cmd/unikraft/integration/instance_test.go
+++ b/cmd/unikraft/integration/instance_test.go
@@ -1842,4 +1842,38 @@ cmd: ["cat", "/marker.txt"]
 		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName, "test-branch-" + branchName})
 		r.Run(t, []string{"unikraft", "instance", "template", "delete", templateName})
 	})
+
+	t.Run("autokill", func(t *testing.T) {
+		r := runner(t, true, []string{staging, stable})
+		instName := uniq()
+
+		r.Run(t, []string{
+			"unikraft", "instance", "create",
+			"--output", "quiet",
+			"--set", "name=test-" + instName,
+			"--set", "metro=" + r.Config.MetroName,
+			"--set", "image=nginx:latest",
+			"--set", "autostart=false",
+			"--set", "resources.memory=128",
+			"--set", "resources.vcpus=1",
+			"--autokill", "time=5s",
+		})
+
+		out := r.Run(t, []string{"unikraft", "instance", "inspect", "test-" + instName})
+		assert.Regexp(t, `autokill:`, out)
+		assert.Regexp(t, `time:\s+5s`, out)
+
+		r.Run(t, []string{
+			"unikraft", "instance", "edit", "test-" + instName,
+			"--output", "quiet",
+			"--autokill", `time=10s,num-requests=100`,
+		})
+
+		out = r.Run(t, []string{"unikraft", "instance", "inspect", "test-" + instName})
+		assert.Regexp(t, `autokill:`, out)
+		assert.Regexp(t, `time:\s+10s`, out)
+		assert.Regexp(t, `num-requests:\s+100`, out)
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
+	})
 }
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -137,6 +137,7 @@ type Instance struct {
 	}
 
 	ScaleToZero InstanceScaleToZero `field:",embed" mirror:"instance.scale_to_zero" create:"set" edit:"set" flag:"scale-to-zero" help:"Scale-to-zero options.\n  policy: on | idle | off\n  cooldown-time: cooldown in ms before scaling to zero\n  notify-time: notification time in ms before scaling to zero\n  stateful: true | false" placeholder:"<key>=<value>" example:"on,policy=on\\,cooldown-time=300,policy=on\\,stateful=true\\,cooldown-time=500\\,notify-time=100"`
+	Autokill    InstanceAutokill    `field:",embed" mirror:"instance.autokill" create:"set" edit:"set" flag:"autokill" help:"Autokill options.\n  time: time after the instance stops before it is deleted\n  num-requests: max requests before the instance is deleted" placeholder:"<key>=<value>" example:"time=5s,num-requests=100,time=5s\\,num-requests=100"`
 
 	Timing struct {
 		Uptime   types.DurationMS `mirror:"instance.uptime_ms"`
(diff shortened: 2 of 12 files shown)


cmd/unikraft/integration/instance_test.go
```diff
@@ -528,4 +528,38 @@ func TestInstances(t *testing.T) {
...
+			"unikraft", "instance", "edit", "test-" + instName,
+			"--output", "quiet",
+			"--autokill", `time=10s,num-requests=100`,
+		})
+
+		out = r.Run(t, []string{"unikraft", "instance", "inspect", "test-" + instName})
+		assert.Regexp(t, `autokill:`, out)
+		assert.Regexp(t, `time:\s+10s`, out)
+		assert.Regexp(t, `num-requests:\s+100`, out)
+
```
craciunoiuc, line 1247: Can we make it auto-kill on 1 request and then do an httpget on it and check again if it dissapeared?

internal/cmd/instances.go
```diff
@@ -88,6 +88,8 @@ type InstanceCreateCmd struct {
 
 	ScaleToZero InstanceScaleToZero `group:"flag-create" shortcut:"scale-to-zero" help:"Scale-to-zero options.\n  policy: on | idle | off\n  cooldown-time: cooldown in ms before scaling to zero\n  notify-time: notification time in ms before scaling to zero\n  stateful: true | false" placeholder:"<key>=<value>" example:"on,policy=on\\,cooldown-time=300,policy=on\\,stateful=true\\,cooldown-time=500\\,notify-time=100"`
 
+	Autokill InstanceAutokill `group:"flag-create" shortcut:"autokill" help:"Autokill options.\n  time: time in ms after the instance stops before it is deleted\n  num-requests: max requests before the instance is deleted" placeholder:"<key>=<value>" example:"time=5s,num-requests=100,time=5s\\,num-requests=100"`
```
craciunoiuc: Something is wrong with `time:`

You say time in `ms`, but the example has `s`

I would just remove the mention of `in ms`

internal/cmd/instances.go
```diff
@@ -127,6 +129,8 @@ type InstanceEditCmd struct {
 	Rom []InstanceRom `group:"flag-edit" shortcut:"roms" sep:"none" help:"Attach ROM." placeholder:"image=<ref>,at=<path>" example:"image=myuser/my-rom:latest\\,at=/rom0\\,name=my-rom,dir=./mydata\\,at=/rom"`
 
 	ScaleToZero InstanceScaleToZero `group:"flag-edit" shortcut:"scale-to-zero" help:"Scale-to-zero options.\n  policy: on | idle | off\n  cooldown-time: cooldown in ms before scaling to zero\n  notify-time: notification time in ms before scaling to zero\n  stateful: true | false" placeholder:"<key>=<value>" example:"on,policy=on\\,cooldown-time=300,policy=on\\,stateful=true\\,cooldown-time=500\\,notify-time=100"`
+
+	Autokill InstanceAutokill `group:"flag-edit" shortcut:"autokill" help:"Autokill options.\n  time: time in ms after the instance stops before it is deleted\n  num-requests: max requests before the instance is deleted" placeholder:"<key>=<value>" example:"time=5s,num-requests=100,time=5s\\,num-requests=100"`
```
craciunoiuc: same here

craciunoiuc (changes requested):
some comments

cmd/unikraft/integration/instance_test.go
jedevc: Hm. This feels like something for the QA team.

We test that the property can be created + edited + viewed - so the CLI is working correctly. We should avoid testing that the platform behavior is correct (as much as possible in these tests).

More complex tests make sense in some scenarios... but I think this one is a bit over-the-line. If you think otherwise, I can hack something together though.

craciunoiuc: I'm fine either way, but I don't think that the functionality tested is that complex 🙏 

I'll let you decide

internal/cmd/instances.go
```diff
@@ -127,6 +129,8 @@ type InstanceEditCmd struct {
 	Rom []InstanceRom `group:"flag-edit" shortcut:"roms" sep:"none" help:"Attach ROM." placeholder:"image=<ref>,at=<path>" example:"image=myuser/my-rom:latest\\,at=/rom0\\,name=my-rom,dir=./mydata\\,at=/rom"`
 
 	ScaleToZero InstanceScaleToZero `group:"flag-edit" shortcut:"scale-to-zero" help:"Scale-to-zero options.\n  policy: on | idle | off\n  cooldown-time: cooldown in ms before scaling to zero\n  notify-time: notification time in ms before scaling to zero\n  stateful: true | false" placeholder:"<key>=<value>" example:"on,policy=on\\,cooldown-time=300,policy=on\\,stateful=true\\,cooldown-time=500\\,notify-time=100"`
+
+	Autokill InstanceAutokill `group:"flag-edit" shortcut:"autokill" help:"Autokill options.\n  time: time after the instance stops before it is deleted\n  num-requests: max requests before the instance is deleted" placeholder:"<key>=<value>" example:"time=5s,num-requests=100,time=5s\\,num-requests=100"`
```
nurof3n:
```suggestion
	Autokill InstanceAutokill `group:"flag-edit" shortcut:"autokill" help:"Autokill options.\n  time: time after the instance stops before it is deleted\n  num-requests: max requests before the instance is deleted" placeholder:"<key>=<value>" example:"time=5s,num-requests=100\\,num-requests=100"`
```

jedevc: This one would render as two examples, `time=5s` and `num-requests=100,num-requests=100` — the escaped comma joins the pair, so dropping `time=5s\,` from the third example leaves `num-requests` set twice.

Kept the original, which renders as:

```
[examples: time=5s, num-requests=100, time=5s,num-requests=100]
```

cmd/unikraft/integration/instance_test.go
jedevc: Leaving this as a create/edit/view test, per the thread above — happy to revisit if you feel strongly now that `applyCounterContext` and `integ.HTTPGet` exist and make it cheaper than it was in June.

The `--autokill` help text no longer says "in ms", which was your other comment.

internal/cmd/instances.go
nurof3n: oh lol okay

jedevc:
Checkpoint create is unblocked upstream by https://github.com/unikraft-cloud/proto/pull/343 — once that lands and the SDK regenerates, `--autokill` (and `--tag`) can be wired into `instance checkpoint create` here too.

jedevc:
@craciunoiuc @nurof3n ready for another review, I added autokill to a lot more places!

internal/cmd/instances.go
```diff
@@ -554,6 +559,89 @@ func (s InstanceScaleToZero) MarshalJSON() ([]byte, error) {
...
+func (a InstanceAutokill) MarshalJSON() ([]byte, error) {
+	type autokillJSON InstanceAutokill
+	return json.Marshal(autokillJSON(a))
+}
+
+// Autokill is the time-only form of autokill used by service groups and
+// instance templates, whose endpoints have no request-count trigger.
+type Autokill struct {
+	TimeMs types.DurationMS `name:"time" json:"time,omitempty" mirror:"time_ms" field:"time,long"`
+}
```
craciunoiuc, line 621: Why keep 2 structures instead of a single one? The fields have `omitempty` so if they are not set they will just not be there

Do you think maybe they will diverge in the future with more options?

craciunoiuc (commented):
Just a comment from my side, otherwise looks good

(unless you want to add autokill tests for services and the rest also apart from instances hehe)

internal/cmd/instances.go
jedevc: They're not the same shape — only the instance endpoint has `num_requests`:

- [`CreateInstanceRequestAutokill`](https://github.com/unikraft-cloud/go-sdk/blob/b82b2d8e3cfd/platform/model_create_instance_request_autokill.gen.go#L19) — `time_ms` + `num_requests`
- [`CreateServiceGroupRequestAutokill`](https://github.com/unikraft-cloud/go-sdk/blob/b82b2d8e3cfd/platform/model_create_service_group_request_autokill.gen.go#L18) — `time_ms` only
- [`ItemAutokill`](https://github.com/unikraft-cloud/go-sdk/blob/b82b2d8e3cfd/platform/model_item_autokill.gen.go#L18) (templates) — `time_ms` only
- [`ItemCheckpointAutokill`](https://github.com/unikraft-cloud/go-sdk/blob/b82b2d8e3cfd/platform/model_item_checkpoint_autokill.gen.go#L18) (checkpoints) — `time_ms` only

`Autokill` is already shared by all three of those, so it's two CLI types covering four SDK ones.

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #363, opened 2026-06-09
feat: Fallback to dynamic controlplane fetch if no metros

jedevc:
If no metros are available, and we have a controlplane, we can dynamically grab em.


nderjung (approved):
Thanks!

PR #366, opened 2026-06-10
ci: Set `CGO_ENABLED=0` for releases

jedevc:
So that we can run regardless of libc version.

We can also disable the extldflags, since those don't have an effect for CGO_ENABLED=0.

Fixes https://linear.app/unikraft/issue/TOOL-1028/unikraft-cli-doesnt-work-on-debian-11.


craciunoiuc (approved):
All good here. Thanks!

```
```

PR #375, opened 2026-06-17
fix(cmd): Check instances exist before logging

craciunoiuc:
This avoids returning empty logs as success on instances that actually do not exist.


jedevc (approved):
Ah, this is actually *slightly* related to https://linear.app/unikraft/issue/TOOL-1036/unikraft-instance-logs-tail-fails-if-one-of-the-metros-is-down. Part of the problem was the inconsistency.

This fixes at least that part, but I think we still need some other fixes.

```
```

PR #376, opened 2026-06-18
feat(integration): Add simple run test for stable/prod

craciunoiuc:


cmd/unikraft/integration/build_test.go
```diff
@@ -18,6 +19,57 @@ import (
...
+cmd: ["sh", "/entrypoint.sh"]
+`), 0o644),
+			).Apply(dir))
+
+			r.Run(t, []string{"unikraft", "build", ".", "--output", image}, integ.WithWorkDir(dir))
+			if tc.name == "busybox-sanity-prod" {
+				// Wait for image propagation
+				// NOTE(craciunoiuc): To be removed when nodes are updated
+				time.Sleep(5 * time.Second)
+			}
```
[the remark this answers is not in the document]
craciunoiuc: Best we can do for those old nodes

cmd/unikraft/integration/build_test.go
```diff
@@ -18,6 +19,57 @@ import (
 )
 
 func TestBuild(t *testing.T) {
+	for _, tc := range []struct {
```
jedevc: This isn't necessarily a build test. I would call this a "sanity test". Maybe put it in `TestSanity` in `sanity_test.go`.

I would run it over stable, prod and staging.

internal/integration/config.go
```diff
@@ -15,7 +16,15 @@ import (
...
 )
 
-func LoadConfig(t *testing.T) (*Config, error) {
+func LoadConfig(t *testing.T, names ...string) (*Config, error) {
+	if len(names) > 0 {
+		configPath, err := config.ConfigFilePath()
+		if err != nil {
+			return nil, err
+		}
+		return loadConfig(filepath.Join(filepath.Dir(configPath), names[0]))
```
jedevc, line 25: We should probably do a cache map - for each `name`.

cmd/unikraft/integration/build_test.go
craciunoiuc: hmm, okay

internal/integration/config.go
craciunoiuc: In a way, docker and the registries already cache most of the stuff for us I think

Would still be good to cache, but it won't make that big of a difference no?

jedevc: Yah, it's just to stop reading the config and reparsing it once on every test.

It's just for consistency here with the existing code.

.github/workflows/test.yaml
```diff
@@ -110,6 +110,10 @@ jobs:
         run: |
           mkdir -p ~/.config/unikraft
           echo "${UNIKRAFT_CONFIG_CONTENT}" > ~/.config/unikraft/config.yaml
+          echo "${UNIKRAFT_CONFIG_CONTENT_STABLE}" > ~/.config/unikraft/config-stable.yaml
+          echo "${UNIKRAFT_CONFIG_CONTENT_PROD}" > ~/.config/unikraft/config-prod.yaml
```
jedevc, line 114: I like this :tada:

As we discussed though, I think we should have the tests run as a matrix. Then each test should be annotated as to what environment it should run it.

e.g. most tests will run on both staging+stable, and only the most basic ones should run on prod.

PR #382, opened 2026-06-24
fix(cmd): Do not add both ports/domains and a service

craciunoiuc:


jesi-rgb (approved):
haha 

Reviewed-by: Jesús Rascón <jesus@unikraft.io>
Approved-by: Jesús Rascón <jesus@unikraft.io>

PR #388, opened 2026-06-26
fix: Allow GET to show more than expected results

nurof3n:
Closes TOOL-1048

### DIFF
--- a/internal/cmd/certificates.go
+++ b/internal/cmd/certificates.go
@@ -149,17 +149,20 @@ func (Certificate) Get(ctx context.Context, keys []string) ([]resource.Resource,
 		if resp == nil || resp.Data == nil {
 			return nil, nil, nil
 		}
-		for i, certificate := range resp.Data.Certificates {
-			result, err := Certificate{}.load(&refs[i], certificate, &c.Metro)
+		for _, certificate := range resp.Data.Certificates {
+			// Deliberately do not filter by status: pending certificates should
+			// be returned by GET as well.
+			matchedRef := matchRef(refs, certificate.Name, certificate.Uuid)
+			result, err := Certificate{}.load(matchedRef, certificate, &c.Metro)
 			if err != nil {
 				errs = append(errs, err)
 				continue
 			}
-			found = append(found, group.Ref{
-				Metro: c.Metro.Name,
-				Name:  result.Name,
-				UUID:  result.UUID,
-			})
+			if matchedRef != nil {
+				found = append(found, *matchedRef)
+			} else {
+				found = append(found, group.Ref{Metro: c.Metro.Name, Name: result.Name, UUID: result.UUID})
+			}
 			results = append(results, result)
 		}
 		return results, found, errors.Join(errs...)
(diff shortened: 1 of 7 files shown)


internal/cmd/instances.go
```diff
@@ -630,20 +630,39 @@ func (Instance) Get(ctx context.Context, keys []string) ([]resource.Resource, er
...
+					return ref.UUID == instance.Uuid
+				}
+				if ref.Name != "" && instance.Name != "" {
+					return ref.Name == instance.Name
+				}
+				return false
+			}); idx >= 0 {
+				copyRef := refs[idx]
+				matchedRef = &copyRef
+			}
```
jedevc: Okay, can you refactor this to a function - something like:

```go
// findResponse derives the response id
func findResponse(idx, requestRefs, responseRefs) idx {
    // naive commented out case
	// return idx

    // complex case...
    // ...
}
```

(or maybe something funky with generics if you can make the caller side look nicer)

I'd like to share it between all resources. (and also templates, etc, which seem to be missing in this PR).

I'm actually a bit scared of this approach - I think it's okaaayyyyy, but if all the tricky bits are in one nice function together, at least it's easy to update later.

jedevc (changes requested):

internal/cmd/instances.go
nurof3n: pff yeah I forgot to add it to those cuz I only tested with these 3 resources :)) will do

craciunoiuc:
looks good to me, have a free copilot on the house

internal/cmd/util.go
```diff
@@ -54,6 +57,30 @@ loop:
...
+	ref := refs[idx]
+	return &ref
+}
+
+func appendFoundRef(found []group.Ref, matchedRef *group.Ref, metro, name, uuid string) []group.Ref {
+	if matchedRef != nil {
+		return append(found, *matchedRef)
+	}
+	return append(found, group.Ref{Metro: metro, Name: name, UUID: uuid})
+}
```
jedevc: What is this helper 🤔 I would just inline the `nil` check to call sites.

internal/cmd/util.go
```diff
@@ -54,6 +57,30 @@ loop:
 	return found, nil
 }
 
+func matchedRefByNameOrUUID(refs group.Refs, name, uuid string) *group.Ref {
```
jedevc: This could use a comment explaining what this does, and why the naive approach of just indexing with the same request key doesn't work.

internal/cmd/certificates.go
```diff
@@ -149,17 +149,18 @@ func (Certificate) Get(ctx context.Context, keys []string) ([]resource.Resource,
 		if resp == nil || resp.Data == nil {
 			return nil, nil, nil
 		}
-		for i, certificate := range resp.Data.Certificates {
-			result, err := Certificate{}.load(&refs[i], certificate, &c.Metro)
+		for _, certificate := range resp.Data.Certificates {
+			if certificate.Status == nil || *certificate.Status != platform.ResponseStatusSuccess {
+				continue
+			}
```
jedevc: This is deliberate. https://github.com/unikraft-cloud/cli/pull/235

internal/cmd/util.go
jedevc: Could also maybe be called just `matchRef`?

jedevc (approved):
Neat :tada:

```
```

PR #389, opened 2026-06-27
feat: Implement nested JSON in api subcommand

### LINKED ISSUE #369: Use nested JSON syntax in `unikraft api`
`unikraft api` is quite nice. However, in the `-d` flag, it only takes full JSON bodies.

Ideally we would *also* support something like the nested JSON syntax of https://httpie.io/docs/cli/nested-json, and make sure it's documented on our generated markdown docs (like [here](https://unikraft.com/docs/cli/unikraft/login)). Also see the docs here: https://unikraft.com/docs/api/platform/v1/instances

Examples:
- `unikraft api /v1/volumes name=data size_mb:=256`
- `unikraft api /v1/something 'tags[]=production' 'tags[]=eu-west'`

Internal issue: https://linear.app/unikraft/issue/TOOL-927/use-nested-json-syntax-in-unikraft-api

khatibomar:
Fixes unikraft-cloud/cli#369.  <br>Closes unikraft-cloud/cli#369.

Implement nested JSON in api subcommand

### DIFF
--- a/internal/cmd/api.go
+++ b/internal/cmd/api.go
@@ -23,15 +23,17 @@ import (
 
 	"unikraft.com/cli/internal/config"
 	"unikraft.com/cli/internal/httpclient"
+	"unikraft.com/cli/internal/jason"
 )
 
 type APICmd struct {
 	Endpoint string   `arg:"" help:"API endpoint path (e.g. /v1/instances). May be a full URL, in which case the metro endpoint is ignored."`
-	Method   string   `short:"X" name:"method" help:"HTTP method to use. Defaults to GET, or POST if --data is set." placeholder:"method"`
+	Method   string   `short:"X" name:"method" help:"HTTP method to use. Defaults to GET, or POST if data is provided." placeholder:"method"`
 	Metro    string   `help:"Metro to target. Defaults to the profile's default metro." placeholder:"name"`
 	Header   []string `short:"H" name:"header" help:"Add an HTTP header in 'Key: Value' format. May be repeated."`
-	Data     string   `short:"d" name:"data" help:"HTTP request body. Use @path to read from a file, or @- to read from stdin."`
+	Data     string   `short:"d" name:"data" help:"(deprecated) Use positional arguments instead." hidden:""`
 	Insecure bool     `short:"k" name:"insecure" help:"Skip TLS certificate verification."`
+	Args     []string `arg:"" optional:"" name:"data" help:"Request body data. Can be JSON (e.g. '{\"key\":\"val\"}'), nested JSON (e.g. key=val or \"key=line value\"), @file, or @- (stdin)."`
 }
 
 func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
@@ -98,32 +100,25 @@ func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
 	}
 
 	// Resolve the request body.
-	var body io.Reader
-	if c.Data != "" {
-		switch {
-		case c.Data == "@-":
-			body = stdio.Stdin
-		case strings.HasPrefix(c.Data, "@"):
-			f, err := os.Open(c.Data[1:])
-			if err != nil {
-				return jujuerrors.Annotate(err, "reading request body")
-			}
-			defer f.Close()
-			body = f
-		default:
-			body = strings.NewReader(c.Data)
-		}
+	bodyBytes, err := c.resolveBody(ctx, stdio)
+	if err != nil {
+		return err
 	}
 
 	method := strings.ToUpper(c.Method)
 	if method == "" {
-		if c.Data != "" {
+		if c.Data != "" || len(c.Args) > 0 {
 			method = http.MethodPost
 		} else {
 			method = http.MethodGet
 		}
 	}
-	req, err := http.NewRequestWithContext(ctx, method, reqURL, body)
+
+	var bodyReader io.Reader
+	if bodyBytes != nil {
+		bodyReader = bytes.NewReader(bodyBytes)
+	}
+	req, err := http.NewRequestWithContext(ctx, method, reqURL, bodyReader)
 	if err != nil {
 		return err
 	}
@@ -131,7 +126,7 @@ func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
 		req.Header.Set("Authorization", "Bearer "+profile.Token)
 	}
 	req.Header.Set("Accept", "application/json")
-	if c.Data != "" && req.Header.Get("Content-Type") == "" {
+	if bodyBytes != nil && req.Header.Get("Content-Type") == "" {
 		req.Header.Set("Content-Type", "application/json")
 	}
 	for _, h := range c.Header {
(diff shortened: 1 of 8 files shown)


internal/jason/jason.go
```diff
@@ -0,0 +1,528 @@
...
+func Unmarshal[T any](data []byte, v *Jason[T]) error {
+	if err := json.Unmarshal(data, &v.Value); err == nil {
+		return nil
+	}
+	s := strings.TrimSpace(string(data))
+	if s == "" {
+		return json.Unmarshal([]byte("{}"), &v.Value)
+	}
+	var items []string
+	for line := range strings.SplitSeq(s, "\n") {
```
khatibomar: currently it expects that input is being seperated with a `\n` for each element.

Need some thinking.

internal/cmd/api.go
```diff
@@ -188,6 +183,79 @@ func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
...
+	if len(items) == 0 {
+		return nil, nil
+	}
+
+	var bodyData jason.Jason[map[string]any]
+	data := []byte(strings.Join(items, "\n"))
+	if err := jason.Unmarshal(data, &bodyData); err != nil {
+		return nil, err
+	}
+	return jason.Marshal(bodyData)
```
jedevc: Instead of this, you could just loop over all `items`, and unmarshal them one-by-one.

So essentially:
- `tags = []`
- `Unmarshal("tags[]=production")`
- `tags = [production]`
- `Unmarshal("tags[]=eu-west")`
- `tags = [production, eu-west]`

internal/cmd/api.go
```diff
@@ -188,6 +183,79 @@ func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
...
+		log.G(ctx).Warn().Msg("-d/--data is deprecated; use positional arguments instead")
+		srcItems, err := readBodyItems(c.Data, stdio)
+		if err != nil {
+			return nil, err
+		}
+		items = append(items, srcItems...)
+	}
+
+	for _, arg := range c.Args {
+		if strings.HasPrefix(arg, "@") {
```
jedevc: We have a check here and also in `readBodyItems`. It's probably worth consolidating them, ideally if we could just have all the parsing logic in one place, like `readBodyItems`.

internal/jason/jason.go
jedevc, line 1: This is so neat, I quite like this.

Any reason for naming it Jason? It's cute.

My only real big comment here is that we should split the package down. `jason.go` contains top-level public interfaces, while the parsing goes in `parse.go`, debug goes in `debug.go`, etc.

internal/jason/jason.go
```diff
@@ -0,0 +1,528 @@
...
+//	}
+func (n *Jason[T]) Visualize() string {
+	var b strings.Builder
+	visualize(&b, n.Value, 0)
+	return b.String()
+}
+
+// VisualizeItems is a convenience function that parses the given items
+// and returns a tree visualization of the result.
+func VisualizeItems(items []string) (string, error) {
```
jedevc: Probably best to just have `Visualize`.

internal/jason/jason.go
```diff
@@ -0,0 +1,528 @@
...
+			b.WriteString(" = ")
+			visualize(b, val[k], depth+1)
+			b.WriteString("\n")
+		}
+		b.WriteString(indent)
+		b.WriteString("}")
+	case []any:
+		b.WriteString("[\n")
+		for i, elem := range val {
+			b.WriteString(indent)
```
jedevc: You can use `fmt.Fprintf(b, "%s  [%d] =`, indent, i)` to simplify some of this string building.

internal/jason/jason.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+}
+
+// DebugTree returns a detailed multi-line breakdown showing how each
+// input item is parsed and where it lands in the final tree. This is
+// useful for understanding the parsing of complex nested expressions.
+//
+// The output shows every item grouped by its top-level key, with parsed
+// path segments and the final assigned value.
+func DebugTree(items []string) (string, error) {
```
jedevc: Can make this `debugTree` to keep it internal?

internal/jason/jason.go
jedevc: See above, we can allow multiple `Unmarshal`s.

internal/jason/jason.go
```diff
@@ -0,0 +1,528 @@
...
+//	foo[bar][]=x      → path=["foo","bar",""], value="x",     isRaw=false
+//	[0][key]=val      → path=["0","key"],       value="val",  isRaw=false
+//	{"key":"val"}     → path=[],               value=JSON,    isRaw=true
+//	[1,2,3]           → path=[],               value=JSON,    isRaw=true
+//
+// The path is accumulated by scanning a leading segment followed by zero
+// or more bracketed segments. After the path an '=' or ':=' separates the
+// value. If the item is a valid JSON object or array it is returned as a
+// raw value with an empty path so callers can merge it at the root.
+func parseItem(item string) (parsedItem, error) {
```
jedevc: Note to self: read and review this code in a little depth.

internal/jason/jason_test.go
```diff
@@ -0,0 +1,820 @@
...
+	output, err := VisualizeItems([]string{
+		`{"name":"test"}`,
+		"count:=5",
+	})
+	require.NoError(t, err)
+	expected := "{\n  count = 5\n  name = \"test\"\n}"
+	assert.Equal(t, expected, output)
+}
+
+func ExampleDebugTree() {
```
jedevc: Fun little idea. Because this writing a new (fairly fiddly) parser, what do you think about writing a little fuzzer to search for issues? https://go.dev/doc/security/fuzz/

Then just run it for an hour or so, and see if it can find any crashes :tada: (we've had some really annoying and difficult crashes in the last week, so hardening things to avoid needing to debug broken stuff later is very nice).

jedevc (commented):
Sorry for my slowness! I finally got round to it (chaos week sorry) - just some early feedback, I saw this is still in draft. But it's *so* neat!

Much love for this new `Jason` type :tada:

internal/cmd/api.go
khatibomar: the way I am thinking of it, is that, a valid JSON is `[....]` or `{}` with proper syntax that have proper key value.

On other hand a valid nested-JSON is list of key=value pairs or valid JSON

so jason.Unmarshal somehow should take all of these and parse them as a one unit, that was my vision behind that Unmarshaling.

What we can do is keep parser to expect a \n to seperate them with clear comment about expected input, but in api.go we pass items one by one instead of joining them which also results in a valid end result. 
Another thing we can do is make the parser bit smarter about spaces, but that ofc will come with complexity cost depends if spaces are allowed or not. For example `tags [] = production` is this valid nested-JSON or not, based on that the parsing becomes complex.

internal/jason/jason.go
khatibomar: I first named it nested, but it's not only nested, it's nested-JSON + JSON together.
Couldn't find a better name,  jason was closest to my head because of a meme stuck in my head from youtube I guess. Prob Primeagen can't remember.

JavaScript Advanced Object Notation 🥼

jedevc: Aha, I love that :tada: Maybe just make a note of it in the file, then it's all good :grin:

khatibomar:
@jedevc the PR is out of draft and ready for review, I modified the parser to apply what I mentioned here https://github.com/unikraft-cloud/cli/pull/389#discussion_r3523998270 
also added documentation and did some fuzzy testing run each for 1 hour and got 2 bugs that crashed app! useful stuff.

jedevc:
Have a copilot review :tada:

craciunoiuc:
Can you close the comments above that have been resolved? I'm starting to do a pass now also

internal/cmd/api.go
```diff
@@ -188,6 +183,171 @@ func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
 	return nil
 }
 
+func (c *APICmd) resolveBody(ctx context.Context, stdio config.Stdio) ([]byte, error) {
+	var body bytes.Buffer
+	var wroteSource bool
+	legacyOnly := c.Data != "" && len(c.Args) == 0
+
+	if c.Data != "" {
+		log.G(ctx).Warn().Msg("-d/--data is deprecated; use positional arguments instead")
```
craciunoiuc, line 192: Should we also warn here? We already made the flag deprecated and hidden.

Maybe a warn is a bit too annoying. Maybe info or debug for now? 🤔

internal/cmd/api.go
```diff
@@ -188,6 +183,171 @@ func (c *APICmd) Run(ctx context.Context, stdio config.Stdio) error {
...
+  {...}        Inline JSON object (merges with other arguments).
+  [...]        Inline JSON array (replaces the entire body).
+  key=value    Set a key to a literal string value.
+  key:=raw     Set a key to a raw JSON value (number, boolean, null, object,
+               or array constructed from raw JSON).
+
+NESTED JSON SYNTAX
+
+When you use the key=value / key:=raw forms, you can build complex nested
+structures using bracket notation:
```
craciunoiuc: I would reword this maybe to be more general, maybe smth like (personal preference):


```suggestion
key=value / key:=raw forms can be used to build complex nested
structures:
```

internal/jason/debug.go
```diff
@@ -0,0 +1,165 @@
...
+//
+//	{
+//	  name = "HTTPie"
+//	  stars = 54000
+//	  apps = [
+//	    [0] = "Terminal"
+//	    [1] = "Desktop"
+//	    [2] = "Web"
+//	  ]
+//	}
```
craciunoiuc, line 30: Only spaces here pls 😳

internal/jason/jason.go
```diff
@@ -0,0 +1,51 @@
...
+//
+// Nested-item input expresses a path to each value using bracket notation.
+// Use '=' for literal strings and ':=' for raw JSON values.
+//
+// See https://httpie.io/docs/cli/nested-json for the input format.
+//
+// I first named it nested, but it's not only nested, it's nested-JSON + JSON together.
+// Couldn't find a better name, jason was closest to my head because of a meme stuck in my head from youtube I guess. Prob Primeagen can't remember.
+//
+// why name it JASON? https://github.com/unikraft-cloud/cli/pull/389#discussion_r3524002021
```
craciunoiuc, line 17: Remove this part 😭 (though I like it)

internal/jason/parse.go
```diff
@@ -0,0 +1,459 @@
...
+	"unicode"
+	"unicode/utf8"
+)
+
+const maxArrayIndex = 10_000
+
+// parsedItem stores one parsed nested-item entry.
+//
+// path holds the bracketed segments, value holds the right-hand side, and isRaw
+// reports whether the value came from ':=' and should be decoded as JSON.
```
craciunoiuc: Maybe worth moving the comment down to each struct item? aka splitting in three comments?

internal/jason/parse.go
```diff
@@ -0,0 +1,459 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package jason
+
```
craciunoiuc, line 7: Considering how much parsing Jason has, I would suggest running the robot in planning mode to expand the unit tests for it a bit. It's really easy to lose corner cases for this

craciunoiuc (changes requested):
Hey, left some comments

Mostly related to code comments and tests.

One more note on that: Can you add 1-2-3 more e2e tests that use this new formatting, I think our `unikraft api` tests are a bit simple

internal/jason/jason.go
khatibomar: BUT, justin asked for it.

Someone will walk out disappointed.

internal/jason/parse.go
khatibomar: Ham, we are at Friday, maybe I can use Ultracode so it sucks what is left from weekly bandwith

internal/jason/debug.go
khatibomar: context/why?

internal/jason/parse.go
craciunoiuc: yup exactly :yum: 

Though ultracode is a bit weird, after setting a prompt I always see it after set on XHigh? Not sure what is going on

So maybe try Max instead? no idea

Only request: create as a separate commit so they don't get mixed up with the things we already reviewed

internal/jason/debug.go
craciunoiuc: image

Well, the tab/space mix is a bit weird and misleading. Replacing the tabs with 2 spaces should be fine

internal/jason/jason.go
craciunoiuc: It is war then

I'm also fine with calling the package `justin` (Just ur Usual String-Typed INput)

internal/cmd/api.go
khatibomar: I think that's the point of warning to be annoying and noticeable so users switch to modern alternative.

internal/jason/debug.go
khatibomar: Sad news, 2 spaces are not gofmt friendly so it's all now tabs.

internal/jason/parse.go
khatibomar: Okay, so mr robot did stuff and pushed them, mainly a bug in code pushed today.

https://github.com/unikraft-cloud/cli/pull/389/commits/fc4b6c61e4f003041068439efb0d46e70c46d4fb

craciunoiuc (commented):
I actually don't have anything else against Jason

I'll let @jedevc approve

craciunoiuc (approved):
Actually I need to approve to change my status

```
```

internal/jason/fuzz_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	f.Fuzz(func(t *testing.T, input string) {
+		data, err := buildNestedJSON(input)
+		if err != nil {
+			return
+		}
+
+		var value any
+		if err := json.Unmarshal(data, &value); err != nil {
+			t.Fatalf("buildNestedJSON produced invalid JSON: %v\ninput: %q\noutput: %s", err, input, data)
```
jedevc, line 54: `require.NoError` 🙏

jedevc (approved):
```
```

PR #395, opened 2026-07-03
fix(parse)!: Remove csv-style parsing for lists and maps

nurof3n:
Closes TOOL-1125

### DIFF
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -83,29 +83,29 @@ type InstanceCreateCmd struct {
 	Type       *platform.InstanceType `group:"flag-create" shortcut:"type" help:"Type of virtual machine to run. \"full\" requires a plan with full VM support." placeholder:"type" example:"micro,full"`
 
 	Args InstanceArgs `group:"flag-create" shortcut:"runtime.args" help:"Arguments to pass to the instance." placeholder:"arg"`
-	Env  []string     `group:"flag-create" shortcut:"runtime.env" short:"e" sep:"none" help:"Environment variable. Repeat the flag to set more than one." placeholder:"<key>=<value>" example:"DEBUG=true"`
+	Env  []string     `group:"flag-create" shortcut:"runtime.env" short:"e" sep:"none" help:"Environment variable." placeholder:"<key>=<value>" example:"DEBUG=true"`
 
 	Memory        types.SizeMebibytes     `group:"flag-create" shortcut:"resources.memory" short:"m" help:"Memory allocation." placeholder:"size" example:"128MiB,1GiB"`
 	Vcpus         int                     `group:"flag-create" shortcut:"resources.vcpus" help:"Number of vCPUs." placeholder:"n" example:"1,2,4"`
 	Gpus          int                     `group:"flag-create" shortcut:"resources.gpus" help:"Number of GPUs to attach. Requires type \"full\" and a plan with GPU support. Currently limited to 1." placeholder:"n" example:"0,1"`
 	SchedPriority *platform.SchedPriority `group:"flag-create" shortcut:"sched-priority" help:"Scheduling priority for the instance." placeholder:"priority" example:"normal,medium,high,admin"`
 
-	Volume []InstanceVolume `group:"flag-create" shortcut:"volumes" short:"v" help:"Attach volume." placeholder:"<name>:<path>[:<options>]" example:"my-vol:/data,cache:/tmp:ro,data:/mnt:size=10GiB"`
+	Volume []InstanceVolume `group:"flag-create" shortcut:"volumes" short:"v" sep:"none" help:"Attach volume." placeholder:"<name>:<path>[:<options>]" example:"my-vol:/data"`
 	Rom    []InstanceRom    `group:"flag-create" shortcut:"roms" sep:"none" help:"Attach ROM." placeholder:"image=<ref>,at=<path>" example:"image=myuser/my-rom:latest\\,at=/rom0\\,name=my-rom,dir=./mydata\\,at=/rom"`
 
 	Service InstanceService `group:"flag-create" shortcut:"service" help:"Service group name or key." placeholder:"name"`
-	Publish []Service       `group:"flag-create" shortcut:"service.services" short:"p" help:"Publish port." placeholder:"<src>:<dest>[/<handlers>]" example:"443:8080/http+tls,80:8080/http"`
-	Domain  []Domain        `group:"flag-create" shortcut:"service.domains" help:"Service domain." placeholder:"fqdn" example:"example.com,api.example.com"`
+	Publish []Service       `group:"flag-create" shortcut:"service.services" short:"p" sep:"none" help:"Publish port." placeholder:"<src>:<dest>[/<handlers>]" example:"443:8080/http+tls"`
+	Domain  []Domain        `group:"flag-create" shortcut:"service.domains" sep:"none" help:"Service domain." placeholder:"fqdn" example:"example.com"`
 
 	ScaleToZero InstanceScaleToZero `group:"flag-create" shortcut:"scale-to-zero" help:"Scale-to-zero options.\n  policy: on | idle | off\n  cooldown-time: cooldown in ms before scaling to zero\n  notify-time: notification time in ms before scaling to zero\n  stateful: true | false" placeholder:"<key>=<value>" example:"on,policy=on\\,cooldown-time=300,policy=on\\,stateful=true\\,cooldown-time=500\\,notify-time=100"`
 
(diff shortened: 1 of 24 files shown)


internal/cmd/instances.go
```diff
@@ -102,6 +101,15 @@ type InstanceCreateCmd struct {
 }
 
 func (c *InstanceCreateCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *resource.Sandbox, kctx *kong.Context) error {
```
craciunoiuc, line 115: Is instance creation the only place where you can set new env vars?

Can they be set also in instance templates?

internal/cmd/instances.go
```diff
@@ -146,12 +154,108 @@ type InstanceEditCmd struct {
...
+
+	last := strings.TrimSpace(input[start:])
+	if last != "" {
+		assignments = append(assignments, last)
+	}
+
+	return assignments
+}
+
+func firstCommaToken(input string) string {
```
craciunoiuc: `GPT-5.3 Codex` spotted

Can you go through the helper functions that are single use, and inline them a bit?

Some of them, like this one, are really not needed

craciunoiuc: For the ones that remain, can you add some unit tests. Parsers are very good for unit tests hehe

Also add an integration test that sets these comma separated values

internal/cmd/instances.go
```diff
@@ -146,12 +154,108 @@ type InstanceEditCmd struct {
...
+		// Prevent generic shortcut handling from re-processing --env values.
+		c.Env = nil
+	}
 	if err := cmd.ApplyShortcutFlags(&c.SetArgs, kctx.Flags()); err != nil {
 		return err
 	}
 	return c.ResourceEditCmd.Run(ctx, stdio, sandbox)
 }
 
+func runtimeEnvSetArgs(inputs []string) ([]map[string]string, error) {
```
craciunoiuc: TBD: if we should move these to a small internal subpackage

craciunoiuc (changes requested):
Some comments my side 🥺

internal/cmd/instances.go
```diff
@@ -130,7 +138,7 @@ type InstanceEditCmd struct {
 	Image string `group:"flag-edit" shortcut:"image" help:"Image to deploy." placeholder:"<name>:<tag>" example:"nginx:latest,my-app:v1.2.3"`
 
 	Args InstanceArgs `group:"flag-edit" shortcut:"runtime.args" help:"Arguments to pass to the instance." placeholder:"arg"`
-	Env  []string     `group:"flag-edit" shortcut:"runtime.env" short:"e" help:"Environment variables." placeholder:"<key>=<value>" example:"DEBUG=true,PORT=8080"`
+	Env  []string     `group:"flag-edit" shortcut:"runtime.env" short:"e" sep:"none" help:"Environment variables." placeholder:"<key>=<value>" example:"DEBUG=true,PORT=8080"`
```
jedevc: Hm, my suggestion was actually simpler. Instead of guessing what env vars look like, just require that `sep:"none"`, then we *require* that `-e` has only one var at a time, not multiple vars.

Essentially see https://docs.docker.com/reference/cli/docker/container/run/#env.

This *would* be a breaking change, but I think it's the right call.

jedevc: TL;DR I think `-e FOO=foo,BAR=bar` should be invalid. You should have to do `-e FOO=foo -e BAR=bar`.

nurof3n: Rewrote this under the assumption that we no longer accept comma-separated env variables.

craciunoiuc:
@nurof3n tests failing :flushed: 

Also please close copilot comments which no longer apply

Looks much leaner now :biting_lip:

internal/resource/value/parse.go
```diff
@@ -174,25 +174,34 @@ func parseReflect(input []string, value reflect.Value) error {
...
+			// (e.g. map[string]string) the value is taken verbatim, so it may
+			// contain commas, "=", or any other characters. For non-string
+			// element types (e.g. slices, structs), parseReflect may still
+			// interpret commas per the element type's own parsing rules.
+			// Multiple entries are supplied via repeated --set flags (multiple
+			// input strings), not via commas within a single string.
+			//
+			// Skip empty/whitespace-only inputs so a bare "--set field=" can
+			// be used without creating a spurious empty-key entry, and trim
+			// the key only (values are kept verbatim).
```
craciunoiuc: summarize the AI comments or you get the whip

internal/resource/value/parse_test.go
```diff
@@ -298,3 +280,69 @@ func TestParseCSV(t *testing.T) {
 		assert.Equal(t, testStruct{Name: "hello world", Value: 42}, got)
 	})
 }
+
+// TestParseMapLiteralValues covers the behaviour that each --set input for a
+// map field is a single key=value entry: the value is taken verbatim after the
+// first "=", so commas and "=" signs in the value are preserved rather than
+// being interpreted as entry separators.
```
craciunoiuc: you can remove test comments unless the test is really hard to understand

internal/resource/value/parse.go
jedevc: YESSSS, this is what I meant! Perfect :tada: Incredible find 🎉 

I will say though... this might have an impact on other things. Take a quick look and see where else it impacts. Also, note there's above logic for structs where we have similar logic actually!

We could also make the previous `range` behavior optional, similar to `kong`, and obey a `sep` tag on things. If it's `none` then we skip it, otherwise we split on the value.

nurof3n: I tried to do that, let me know if it's too complicated again :x

internal/resource/value/parse.go
```diff
@@ -174,26 +178,26 @@ func parseReflect(input []string, value reflect.Value) error {
...
-					return err
-				}
-				val := reflect.New(output.Type().Elem()).Elem()
-				err = parseReflect([]string{v}, val)
-				if err != nil {
-					return err
-				}
-				mapp.SetMapIndex(key, val)
+
+			// Each input string is a single key=value entry. Split on the
```
jedevc: Shouldn't this section also use `sep`? Right now, it's only being used for slices+structs, not maps.

internal/resource/value/parse.go
```diff
@@ -259,3 +264,64 @@ func parseReflect(input []string, value reflect.Value) error {
...
+	if sep == "none" {
+		return []string{input}
+	}
+	if sep == "" {
+		sep = ","
+	}
+	return strings.Split(input, sep)
+}
+
+func splitStructValues(input string, s reflect.Value, sep string) []string {
```
jedevc: 🤔 Hm, wait why aren't we just using `splitStructValues` here?

Also, again, I don't love methods like `looksLikeStructField` - this means it's going to magically try and work out how to parse something, which is unexpected and will make the user sad.

jedevc:
This is better though <3 I've renamed the title to indicate this is a breaking change.

internal/resource/value/parse.go
nurof3n: the `map` case now uses the same splitter, with `sep`

internal/resource/value/parse.go
nurof3n: deleted stuff like that, it now only uses `splitValues`

craciunoiuc (dismissed):
Does this break anything other than scripts that use the cli? I'm thinking of what this could touch 🤔 

Sounds like not that bad overall

```
```

craciunoiuc:
also CI died :flushed:

craciunoiuc (dismissed):
All good here. Thanks!

```
```

craciunoiuc:
that's a lot of edge cases, now I'm scared there are more :eyes:

jedevc:
The last commit is a pre-existing issue, I can split to a separate PR tbf.

craciunoiuc:
nah, no need

but do check if there are any cases left

I will also do a pass soon, might start a robot in parallel to look for any edge cases

craciunoiuc:
There are some weird comments but today I feel like that sugar should not be hoisted on the chopping block

PR looks ok, I'm testing this cool extension and then reporting what it found

image

craciunoiuc:
static check failed, in case you are wondering why, it's because golangci-lint got updated and the new version has more checks

in X it's better to pin down, but here maybe we can live with the bump?

that being said, should the fix be here or in a separate PR?

jedevc:
Let's fix it separately, yes. I'll pin it down in x.

craciunoiuc (approved):
All good here. Thanks!

```
```

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #400, opened 2026-07-06
feat: Allow partial success results

nurof3n:
To be merged after https://github.com/unikraft-cloud/cli/pull/388
Closes TOOL-1048

### DIFF
--- a/internal/cmd/certificates.go
+++ b/internal/cmd/certificates.go
@@ -112,16 +112,14 @@ func (Certificate) List(ctx context.Context) ([]resource.Resource, error) {
 	}
 	return group.CollectAllSlices(ctx, g, func(ctx context.Context, c multimetro.MetroClient) ([]resource.Resource, error) {
 		log.G(ctx).Trace().Msg("listing certificates")
-		resp, err := c.GetCertificates(ctx, nil, platform.GetCertificatesOpts{Details: new(true)})
-		if err != nil {
-			return nil, err
+		resp, opErr := c.GetCertificates(ctx, nil, platform.GetCertificatesOpts{Details: new(true)})
+		var certificates []platform.Certificate
+		if resp != nil && resp.Data != nil {
+			certificates = resp.Data.Certificates
 		}
 		var results []resource.Resource
-		var errs []error
-		if resp == nil || resp.Data == nil {
-			return nil, nil
-		}
-		for _, certificate := range resp.Data.Certificates {
+		errs := []error{opErr}
+		for _, certificate := range certificates {
 			result, err := Certificate{}.load(nil, certificate, &c.Metro)
 			if err != nil {
 				errs = append(errs, err)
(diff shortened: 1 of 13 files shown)


jedevc:
Ho hum. Can I get an explanation of what this one is for? :thinking: It's not immediately obvious to me sorry!

nurof3n:
> Ho hum. Can I get an explanation of what this one is for? 🤔 It's not immediately obvious to me sorry!

@jedevc  Yup, so the router can return a 207 on listing or bulk requests when some of the metros returned errors, and it would contain both the results for successful calls and the errors for the others.
And I needed a way to show the error in the CLI and also show the listing output.

internal/cmd/instances.go
```diff
@@ -716,21 +726,29 @@ func (Instance) Delete(ctx context.Context, keys []string) error {
...
-			deleted = append(deleted, group.Ref{
-				Metro: c.Metro.Name,
-				Name:  instance.Name,
-				UUID:  instance.Uuid,
-			})
 		}
-		return deleted, nil
+		// If there were deletions despite the error, report partial success
+		if len(deleted) > 0 && err != nil && !platform.ErrorContainsOnly(err, platform.APIHTTPErrorNotFound) {
+			pr := NewPartialResult()
```
jedevc: We should match how `Get` already handles this? And just use `errors.Join`.

jedevc:
> > Ho hum. Can I get an explanation of what this one is for? 🤔 It's not immediately obvious to me sorry!
> 
> @jedevc Yup, so the router can return a 207 on listing or bulk requests when some of the metros returned errors, and it would contain both the results for successful calls and the errors for the others. And I needed a way to show the error in the CLI and also show the listing output.

Makes sense.

The bit that I don't get is the new helpers in `utils.go`. Why `listGetOpError` and similar? IMO, if the response has a partial-success (i.e. `resp.Status == "partial_success"` and status is 207), we should never abort processing. We should return both all the errors and all the successful responses. It feels like it's a bit overcomplicated but maybe I'm missing something.

internal/cmd/instances.go
nurof3n: Done — `PartialResult` and the other helpers are gone. Delete now builds the deleted list straight from the response and returns `ignoreNotFound(err)`, which is the same shape `Get` uses, and everything is combined with a plain `errors.Join`.

One caveat worth recording while it is fresh: `group.DoRefs` only records the refs a callback returns when that callback returns a nil error. So the deleted list reaches the caller only on the not-found path, where `ignoreNotFound` nils the error out and `DoRefs` reports the missing refs itself. For a 207 that mixes not-found with anything else, the helper still drops them — surfacing those would need a change on the SDK side. I reworded the comment so it describes what actually happens instead of overclaiming.

— Alex-Andrei Cioc <andrei.cioc@unikraft.io>

nurof3n:
@jedevc My bad, I simplified it and now we don't abort processing on partial success.
I also added some tests

internal/cmd/certificates.go
jedevc: Ah, yes, because `errors.Join` ignores `nil` errors, we can simplify this:

```suggestion
errs = append(errs, err)
```

Keeps this block nice and small.

internal/cmd/instances.go
```diff
@@ -1250,9 +1244,12 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 		return nil, err
 	}
 	results, err := Instance{}.Get(ctx, keys.Strings())
-	if err != nil {
+	if err != nil && len(results) == 0 {
 		return nil, err
 	}
+	if err != nil {
+		log.G(ctx).Warn().Err(err).Msg("instance created with partial lookup errors")
+	}
```
jedevc: How can this happen? I'm not sure I'd know what to do if I was greeted with this error.

jedevc (commented):
Ahh, this is super neat. I like this a lot :tada:

Small comments, good finds all around.

jedevc (approved):
@nurof3n can you do a quick pass? I made some modifications, can you check this still works as you intended, but I'd like to merge this now! :tada:

```
```

nurof3n (commented):
changes look good to me!
gotta test for regression as well

PR #412, opened 2026-07-08
fix(integration): Several fixes to get tests to pass

craciunoiuc:


cmd/unikraft/integration/instance_test.go
```diff
@@ -191,20 +191,30 @@ func TestInstances(t *testing.T) {
...
+`), 0o644),
 			fstest.CreateFile("base/Kraftfile", []byte(`
 spec: v0.7
 name: busybox-start-follow-e2e
 runtime: base-compat:latest
 rootfs:
   format: erofs
   source: ./Dockerfile
-cmd: ["cat", "/rom/hello.txt"]
+cmd: ["sleep", "30s"]
```
jedevc: Why is there a cmd here at all? In the diff, why was it looking at a ROM?

When we run this image, we always set a `cmd`, so it always get overridden.

jedevc (commented):
Tests still failing 🤔

cmd/unikraft/integration/instance_test.go
craciunoiuc: Yes, previously it was a typo

Right now I set it to something that is never reached so it has a cmd

Cause I'm not sure it behaves without a cmd set in there. I guess it's fine?

jedevc: It should be fine yup, you can leave it off.

craciunoiuc: ok, I removed the line

jedevc (dismissed):
Great, this is just the kernel<->platform failure.

```
```

jedevc (approved):
```
```

PR #413, opened 2026-07-09
feat(builder): Add rootfs implementation for reading OCI images

craciunoiuc:
Tested work and it seemed fine. Pasting again uses:
Case 1:
```
rootfs:
  format: erofs
  type: oci
  source: index.unikraft.io/test/check-pr:latest
```

Case 2:
```
rootfs:
  format: erofs
  type: oci
  source: index.docker.io/hello-world:latest
```

~~You can use these prefixes: `oci://`, `oci-layout://`, `oci-archive://`  taken from `image-spec`:~~
```
case imagespec.URISchemeOCI, imagespec.URISchemeOCILayout, imagespec.URISchemeOCIArchive:
```

We decided to use only the oci type variable.

~~Also needs another bump to the gomod.~~

Closes: TOOL-1073
Depends-on: https://github.com/unikraft-cloud/x/pull/400

### DIFF
--- a/internal/builder/build.go
+++ b/internal/builder/build.go
@@ -231,19 +231,9 @@ func Build(ctx context.Context, opts BuildOpts) ([]*imagespec.Image, error) {
 			imgOpts = append(imgOpts, imagespec.WithInitrd(roots[i].Initrd))
 			roots[i].Initrd = nil
 			// The rootfs build may have produced a richer config (e.g. from
-			// a Dockerfile). Use it as the base and layer our overrides on top.
-			cfg = roots[i].Image.Config
-			if opts.Cmd != nil {
-				cfg.Cmd = opts.Cmd
-			}
-			if opts.Env != nil {
-				env := make([]string, 0, len(opts.Env))
-				for _, kv := range opts.Env {
-					env = append(env, fmt.Sprintf("%s=%s", kv.Key, kv.Value))
-				}
-				cfg.Env = append(env, cfg.Env...)
-			}
-			cfg.Labels = opts.Labels
+			// a Dockerfile or an OCI image). Use it as the base and layer our
+			// overrides on top.
+			cfg = applyConfigOverrides(roots[i].Image.Config, opts)
 		}
 		imgOpts = append(imgOpts, imagespec.WithImageConfig(cfg))
 
--- a/internal/builder/kraftfile.go
+++ b/internal/builder/kraftfile.go
@@ -6,9 +6,7 @@
 package builder
 
 import (
-	"cmp"
 	"fmt"
-	"path/filepath"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 
@@ -58,22 +56,17 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
 		if rom.Source == nil || rom.Source.Path == "" {
 			return BuildOpts{}, fmt.Errorf("rom entry is missing a source path")
 		}
-		romPath := filepath.Join(dir, rom.Source.Path)
-		romFormat := cmp.Or(rom.Format, kraftfile.FsTypeErofs)
+		romFormat := defaultRomFormat(rom.Format, rom.Source.Type)
 		romOpt := FSOpts{
-			Path:   romPath,
+			Path:   rom.Source.Path,
 			Format: romFormat,
 			Type:   rom.Source.Type,
 			// Pad the file to page-size alignment. This is required by the platform
 			// which rejects ROM files that are not page-aligned.
 			Pad: 4096,
 		}
-		if romOpt.Type == "" {
-			typ, err := DetectSourceType(romPath)
-			if err != nil {
-				return BuildOpts{}, fmt.Errorf("detecting rom type for %q: %w", romPath, err)
-			}
-			romOpt.Type = typ
+		if err := resolveSource(dir, &romOpt); err != nil {
+			return BuildOpts{}, fmt.Errorf("resolving rom source %q: %w", rom.Source.Path, err)
 		}
 		opts.Roms = append(opts.Roms, romOpt)
 	}
@@ -82,18 +75,12 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
 		if kf.Rootfs.Source == nil || kf.Rootfs.Source.Path == "" {
 			return BuildOpts{}, fmt.Errorf("rootfs entry is missing a source path")
 		}
-		opts.Rootfs.Path = filepath.Join(dir, kf.Rootfs.Source.Path)
+		opts.Rootfs.Path = kf.Rootfs.Source.Path
 		opts.Rootfs.Format = kf.Rootfs.Format
 		opts.Rootfs.Type = kf.Rootfs.Source.Type
 		opts.Rootfs.Dockerfile = kf.Rootfs.Source.Dockerfile
-		if opts.Rootfs.Dockerfile != "" && opts.Rootfs.Type == "" {
-			opts.Rootfs.Type = kraftfile.SourceTypeDockerfile
-		} else if opts.Rootfs.Type == "" {
-			typ, err := DetectSourceType(opts.Rootfs.Path)
-			if err != nil {
-				return BuildOpts{}, fmt.Errorf("detecting rootfs type for %q: %w", opts.Rootfs.Path, err)
-			}
-			opts.Rootfs.Type = typ
+		if err := resolveSource(dir, &opts.Rootfs); err != nil {
+			return BuildOpts{}, fmt.Errorf("resolving rootfs source %q: %w", kf.Rootfs.Source.Path, err)
 		}
 	}
 
--- a/internal/builder/rootfs.go
+++ b/internal/builder/rootfs.go
@@ -22,6 +22,7 @@ import (
 	"github.com/containerd/platforms"
 	dockerconfig "github.com/docker/cli/cli/config"
 	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/exporter/containerimage/exptypes"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/identity"
@@ -39,26 +40,34 @@ import (
 	"unikraft.com/cli/internal/buildkit"
 	"unikraft.com/cli/internal/config"
 	"unikraft.com/cli/internal/images"
+	ukio "unikraft.com/x/io"
 	"unikraft.com/x/kraftfile"
 	"unikraft.com/x/log"
 )
 
-// buildImageConfig constructs a minimal OCI image config from build options.
-// Used when a BuildKit solve is not performed.
-func buildImageConfig(opts BuildOpts) ocispec.ImageConfig {
-	var cfg ocispec.ImageConfig
+// applyConfigOverrides layers the build options' Cmd and Env on top of base,
+// which is the config of the image the rootfs was built from. Env is prepended
+// so the caller's values take precedence. Labels replace the base's entirely.
+func applyConfigOverrides(base ocispec.ImageConfig, opts BuildOpts) ocispec.ImageConfig {
+	cfg := base
 	if opts.Cmd != nil {
 		cfg.Cmd = opts.Cmd
 	}
 	if opts.Env != nil {
-		env := make([]string, 0, len(opts.Env))
+		env := make([]string, 0, len(opts.Env)+len(cfg.Env))
 		for _, kv := range opts.Env {
 			env = append(env, fmt.Sprintf("%s=%s", kv.Key, kv.Value))
 		}
-		cfg.Env = env
+		cfg.Env = append(env, cfg.Env...)
 	}
 	cfg.Labels = opts.Labels
 	return cfg
+}
+
+// buildImageConfig constructs a minimal OCI image config from build options.
+// Used when there is no source image config to build on top of.
+func buildImageConfig(opts BuildOpts) ocispec.ImageConfig {
+	return applyConfigOverrides(ocispec.ImageConfig{}, opts)
 }
 
 // DetectSourceType inspects path and returns the detected RootfsType.
@@ -74,21 +83,15 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
 
 	fi, err := os.Stat(path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return "", fmt.Errorf("rootfs path does not exist")
-		}
-		return "", fmt.Errorf("checking rootfs source %q: %w", path, err)
+		return "", fmt.Errorf("checking rootfs source: %w", err)
 	}
 
 	switch {
 	case fi.IsDir():
 		return kraftfile.SourceTypeDirectory, nil
 	case fi.Mode().IsRegular(), fi.Mode()&os.ModeSymlink != 0:
-		if gocpio.IsValidPath(path) {
-			return kraftfile.SourceTypeCpio, nil
-		}
-		if goerofs.IsValidPath(path) {
-			return kraftfile.SourceTypeErofs, nil
+		if format, err := detectPackagedFormat(path); err == nil {
+			return kraftfile.SourceType(format), nil
 		}
 		if f, err := os.Open(path); err == nil {
 			defer f.Close()
@@ -104,6 +107,59 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
 	default:
 		return "", fmt.Errorf("could not detect rootfs type %q", path)
 	}
+}
+
+// detectPackagedFormat reports the rootfs format of an already-packaged file.
+func detectPackagedFormat(path string) (kraftfile.FsType, error) {
+	switch {
+	case gocpio.IsValidPath(path):
+		return kraftfile.FsTypeCpio, nil
+	case goerofs.IsValidPath(path):
+		return kraftfile.FsTypeErofs, nil
+	default:
+		return "", fmt.Errorf("could not detect rootfs format of %q", path)
+	}
+}
+
+// resolveSource resolves the source of fsOpts against root and fills in the
+// source type when it was not requested explicitly.
+func resolveSource(root string, fsOpts *FSOpts) error {
+	if fsOpts.Type == kraftfile.SourceTypeOCI {
+		if fsOpts.Dockerfile != "" {
+			return fmt.Errorf("a dockerfile cannot be set when the source type is %q", kraftfile.SourceTypeOCI)
+		}
+		return nil
+	}
+
+	if root != "" {
+		fsOpts.Path = filepath.Join(root, fsOpts.Path)
+	}
+
+	if fsOpts.Dockerfile != "" {
+		if fsOpts.Type != "" && fsOpts.Type != kraftfile.SourceTypeDockerfile {
+			return fmt.Errorf("source type must be %q when a dockerfile is set, got %q", kraftfile.SourceTypeDockerfile, fsOpts.Type)
+		}
+		fsOpts.Type = kraftfile.SourceTypeDockerfile
+	}
+
+	if fsOpts.Type == "" {
+		typ, err := DetectSourceType(fsOpts.Path)
+		if err != nil {
+			return err
+		}
+		fsOpts.Type = typ
+	}
+
+	return nil
+}
+
+// defaultRomFormat fills in the ROM default format. An OCI source carries its
+// own format, so leave it unset and let the source dictate it.
+func defaultRomFormat(format kraftfile.FsType, typ kraftfile.SourceType) kraftfile.FsType {
+	if format != "" || typ == kraftfile.SourceTypeOCI {
+		return format
+	}
+	return kraftfile.FsTypeErofs
 }
 
 func BuildRoms(ctx context.Context, opts BuildOpts) (_ [][]imagespec.File, rerr error) {
@@ -128,7 +184,7 @@ func BuildRoms(ctx context.Context, opts BuildOpts) (_ [][]imagespec.File, rerr 
 			Rootfs: FSOpts{
 				Path:   rom.Path,
 				Type:   rom.Type,
-				Format: cmp.Or(rom.Format, kraftfile.FsTypeErofs),
+				Format: defaultRomFormat(rom.Format, rom.Type),
 				Pad:    rom.Pad,
 			},
 			// propagate BuildKit options from the parent build
@@ -172,9 +228,12 @@ func BuildRootfs(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image, rer
 		return nil, fmt.Errorf("at least one platform must be specified")
 	}
 
+	// Sources that are already packaged carry their own format, so leave it
+	// unset and let the branch that knows the source fill it in.
 	if opts.Rootfs.Format == "" &&
 		opts.Rootfs.Type != kraftfile.SourceTypeCpio &&
-		opts.Rootfs.Type != kraftfile.SourceTypeErofs {
+		opts.Rootfs.Type != kraftfile.SourceTypeErofs &&
+		opts.Rootfs.Type != kraftfile.SourceTypeOCI {
 		opts.Rootfs.Format = DefaultRootfsFormat(opts.Platform)
 	}
 
@@ -219,6 +278,8 @@ func BuildRootfs(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image, rer
 			}
 		}
 		return buildRootfsDockerfile(ctx, opts)
+	case kraftfile.SourceTypeOCI:
+		return buildRootfsOCI(ctx, opts)
 	default:
 		return nil, fmt.Errorf("unsupported rootfs type %q", opts.Rootfs.Type)
 	}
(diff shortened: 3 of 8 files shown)


craciunoiuc:
let me know if these single use functions you want to inline

I found them significant enough to keep separate for now

internal/builder/rootfs_oci_test.go
```diff
@@ -0,0 +1,202 @@
...
+			OS:           "linux",
+		},
+		Config: ocispec.ImageConfig{
+			Cmd: []string{"/bin/sh"},
+		},
+		RootFS: ocispec.RootFS{
+			Type:    "layers",
+			DiffIDs: []digest.Digest{layerDesc.Digest},
+		},
+	}
```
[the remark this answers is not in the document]
craciunoiuc: @jedevc sounds :billed_cap: 

should I change?

internal/builder/rootfs.go
```diff
@@ -303,6 +327,215 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+			return nil, err
+		}
+		return []ocispec.Manifest{m}, nil
+	default:
+		return nil, fmt.Errorf("unexpected media type %q for %s", desc.MediaType, desc.Digest)
+	}
+}
+
+// applyLayer reads a single OCI layer from provider and applies it to dir.
+func applyLayer(ctx context.Context, provider content.Provider, desc ocispec.Descriptor, dir string) error {
```
jedevc: This is a bit dangerous - remember the issues around creating permissions, device files, etc.

We can't create these directly in a `dir`. We need to do something a bit smarter.

internal/builder/kraftfile.go
```diff
@@ -58,22 +58,26 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
...
 			Type:   rom.Source.Type,
 			// Pad the file to page-size alignment. This is required by the platform
 			// which rejects ROM files that are not page-aligned.
 			Pad: 4096,
 		}
-		if romOpt.Type == "" {
-			typ, err := DetectSourceType(romPath)
-			if err != nil {
-				return BuildOpts{}, fmt.Errorf("detecting rom type for %q: %w", romPath, err)
+		if romOpt.Type == kraftfile.SourceTypeOCI || isOCIRef(rom.Source.Path) {
```
jedevc: Hm, it feels weird that we have to special case this...

Why do we need to do this? Is it just for the path joining?

If so, this is a good point - we shouldn't do the joining, instead we should propagate the kraftfile path separately, maybe in `BuildOpts` as `RootPath`?

internal/builder/rootfs.go
```diff
@@ -60,12 +64,30 @@ func buildImageConfig(opts BuildOpts) ocispec.ImageConfig {
...
+	uri, err := imagespec.ParseURI(path)
+	if err != nil {
+		return false
+	}
+	switch uri.Scheme {
+	case imagespec.URISchemeOCI, imagespec.URISchemeOCILayout, imagespec.URISchemeOCIArchive:
+		return true
+	}
+	return false
+}
```
jedevc: Hmmmmm. I'm not sure about this.

Let's not mix and match the URI types into the kraftfile. We have the `type` field to distinguish in the Kraftfile.

We can *just* have `oci`, which is just registry images for now - but if we wanted to support more, we'd need to add `kraftfile.SourceTypeOCIArchive`, etc.

internal/builder/rootfs.go
```diff
@@ -303,6 +327,215 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+		if err := applyLayer(ctx, img.Provider, layer, dir); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// manifestsForPlatform reads the descriptor as an index or manifest and returns
+// the manifests matching the given platform.
+func manifestsForPlatform(ctx context.Context, provider content.Provider, desc ocispec.Descriptor, platform platforms.MatchComparer) ([]ocispec.Manifest, error) {
```
jedevc: I think you want https://github.com/unikraft-cloud/x/blob/9e6e9e53c51b8ee041c290c20f4629639db63c45/image-spec/util.go#L21-L23

We should make it public, and call it `util.Manifests`.

internal/builder/rootfs.go
craciunoiuc: uh right, we need to use smth like a tar reader/writer

internal/builder/rootfs.go
craciunoiuc: True

When building this they felt like synonyms, but yeah I can remove them (and the function)

internal/builder/kraftfile.go
craciunoiuc: Yes, exactly

I'll propagate the path

internal/builder/rootfs.go
craciunoiuc: Nice, yes, exactly

internal/builder/rootfs_oci_test.go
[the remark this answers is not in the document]
jedevc: we don't compress the blobs. so it's fine.

internal/builder/rootfs.go
```diff
@@ -470,10 +480,23 @@ func buildRootfsOCI(ctx context.Context, opts BuildOpts) (_ []*imagespec.Image,
...
-func flattenImageLayers(ctx context.Context, img *imagespec.Image, platform ocispec.Platform, dir string) error {
+// Whiteout markers as defined by the OCI image layer specification.
+// https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts
+const (
+	whiteoutPrefix    = ".wh."
+	whiteoutOpaqueDir = whiteoutPrefix + whiteoutPrefix + ".opq"
+)
+
+// flattenImageLayers resolves the manifest layers of img and merges them, in
+// order (base first), into a single uncompressed tarball written to dst. This
```
craciunoiuc: Damn, bro cooked and burnt food

internal/builder/kraftfile.go
craciunoiuc: adding `RootPath`

internal/builder/rootfs.go
craciunoiuc: Ok, so I'll try to unpack the oci tar into a regular tar and pass that forward

Just have to convince the robot to behave

jedevc: Remember we have to handle whiteout files and overlayfs file semantics — you can't just squash em all together.

One option for this — just don't try and handle this, and just sent the entire request to buildkit, and unpack it to a tar.

craciunoiuc: Yeah, it actually did that, to my horror

Was saying I need to dig out a better way (like the buildkit offload yeah)

craciunoiuc: Ok so I left it to dig and it was close to giving up. But apparently it liked the buildkit idea

I didn't read it yet or clean it up, but the general structure looks fine and using buildkit for this is conceptually sound

So tomorrow I'm cleaning up and undrafting this PR after testing again

internal/builder/rootfs.go
```diff
@@ -105,10 +105,37 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
...
+	}
+
+	if root != "" {
+		fsOpts.Path = filepath.Join(root, fsOpts.Path)
+	}
+
+	switch {
+	case fsOpts.Type != "":
+	case fsOpts.Dockerfile != "":
+		fsOpts.Type = kraftfile.SourceTypeDockerfile
```
craciunoiuc: smth is weird here

internal/builder/rootfs.go
```diff
@@ -303,6 +336,151 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+		if err := flattenImageLayers(ctx, uri, src, p, layers); err != nil {
+			return nil, fmt.Errorf("unpacking rootfs image %q: %w", opts.Rootfs.Path, err)
+		}
+
+		srcFS, err := buildfs.TarballFS(layers)
+		if err != nil {
+			return nil, fmt.Errorf("could not open flattened rootfs image as filesystem: %w", err)
+		}
+
+		f, err := os.CreateTemp("", "unikraft-rootfs-*."+string(opts.Rootfs.Format))
```
craciunoiuc: We need to start to think if we want to start moving these image build operations into the local directory

Tmpfs gets really crowded when we chuck so many image artifacts into it

internal/builder/rootfs.go
```diff
@@ -35,31 +32,40 @@ import (
...
 
+// buildImageConfig constructs a minimal OCI image config from build options.
+// Used when there is no source image config to build on top of.
+func buildImageConfig(opts BuildOpts) ocispec.ImageConfig {
+	return applyConfigOverrides(ocispec.ImageConfig{}, opts)
+}
+
+// errSourceNotExist is returned by DetectSourceType when the source path is not
+// present on disk.
+var errSourceNotExist = errors.New("rootfs path does not exist")
```
jedevc: `fs.ErrFileNotFound` or whatever it's called 🤔

internal/builder/rootfs.go
```diff
@@ -105,10 +111,55 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
...
+func detectPackagedFormat(path string) (kraftfile.FsType, error) {
+	switch {
+	case gocpio.IsValidPath(path):
+		return kraftfile.FsTypeCpio, nil
+	case goerofs.IsValidPath(path):
+		return kraftfile.FsTypeErofs, nil
+	default:
+		return "", fmt.Errorf("could not detect rootfs format of %q", path)
+	}
+}
```
jedevc, line 122: We can use this in `DetectSourceType` too.

internal/builder/kraftfile.go
```diff
@@ -58,43 +58,25 @@ func KraftfileToBuildOpts(dir string, kf *kraftfile.Kraftfile) (BuildOpts, error
...
 	if kf.Rootfs != nil {
 		if kf.Rootfs.Source == nil || kf.Rootfs.Source.Path == "" {
 			return BuildOpts{}, fmt.Errorf("rootfs entry is missing a source path")
 		}
-		opts.Rootfs.Path = filepath.Join(dir, kf.Rootfs.Source.Path)
+		opts.Rootfs.Path = kf.Rootfs.Source.Path
 		opts.Rootfs.Format = kf.Rootfs.Format
 		opts.Rootfs.Type = kf.Rootfs.Source.Type
 		opts.Rootfs.Dockerfile = kf.Rootfs.Source.Dockerfile
-		if opts.Rootfs.Dockerfile != "" && opts.Rootfs.Type == "" {
```
jedevc, line 89: I think it makes sense to call `resolveSource` here? Why do we defer it?

The opts are invalid, we should make them valid *here*.

internal/builder/rootfs.go
```diff
@@ -105,10 +111,55 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
...
+
+// resolveSource resolves the source of fsOpts against root and fills in the
+// source type when it was not requested explicitly.
+func resolveSource(root string, fsOpts *FSOpts) error {
+	if fsOpts.Type == kraftfile.SourceTypeOCI {
+		return nil
+	}
+
+	if root != "" {
+		fsOpts.Path = filepath.Join(root, fsOpts.Path)
```
jedevc, line 135: This feels a bit icky. Instead, the caller should resolve these, it should determine these, instead of magically combining them here.

internal/builder/build.go
```diff
@@ -13,16 +13,17 @@ import (
...
 	"unikraft.com/x/kraftfile"
 	"unikraft.com/x/log"
 
 	"unikraft.com/cli/internal/builder/buildflags"
 )
 
 type BuildOpts struct {
-	Rootfs FSOpts
-	Roms   []FSOpts
+	RootPath string
```
jedevc: Add a comment explaining what this is for.

internal/builder/rootfs.go
```diff
@@ -498,20 +681,26 @@ func packageFS(ctx context.Context, format kraftfile.FsType, destFS *os.File, sr
...
 	}
 
-	if opts.Pad > 0 {
-		pos, err := destFS.Seek(0, io.SeekEnd)
+	return padFile(destFS, opts.Pad)
+}
+
+// padFile pads f up to a multiple of pad bytes and syncs it. A pad of zero only
+// syncs.
+func padFile(f *os.File, pad int64) error {
```
jedevc: Split out padding + syncing. They shouldn't be in the same func.

internal/builder/rootfs_buildkit.go
```diff
@@ -0,0 +1,153 @@
...
+	if _, err := dst.Seek(0, io.SeekStart); err != nil {
+		return fmt.Errorf("could not rewind tarball: %w", err)
+	}
+
+	return nil
+}
+
+// flattenImageLayers writes the flattened filesystem of a regular OCI image to
+// dst as an uncompressed tarball, using BuildKit to do the flattening.
+func flattenImageLayers(ctx context.Context, opts BuildOpts, src *imagespec.Image, uri *imagespec.URI, dst *os.File) error {
```
jedevc: This should go in `rootfs.go` I think. Ideally inline, it's only used in one place.

internal/builder/rootfs_buildkit.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	// Resolve against the digest discovery already settled on, rather than
+	// letting BuildKit resolve the tag a second time.
+	ref := uri.Path
+	if src.Descriptor.Digest != "" && !strings.Contains(ref, "@") {
+		ref += "@" + src.Descriptor.Digest.String()
+	}
+
+	imagePlatform := getPlatforms([]ocispec.Platform{src.Image.Platform})[0].Platform
```
jedevc: Split this to `getPlatform`, then this reads a lot better.

internal/builder/rootfs_buildkit.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return nil
+}
+
+// nopWriteCloser adapts a writer we do not own into an io.WriteCloser.
+type nopWriteCloser struct {
+	io.Writer
+}
+
+func (nopWriteCloser) Close() error { return nil }
```
jedevc: Make this an `x/io` helper.

internal/cmd/volumes.go
```diff
@@ -735,14 +735,9 @@ func (c *VolumeImportCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *
 		return fmt.Errorf("unexpected resource type %T", resources[0])
 	}
 
-	sourceType, err := builder.DetectSourceType(c.Source)
-	if err != nil {
-		return fmt.Errorf("detecting source type for %q: %w", c.Source, err)
-	}
```
jedevc, line 754: :eyes: this feels weird to remove.

I really liked the previous property of `opts` passed in that once we passed them in *they were ready to go*. Not anything left to be resolved, or defaults.

internal/builder/rootfs_oci_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Mode: 0o644,
+		Size: int64(len(data)),
+	}))
+	_, err := io.Copy(tw, bytes.NewReader(data))
+	require.NoError(t, err)
+}
+
+func digestHex(dgst digest.Digest) string {
+	return strings.TrimPrefix(string(dgst), "sha256:")
+}
```
jedevc: replace with `dgst.Encoded()`.

internal/builder/rootfs_test.go
```diff
@@ -736,3 +769,245 @@ func runBuildRootfs(t *testing.T, opts BuildOpts) []*imagespec.Image {
...
+}
+
+// runBuildRootfsOCIBuildkit calls BuildRootfs with BuildKit wired up, which
+// flattening the layers of a regular OCI image requires. Since those are read
+// from a registry, this needs network access too.
+func runBuildRootfsOCIBuildkit(t *testing.T, opts BuildOpts) []*imagespec.Image {
+	t.Helper()
+
+	return runBuildRootfsIntegration(t, rootfsIntegrationContext(t), opts)
+}
```
jedevc: What are these helpers 😓

internal/builder/rootfs.go
craciunoiuc: I need to double check, maybe that error is swallowed or or needs to be matched further down aka unpacked

internal/builder/build.go
craciunoiuc: Whoops, and I removed it intentionally :dizzy:

internal/builder/rootfs_buildkit.go
craciunoiuc: It is a pretty big function tbh. I already inlined like 3-4 smaller ones into it

I was afraid it would make the original actually too big (can't remember exactly but it might be 150+ lines)

I can merge them if you think we can handle it though ;d

internal/cmd/volumes.go
craciunoiuc: This I need to recheck why

internal/builder/rootfs_test.go
craciunoiuc: don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests don't look at the tests

jedevc: I'm gonna look at the tests 😞

internal/builder/build.go
craciunoiuc: No longer needed :flushed:

craciunoiuc:
Code changed quite a bit cause there were 80 commits merged since I last touched this.

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
-		authprovider.NewDockerAuthProvider(authprovider.DockerAuthProviderConfig{
-			AuthConfigProvider: images.LoadBuildkitAuthConfig(dockerConfig, profile),
-		}),
+	uri, err := imagespec.ParseURIDefault(opts.Rootfs.Path)
+	if err != nil {
+		return nil, fmt.Errorf("parsing rootfs image reference %q: %w", opts.Rootfs.Path, err)
+	}
+
+	// A unikraft image advertises the unikraft target platform, whereas a
+	// regular OCI image advertises a real one (linux/amd64 and friends).
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+		return nil, fmt.Errorf("pulling rootfs image %q: %w", opts.Rootfs.Path, err)
+	}
+	defer func() {
+		for _, img := range loaded {
+			_ = img.Close()
+		}
+	}()
+
+	// Index the loaded images by normalized platform so we can match the
+	// requested platforms against them.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+
+	var imgs []*imagespec.Image
+	for i, p := range opts.Platform {
+		src := byPlatform[platforms.Format(platforms.Normalize(p))]
+		if src == nil {
+			src = byPlatform[platforms.Format(imagePlatforms[i].Platform)]
+		}
+		if src == nil {
+			// Fall back to the single loaded image (e.g. for images without a
+			// platform-specific manifest).
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+			if len(loaded) == 1 {
+				src = loaded[0]
+			} else {
+				return nil, fmt.Errorf("rootfs image %q does not contain platform %q", opts.Rootfs.Path, platforms.Format(p))
+			}
+		}
+
+		// Build on top of the source image's config, so that e.g. the Cmd and Env
+		// of a regular OCI image survive, then layer our overrides on top.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+		cfg := buildImageConfig(opts)
+		if src.Image != nil {
+			cfg = applyConfigOverrides(src.Image.Config, opts)
+		}
+
+		// Unikraft image with a dedicated initrd (rootfs) component: use it
+		// directly.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+			}
+			if _, err := io.Copy(f, rc); err != nil {
+				rc.Close()
+				return nil, fmt.Errorf("reading rootfs layer: %w", err)
+			}
+			if err := rc.Close(); err != nil {
+				return nil, fmt.Errorf("closing rootfs layer: %w", err)
+			}
+			// detectPackagedFormat reads by path, so the bytes have to be on
+			// disk before it runs.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+			if err := syncFile(f); err != nil {
+				return nil, err
+			}
+
+			// The initrd is not repackaged, so the requested format has to match
+			// the one the image already carries.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+			format, err := detectPackagedFormat(f.Name())
+			if err != nil {
+				return nil, fmt.Errorf("inspecting initrd of rootfs image %q: %w", opts.Rootfs.Path, err)
+			}
+			if opts.Rootfs.Format != "" && opts.Rootfs.Format != format {
+				return nil, fmt.Errorf("unsupported rootfs format mismatch: source is %s but requested format is %s", format, opts.Rootfs.Format)
+			}
+
+			// Padding is format-agnostic, so unlike the rest of the output
+			// options it does apply to a pass-through. ROMs rely on it.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+			imgs = append(imgs, imagespec.NewImage(
+				imagespec.WithImageConfig(cfg),
+				imagespec.WithPlatform(p),
+				imagespec.WithInitrd(imagespec.NewTempOSFile(f)),
+			))
+			continue
+		}
+
+		// Regular OCI image: flatten its layers into a single tarball and
+		// package that into the requested rootfs format.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+		format := cmp.Or(opts.Rootfs.Format, DefaultRootfsFormat(opts.Platform))
+
+		srcFS, ok := flattened[src.Descriptor.Digest.String()]
+		if !ok {
+			layers, err := os.CreateTemp("", "unikraft-buildkit-*.tar")
+			if err != nil {
+				return nil, fmt.Errorf("could not create temporary file: %w", err)
+			}
+			// Held open for the lifetime of the call: srcFS reads through it.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -324,18 +373,239 @@ func buildRootfsTarball(ctx context.Context, opts BuildOpts) (_ []*imagespec.Ima
...
+	if uri.Scheme != imagespec.URISchemeOCI {
+		return fmt.Errorf("rootfs image %q must be a registry reference, %q is not supported", uri.Path, uri.Scheme)
+	}
+
+	if src.Image == nil {
+		return fmt.Errorf("rootfs image %q has no config to take a platform from", uri.Path)
+	}
+
+	// Resolve against the digest discovery already settled on, rather than
+	// letting BuildKit resolve the tag a second time.
```
craciunoiuc: Remove comment

internal/builder/rootfs.go
```diff
@@ -524,22 +782,103 @@ func packageFS(ctx context.Context, format kraftfile.FsType, destFS *os.File, sr
...
+
+// solveToTar runs build against BuildKit, exporting an uncompressed tarball of
+// the result to dst, and waits for the progress writer to drain.
+func solveToTar(ctx context.Context, dst *os.File, solveOpt client.SolveOpt, build gateway.BuildFunc) error {
+	solveOpt.Ref = identity.NewID()
+	solveOpt.Exports = []client.ExportEntry{{
+		Type: client.ExporterTar,
+		Output: func(map[string]string) (io.WriteCloser, error) {
+			// We own dst, so do not let BuildKit close it out from under the
+			// caller.
```
craciunoiuc: Remove comment

craciunoiuc (commented):
Comment removal comments

internal/builder/rootfs.go
```diff
@@ -106,6 +109,59 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
...
+		fsOpts.Type = typ
+	}
+
+	return nil
+}
+
+// defaultRomFormat fills in the ROM default format. An OCI source carries its
+// own format, so leave it unset and let the source dictate it.
+func defaultRomFormat(format kraftfile.FsType, typ kraftfile.SourceType) kraftfile.FsType {
+	if format != "" || typ == kraftfile.SourceTypeOCI {
```
jedevc: bug: A *regular* OCI image has no format of its own to dictate, so it falls out of here with an empty format and lands on `cmp.Or(opts.Rootfs.Format, DefaultRootfsFormat(opts.Platform))` — cpio, unless the runtime advertises erofs. Same Kraftfile, two roms, neither with a `format:`, and they come out in different formats.

Only the unikraft-image branch actually carries a format worth deferring to. Could we keep erofs here and relax the mismatch check in the initrd branch instead?

```
roms:
  - source: {path: ./romdir}
  - source: {type: oci, path: index.docker.io/library/hello-world:latest}

$ xxd -s 1024 -l 4 rom-from-directory   # erofs superblock magic
00000400: e2e1 f5e0                                ....
$ xxd -l 6 rom-from-oci-image           # cpio "newc" magic
00000000: 3037 3037 3031                           070701
```

internal/builder/rootfs.go
```diff
@@ -39,28 +40,36 @@ import (
...
 		cfg.Cmd = opts.Cmd
 	}
 	if opts.Env != nil {
-		env := make([]string, 0, len(opts.Env))
+		env := make([]string, 0, len(opts.Env)+len(cfg.Env))
 		for _, kv := range opts.Env {
 			env = append(env, fmt.Sprintf("%s=%s", kv.Key, kv.Value))
 		}
-		cfg.Env = env
+		cfg.Env = append(env, cfg.Env...)
```
jedevc, line 61: bug: "Env is prepended so the caller's values take precedence".

Oops. This might be the wrong way round, pulled from: https://github.com/unikraft-cloud/cli/blob/bd663f30592f45447ae61d9bbb96399568e1ad86/internal/builder/rootfs.go#L459-L465

Because to get precedence for an env list, you actually want to be the *last* setter.

internal/builder/rootfs_oci_test.go
```diff
@@ -0,0 +1,216 @@
...
+	layerDesc, layerBlob, diffID := tarGzipLayer(t, srcDir)
+	return writeOCIArchive(t, "regular-image.tar",
+		[]ocispec.Descriptor{layerDesc}, [][]byte{layerBlob}, []digest.Digest{diffID})
+}
+
+// writeOCIArchive wraps the given layers in a regular OCI image and saves it as
+// an OCI archive tarball named name. diffIDs are the digests of the uncompressed
+// layers, which is what the config records, as opposed to the descriptors'
+// digests of the compressed blobs. It returns the path to the tarball.
+func writeOCIArchive(t *testing.T, name string, layerDescs []ocispec.Descriptor, layerBlobs [][]byte, diffIDs []digest.Digest) string {
```
jedevc, line 72: nit: A `name` param, three parallel slices and a `require.Len` to keep them aligned?

Should these maybe be in a struct? Or folded into `writeRegularOCIArchive`?

internal/builder/rootfs_oci_test.go
```diff
@@ -0,0 +1,216 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package builder
```
jedevc, line 6: nit: There are no tests in `rootfs_oci_test.go` — the OCI tests are all over in `rootfs_test.go`. Move them here?

internal/builder/rootfs_test.go
```diff
@@ -837,3 +928,237 @@ func runBuildRootfs(t *testing.T, opts BuildOpts) []*imagespec.Image {
...
+	})
+	require.ErrorContains(t, err, "must be a registry reference")
+}
+
+// TestRootfsOCIRegularImageIntegration reads a regular OCI image (plain layers,
+// no unikraft components) from a registry and verifies that BuildKit flattens
+// the layers and that the result is re-packaged into the requested rootfs
+// format.
+func TestRootfsOCIRegularImageIntegration(t *testing.T) {
+	const ref = "index.docker.io/library/hello-world:latest"
```
jedevc, line 1082: Four integration tests now pull `hello-world` from Docker Hub anonymously, which is rate limited per-IP and will flake in CI eventually. 

suggestion: Maybe we should build an image using `SharedImage` for this? Then it's on `index.unikraft.io` with no rate limits.

internal/builder/rootfs.go
```diff
@@ -128,7 +184,7 @@ func BuildRoms(ctx context.Context, opts BuildOpts) (_ [][]imagespec.File, rerr
 			Rootfs: FSOpts{
 				Path:   rom.Path,
 				Type:   rom.Type,
-				Format: cmp.Or(rom.Format, kraftfile.FsTypeErofs),
+				Format: defaultRomFormat(rom.Format, rom.Type),
```
jedevc, line 187: Hm, why has this changed?

internal/builder/rootfs_oci_test.go
craciunoiuc: hmm yeah, I think I wanted to remove this but forgor? So I guess I should move it here?

internal/builder/rootfs.go
craciunoiuc: good question, can't remember

internal/builder/rootfs_oci_test.go
craciunoiuc: :thinking: :thinking: :thinking:

internal/builder/rootfs_test.go
craciunoiuc: Actually, it's a bit weird, we would need to use an image which is standard, so not our usual projects, and it needs to be public also

Maybe something in `kraftkit.sh`? :thinking: 

image


There's this, only 3MB big, maybe ok? `kraftkit.sh/xen:4.19`

It's stable, not changing any time soon

craciunoiuc: alternatively `4.18` which is unchanged since 2024

internal/builder/rootfs.go
craciunoiuc: So the reason is basically because you want to match the type of the rom if not specified:

the function itself is:
```
	if format != "" || typ == kraftfile.SourceTypeOCI {
		return format
	}
	return kraftfile.FsTypeErofs
```

In the old case of `cmp.Or`, if the rom did not have a format set it would assume `erofs`, even if the rom inputted was of type `cpio`.

jedevc: 🤔 hm. I thought that only EROFS ROMs auto-mounted.

internal/builder/rootfs_test.go
jedevc: Why can't we mirror it temporarily? And just delete it? Why does it need to be standard?

I ideally don't want dependencies on ye olde kraftkit stuff.

craciunoiuc: (It's actually a build of the go bindings for using the xen `xl` library :nerd_face: ) 

Ok, but then where would that sit? A new project? I currently know only of the `kraftkit.sh` where we put regular docker images (note: there are not the user/ukc projects)

Should we create a new one `cli`? or?

internal/builder/rootfs.go
craciunoiuc: yeah

still, with the old code, if we were to give a `cpio` file to the builder with an `erofs` format you would get:
```
unsupported rootfs format mismatch: source is cpio but requested format is erofs
```

I think I have a cleaner fix, rechecking

craciunoiuc: got annoyed, I'm removing those changes as they are not related to this work at all

internal/builder/rootfs_test.go
craciunoiuc: Ok, so it seems that we can't use a shared-image because this is not an e2e image, it's the thing underneath the tests, meaning we can't "inject" the shared image.

That being said, I will push the fully copy of the `hello-world` image into the `cli-test` project as discussed.

In `index.unikraft.io/cli-test/hello-world:test-fixture`

Currently waiting out my rate limit

craciunoiuc:
Something has happened.


Like the `shared image`, created a `mirror image` that does what we discussed: pulling once and pulling to our registry, and then using that.

Placed it in a different commit.

Issue #385, opened 2026-06-24
Allow graphing arbitary field data


jedevc:
We should add `unikraft <resource> graph`. This new command, available on all resources would allow graphing fields of that resource over time. You would be able to observe those properties as they changed.

We should probably use [https://github.com/NimbleMarkets/ntcharts](<https://github.com/NimbleMarkets/ntcharts>).

This would generate a line graph which would keep updating as more information arrived of the memory usage of `instance-foo`:

```
$ unikraft instance graph <instance-foo> -f metrics.rss

 10│      ⣀⠤⠒⠉⠒⠤⡀                        
   │    ⡠⠊      ⠈⠢⡀                      
  5│  ⡠⠊          ⠈⠢⡀                    
   │⡠⠊              ⠈⠑⢄                 ⢀
  0│                   ⠑⡄              ⡔⠁
   │                    ⠈⠢⡀          ⡠⠊  
 -5│                      ⠈⠢⡀      ⡠⠊    
   │                        ⠈⠑⠢⢄⡠⠔⠊      
-10└─────────────────────────────────────
   '24 03/27   03/31   04/03   04/05     
```

We could also show this against the total amount of memory available (as a solid bar):

```
$ unikraft instance graph <instance-foo> -f resources.memory -f metrics.rss
...
```

What about showing it as a percentage? We need some processing:

```
$ unikraft instance graph <instance-foo> -f metrics.rss/resources.memory
...
```

We could even allow plotting two instances against each other, one of `instance-foo` and another of `instance-bar` together:

```
$ unikraft instance graph <instance-foo> <instance-bar> -f metrics.rss
...
```

We could show CPU usage by computing a time delta (or some other function):

```
$ unikraft instance graph <instance-foo> -f derivative(cpu_time)
...
```

> [!NOTE]
>
> `-f metrics.rss/resources.memory` and `-f derivative(cpu_time)` do not exist yet.
>
> There are kinda interesting though... and feel related to unikraft/cli#373. They're almost like "virtual fields"? Not really in the data model, but can be computed from it.

We should also support non-integer data. Some ideas:

* `bool`s could be graphed as a square signal
* `enum`s could be graphed as a histogram (when we have multiple inputs)

ngoldack:
Great idea!

jedevc:
Hey @ngoldack did you manage to take a look at this?

ngoldack:
Hi @jedevc , of cause!

[...]

PR #417, opened 2026-07-10
Add graph command for instance resource usage

ngoldack:
Introduces `unikraft instance graph`: renders a live-updating terminal line chart (via [ntcharts](<https://github.com/NimbleMarkets/ntcharts>) v2) of an instance's metrics. Part of unikraft-cloud/cli#385.

* `-f` picks what to graph: `metrics.rss`, `metrics.cpu`, `resources.memory`, `resources.vcpus` (defaults to `metrics.rss`)
* Axis + header show nicely formatted values (bytes, durations, etc.)
* Built on top of the existing `watcher.WatchOutput` polling/TTY infra, no new TUI plumbing

Draft since it's an MVP — multi-field, multi-instance, and computed fields (from the issue discussion) are follow-ups.

@jedevc let me know what you think about that


internal/cmd/instances.go
```diff
@@ -1426,6 +1436,234 @@ func (cmd *InstancesLogsCmd) Run(ctx context.Context, stdio config.Stdio) error
...
+
+var instanceGraphFields = map[string]instanceGraphField{
+	"metrics.rss": {
+		Label:      "RSS",
+		Source:     instanceGraphSourceMetrics,
+		AxisFormat: instanceGraphAxisFormatBytes,
+		FromMetrics: func(m platform.GetInstancesMetricsResponseInstanceMetrics) float64 {
+			return float64(m.RssBytes)
+		},
+	},
```
jedevc, line 1522: I think we could make this generic, and not just on `instance`, by implementing something like https://github.com/unikraft-cloud/cli/blob/adb11b48ffe857d9e4793fad40af75716b8284f1/internal/resource/cmd/cmd.go#L53-L55

Then we operate *just* on a generic resource in there, then this should simplify a lot of the boilerplate here, ideally no changes to `instances.go` is needed beyond embedding a new `GraphableResourceCmd`.

internal/cmd/instances.go
```diff
@@ -1426,6 +1436,234 @@ func (cmd *InstancesLogsCmd) Run(ctx context.Context, stdio config.Stdio) error
...
+		width = 80
+	}
+	axisFormatter := instanceGraphAxisFormatter(field.AxisFormat)
+	chart := timeserieslinechart.New(
+		width, instanceGraphHeight,
+		timeserieslinechart.WithYLabelFormatter(axisFormatter),
+	)
+	chart.SetStyle(lipgloss.NewStyle().Foreground(colors.Primary))
+
+	render := func(out io.Writer) error {
```
jedevc, line 1604: I like how this uses the existing watcher - I wonder if we should make the `render` func take a width/height though, since they're very efficient to get with bubbletea (we can catch events when they change, instead of constantly repolling the size).

internal/cmd/instances.go
```diff
@@ -1426,6 +1436,234 @@ func (cmd *InstancesLogsCmd) Run(ctx context.Context, stdio config.Stdio) error
...
+	width := xio.TermWidth(stdio.Stdout)
+	if width <= 0 {
+		width = 80
+	}
+	axisFormatter := instanceGraphAxisFormatter(field.AxisFormat)
+	chart := timeserieslinechart.New(
+		width, instanceGraphHeight,
+		timeserieslinechart.WithYLabelFormatter(axisFormatter),
+	)
+	chart.SetStyle(lipgloss.NewStyle().Foreground(colors.Primary))
```
jedevc, line 1602: I like the look of this!
Image

The x-axis doesn't look quite right though.

I wonder if we can also maybe scale the y-axis so it shows in more reasonable units? Instead of in increments of 131ms 😆

internal/cmd/instances.go
```diff
@@ -1426,6 +1436,234 @@ func (cmd *InstancesLogsCmd) Run(ctx context.Context, stdio config.Stdio) error
...
+
+		value, err := pollInstanceGraphValue(ctx, g, refs, target, field)
+		if err != nil {
+			return err
+		}
+		chart.Push(timeserieslinechart.TimePoint{Time: time.Now(), Value: value})
+		chart.DrawBraille()
+
+		fmt.Fprintf(out, "Instance: %s  Field: %s (%s): %s\n", target, fieldPath, field.Label, axisFormatter(0, value))
+		fmt.Fprintln(out, chart.View())
```
jedevc, line 1617: It might be neat to support zooming and panning on the graph 🤔 

And maybe even pausing new data coming in (but still pulling data in) with `space`.

Just some ideas!

jedevc:
Left some initial impressions! It looks neat, already excited to see where it goes :tada:

PR #418, opened 2026-07-11
feat(filters): Support comparison operators

### LINKED ISSUE #374: Support `--filter`s with `<` and `>`
To be able to things like `--filter size>50mb`.

The `filters` code all lives in https://github.com/unikraft-cloud/x/tree/prod-staging/filters.

Omkar-Ugal:
adds `>`, `<`, `>=`, `<=` support in `--filter` (e.g `--filter "resources.memory>128mib"`).

Tested manually against live instances, volumes, and services - memory, vcpus, size, limits all work with all four operators.

Closes unikraft-cloud/cli#374.  <br>Requires [https://github.com/unikraft-cloud/x/pull/307](<https://github.com/unikraft-cloud/x/pull/307>).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 	unikraft.com/cloud/sdk v0.3.1-0.20260730110715-92b5bae5afbe
 	unikraft.com/x/colors v0.0.0-20260710212606-b68ba398853d
-	unikraft.com/x/filters v0.0.0-20260416164455-ec39ae908f3f
+	unikraft.com/x/filters v0.0.0-20260804153219-d1b47a40e047
 	unikraft.com/x/fingerprint v0.0.0-20260126094137-ab6e717e5679
 	unikraft.com/x/guesstermwidth v0.0.0-20260708141520-0cbb0a92056b
 	unikraft.com/x/iata v0.0.0-20260713183529-fd34645687a0
@@ -194,5 +194,3 @@ require (
 )
 
 tool github.com/caarlos0/svu/v3
-
-replace unikraft.com/x/filters => ../x/filters
--- a/go.mod
+++ b/go.mod
@@ -194,3 +194,5 @@ require (
 )
 
 tool github.com/caarlos0/svu/v3
+
+replace unikraft.com/x/filters => ../x/filters
--- a/internal/resource/cmd/cmd.go
+++ b/internal/resource/cmd/cmd.go
@@ -14,13 +14,10 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"reflect"
 	"slices"
-	"strconv"
 	"strings"
 	"time"
 
-	"github.com/charmbracelet/x/ansi"
 	"unikraft.com/x/filters"
 	"unikraft.com/x/kingkong"
 	"unikraft.com/x/log"
@@ -29,7 +26,6 @@ import (
 	"unikraft.com/cli/internal/multimetro"
 	"unikraft.com/cli/internal/resource"
 	"unikraft.com/cli/internal/resource/patch"
-	"unikraft.com/cli/internal/resource/value"
 	"unikraft.com/cli/internal/tui/watcher"
 	xkong "unikraft.com/cli/internal/x/kong"
 	"unikraft.com/cloud/sdk/platform/group"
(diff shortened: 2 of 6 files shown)


internal/resource/cmd/cmd.go
```diff
@@ -413,56 +414,144 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
...
-						idx, err := strconv.Atoi(key[len(key)-1])
-						if err == nil && idx >= 0 && idx < len(slice) {
-							return slice[idx], nil, true
-						}
+
+// fieldAdaptor implements filters.Adaptor over a resource's fields, using
+// resource.GetFieldByPath to resolve field paths, and exposing raw typed
+// values so comparison operators (>, <, >=, <=) can work against native
+// types (e.g. memory>128mib, created_at>2026-01-01).
+type fieldAdaptor struct {
```
jedevc: Let's move all this out into `filter.go` since it's now more complex.

internal/resource/cmd/cmd.go
```diff
@@ -413,56 +414,144 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
 
 // newFieldAdaptor creates a filters.Adaptor that can traverse resource fields.
 // It handles both structured fields (with subfields) and slice values (like []string tags).
```
jedevc: Dangling comment.

internal/resource/cmd/cmd.go
jedevc, line 1: We need some tests for this logic too. See `cmd_test.go` and look for existing filter tests we could extend.

internal/resource/cmd/cmd.go
```diff
@@ -413,56 +414,144 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
...
+	}
+	return nil, names, "", false, true
+}
+
+func (a *fieldAdaptor) Select(key []string) (filters.Adaptor, bool) {
+	full := a.fullPath(key)
+	if _, _, _, _, ok := a.resolve(full); !ok {
+		return nil, false
+	}
+	return &fieldAdaptor{fields: a.fields, prefix: full}, true
```
jedevc: Instead of storing the prefix, couldn't we store the whole `field`?

internal/resource/cmd/cmd.go
```diff
@@ -413,56 +414,144 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
...
+}
+
+func (a *fieldAdaptor) fullPath(key []string) []string {
+	full := make([]string, 0, len(a.prefix)+len(key))
+	full = append(full, a.prefix...)
+	full = append(full, key...)
+	return full
+}
+
+func (a *fieldAdaptor) resolve(path []string) (field *resource.Field, entries []string, sliceVal string, sliceOK bool, ok bool) {
```
jedevc: We should return a whole `fieldAdaptor` here I think. This would simplify the code, and avoid so many return values.

Then we also don't need to keep calling `resolve` over and over, and means we should be able to avoid storing the `prefix`.

internal/resource/cmd/cmd.go
```diff
@@ -413,56 +414,144 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
...
+		return 0, false
+	}
+	parsed, err := value.ParseNew([]string{other}, field.Value)
+	if err != nil {
+		return 0, false
+	}
+	return value.Compare(field.Value, parsed), true
+}
+
+func hasOrdering(v any) bool {
```
jedevc: Hm, I think this should probably be opt-in not opt-out.

E.g. currently this would include arrays/structs/etc. Which aren't ordered.

internal/resource/cmd/cmd.go
```diff
@@ -413,56 +414,144 @@ func filterResources(ctx context.Context, resources []resource.Resource, filter
...
+		return 0, false
+	}
+	if !hasOrdering(field.Value) {
+		return 0, false
+	}
+	parsed, err := value.ParseNew([]string{other}, field.Value)
+	if err != nil {
+		return 0, false
+	}
+	return value.Compare(field.Value, parsed), true
```
jedevc: Hm, worth checking to see how this handles comparing a float field to an integer result - e.g. `1.2<5`. Or something like that (something to check works in the tests - it should work as is).

jedevc (commented):
Looooks neat! Sorry for the delay again (I was on holiday 🌴).

internal/resource/cmd/filter.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		rv = rv.Elem()
+	}
+	switch rv.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+		reflect.Float32, reflect.Float64,
+		reflect.Bool:
+		return true
+	}
```
jedevc: While strings aren't necessarily orderable, they can be compared to be equal.

internal/resource/cmd/cmd_test.go
```diff
@@ -2066,3 +2068,83 @@ func TestDeleteBulk(t *testing.T) {
...
+		err := cmd.Run(ctx, testStdio(&out), sandbox)
+		return out.String(), err
+	}
+
+	t.Run("greater", func(t *testing.T) {
+		output, err := runFilter(t, "settings.foo>7")
+		require.NoError(t, err)
+		assert.Contains(t, output, "test1")
+		assert.NotContains(t, output, "test2")
+	})
```
jedevc, line 2105: We should have an `equal` test.

internal/resource/cmd/filter.go
```diff
@@ -0,0 +1,160 @@
...
+	"unikraft.com/cli/internal/resource/value"
+	"unikraft.com/x/filters"
+)
+
+type fieldAdaptor struct {
+	siblings []resource.Field 
+	field    *resource.Field  
+	entries  []string         
+	sliceVal string           
+	sliceOK  bool
```
jedevc: Consolidate to `*string`.

internal/resource/cmd/filter.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/charmbracelet/x/ansi"
+
+	"unikraft.com/cli/internal/resource"
+	"unikraft.com/cli/internal/resource/value"
+	"unikraft.com/x/filters"
+)
+
+type fieldAdaptor struct {
+	siblings []resource.Field
```
jedevc: Technically, this should be `children`.

go.mod
```diff
@@ -194,3 +194,11 @@ require (
 )
 
 tool github.com/caarlos0/svu/v3
+
+// HACK: hot-fix for https://github.com/charmbracelet/lipgloss/pull/636
+replace charm.land/lipgloss/v2 => github.com/jedevc/lipgloss/v2 v2.0.0-20260322175907-46dd1f11f91e
+
+// HACK: pin to rc.2 due to bug in rc.3, see https://github.com/moby/buildkit/pull/6616
+replace github.com/containerd/platforms => github.com/containerd/platforms v1.0.0-rc.2
```
jedevc: These have been removed.

jedevc (commented):
This is great! Just a few last things, I'm a bit confused why `==` seems to work, but strings aren't orderable, and so can't be compared with `==` (as the code is written).

Omkar-Ugal:
`==` was doing its own thing all along.

jedevc:
Fixed some stuff to do with equals - for example, strings and booleans can be `==` compared, but shouldn't be able to be `<` ordered.

Also there was another bug in `tags.*` comparisons, which would have broken, so I fixed that too.

Should be good to go once tests pass.

jedevc (dismissed):
```
```

Omkar-Ugal:
Thanks, Justin! Really appreciate all the feedback and the improvements.

PR #420, opened 2026-07-14
feat(config): Add option to use profile from env

craciunoiuc:
Reads four new environmental variables:
- UNIKRAFT_PROFILE_ENV=true/false
- UKC_TOKEN=\<string>
- UKC_METRO=\<string>
- UKC_ALLOW_INSECURE=true/false

Closes: TOOL-1163

### DIFF
--- a/internal/config/profile.go
+++ b/internal/config/profile.go
@@ -23,6 +23,11 @@ import (
 // any profiles yet.
 const DefaultProfile = "default"
 
+// EnvProfile is the name of the virtual profile created from the environment
+// variables UKC_TOKEN and UKC_METRO. It is used when the UNIKRAFT_PROFILE_ENV
+// environment variable is set to true.
+const EnvProfile = "env"
+
 // ErrNotSetup is returned when there are no profiles available in the
 // configuration.
 var ErrNotSetup = jujuerrors.New(heredoc.Docf(`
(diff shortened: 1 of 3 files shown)


internal/config/profile.go
jedevc, line 242: Should we update `CurrentProfileName` to a special value here?

internal/config/profile.go
```diff
@@ -202,6 +214,11 @@ func (config *Config) ProfileList() []Profile {
 // CurrentProfile returns the currently selected profile from the configuration.
 // If the profile does not exist, it returns an error.
 func (config *Config) CurrentProfile() (*Profile, error) {
+	if b, err := strconv.ParseBool(os.Getenv("UNIKRAFT_PROFILE_ENV")); err == nil && b {
```
jedevc:
```suggestion
	if b, _ := strconv.ParseBool(os.Getenv("UNIKRAFT_PROFILE_ENV")); b {
```

jedevc: Also README update.

internal/config/profile.go
```diff
@@ -90,38 +90,50 @@ func (p Profile) GetDefaultMetro() string {
...
+func profileFromEnv(name string) Profile {
+	p := Profile{
+		Name:  name,
+		Type:  ProfileTypeLegacy,
+		Token: os.Getenv("UKC_TOKEN"),
+	}
+
+	if raw := os.Getenv("UKC_METRO"); raw != "" {
+		p.Metros = []Metro{metroFromEnv(raw)}
+	}
```
jedevc, line 114: What do you think about supporting comma-separated metros? For multiple metros? 🤔

internal/config/profile.go
craciunoiuc: what is a special value? :eye:

internal/config/profile.go
craciunoiuc: Sounds a bit cursed tbh

I'd say keep it single metro to align with `kraft`

The variable exists anyway mostly for compat reasons

jedevc: Fair fair 🎉

internal/config/profile.go
jedevc: Maybe call it something like "env"?

jedevc (dismissed):
```
```

jedevc:
One last nit - can you remove the `legacy` profile type from the quickstart in the README, and replace it with the new env var.

jedevc (approved):
```
```

PR #422, opened 2026-07-15
feat(cmd): Add command to unpack rootfses

craciunoiuc:
Currently blocked on these, so in draft:
https://github.com/unikraft/go-archivefs/pull/10
https://github.com/unikraft/go-cpio/pull/2

erofs one worth discussing if it can be done cleaner

Closes: TOOL-1029

### DIFF
--- a/cmd/unikraft/integration/image_test.go
+++ b/cmd/unikraft/integration/image_test.go
@@ -7,9 +7,15 @@ package integration
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/containerd/continuity/fs/fstest"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	integ "unikraft.com/cli/internal/integration"
 )
 
 func TestImages(t *testing.T) {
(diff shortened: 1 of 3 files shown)


jedevc:
Copying from https://linear.app/unikraft/issue/TOOL-1029/add-unikraft-image-unpack#comment-edc8098e

> Hm. So I see [https://github.com/unikraft/go-archivefs/pull/10,](<https://github.com/unikraft/go-archivefs/pull/10,>) but why do the unpack instead of just providing an implementation of `fs.FS`? Then we have a generic `unpack` function which just takes an `fs.FS` in one place.
> 
> If everything is using an `fs.FS`, then for example, it makes it super easy to translate a cpio to an erofs file, without needing to unpack to disk. Essentially like the `tarfs` already does.
> 
> Remember there's an extension of this as [TOOL-1030](https://linear.app/unikraft/issue/TOOL-1030/add-unikraft-image-mount)**,** which would mount the `fs.FS` as a FUSE filesystem.

> Wait. Doesn't it already do this? [https://github.com/unikraft/go-archivefs/blob/a06ac4112571c729baf5bcafdcf1f255b82796a7/erofs/erofs.go#L41-L53](<https://github.com/unikraft/go-archivefs/blob/a06ac4112571c729baf5bcafdcf1f255b82796a7/erofs/erofs.go#L41-L53>)
>
> This is the erofs file → fs.FS reader?

craciunoiuc:
Hmm, right, I think I forgot about that in a way, I'll have to revisit

PR #423, opened 2026-07-18
feat: Add volume usage progress bar

### LINKED ISSUE #373: Add new sparkline fields
For example, when running `unikraft volume ls` it would be great to see "usage" and a quick glance progress bar.

I think this involves:
- Adding new fields to represent a percentage.
- Then rendering these as a fancy little sparkline (interesting in relation to https://github.com/unikraft/cli/issues/372, since this is a dynamic size)
- Finally, making sure that `filter` and `sort` don't just use the weird sparkline for this. So you'd be able to do things like `--filter usage=0.5`

Before:

```
METRO  NAME                  STATE    SIZE    CREATED
dal2   cloud-jetstream-data  mounted  128MiB  4 weeks ago
```

After:

```
METRO  NAME                  STATE    USAGE                  CREATED     
dal2   cloud-jetstream-data  mounted  █░░░░░░░░░ 12/128 MiB  4 weeks ago
```

maybe even some nice coloring 🙂 

* green <= 50%
* amber > 50% && <= 80%
* red > 80%

Internal issue: https://linear.app/unikraft/issue/TOOL-1027/

khatibomar:
Fixes unikraft-cloud/cli#373  <br>Closes unikraft-cloud/cli#373

image

### DIFF
--- a/internal/cmd/testdata/TestOutput/volumes
+++ b/internal/cmd/testdata/TestOutput/volumes
@@ -4,8 +4,9 @@ name:         my-volume
 uuid:         a1b2c3d4-e5f6-7890-abcd-ef1234567890
 tags:         ["env-prod"]
 state:        available
-size:         50MiB
+usage:        80%  ⣿⣿⣿⣿⣿⣿⣿⣿⣀⣀
 free:         10MiB
+size:         50MiB
 filesystem:   ext4
 quota-policy: hard
 persistent:   true
@@ -17,8 +18,9 @@ name:         my-volume
 uuid:         a1b2c3d4-e5f6-7890-abcd-ef1234567890
 tags:         ["env-prod"]
 state:        available
-size:         50MiB
+usage:        80%  ⣿⣿⣿⣿⣿⣿⣿⣿⣀⣀
 free:         10MiB
+size:         50MiB
 filesystem:   ext4
 quota-policy: hard
 persistent:   true
@@ -31,8 +33,8 @@ mounted-by:
 delete-lock:  false
 
 ==================================== table =====================================
-METRO  NAME       STATE      SIZE   CREATED
-fra    my-volume  available  50MiB
+METRO  NAME       STATE      USAGE            FREE   SIZE   CREATED
+fra    my-volume  available  80%  ⣿⣿⣿⣿⣿⣿⣿⣿⣀⣀  10MiB  50MiB
 
 ==================================== debug =====================================
 [
@@ -89,6 +91,16 @@ fra    my-volume  available  50MiB
     "verbosity": "short"
   },
   {
+    "name": "usage",
+    "value": "40MiB/50MiB",
+    "verbosity": "short"
+  },
+  {
+    "name": "free",
+    "value": "10MiB",
+    "verbosity": "short"
+  },
+  {
     "name": "size",
     "value": "50MiB",
     "verbosity": "short",
@@ -98,11 +110,6 @@ fra    my-volume  available  50MiB
     "edit": {
       "set": "50MiB"
     }
-  },
-  {
-    "name": "free",
-    "value": "10MiB",
-    "verbosity": "long"
   },
   {
     "name": "filesystem",
--- a/internal/cmd/volumes.go
+++ b/internal/cmd/volumes.go
@@ -247,15 +247,16 @@ type Volume struct {
 
 	Tags []string `mirror:"volume.tags" field:",long" create:"set" edit:"set,add,del"`
 
-	State       types.VolumeState   `mirror:"volume.state" field:",short"`
-	Size        types.SizeMebibytes `mirror:"volume.size_mb" field:",short" create:"set" edit:"set"`
-	Free        types.SizeMebibytes `mirror:"volume.free_mb" field:"free,long"`
-	Filesystem  string              `mirror:"volume.filesystem" field:",long" create:"set"`
-	QuotaPolicy string              `mirror:"volume.quota_policy" field:"quota-policy,long" create:"set" edit:"set"`
-	Persistent  bool                `mirror:"volume.persistent" field:",long"`
-	AccessMode  *types.AccessMode   `mirror:"volume.access_mode" field:",long" create:"set"`
-	HostPath    *string             `mirror:"volume.host_path" field:"host-path,long"`
-	Template    string              `field:"template,invisible,valueless" create:"set"`
+	State       types.VolumeState                     `mirror:"volume.state" field:",short"`
+	Usage       types.MeterUsage[types.SizeMebibytes] `field:"usage,short"`
+	Free        types.SizeMebibytes                   `mirror:"volume.free_mb" field:"free,short,keepzero"`
+	Size        types.SizeMebibytes                   `mirror:"volume.size_mb" field:",short" create:"set" edit:"set"`
+	Filesystem  string                                `mirror:"volume.filesystem" field:",long" create:"set"`
+	QuotaPolicy string                                `mirror:"volume.quota_policy" field:"quota-policy,long" create:"set" edit:"set"`
+	Persistent  bool                                  `mirror:"volume.persistent" field:",long"`
+	AccessMode  *types.AccessMode                     `mirror:"volume.access_mode" field:",long" create:"set"`
+	HostPath    *string                               `mirror:"volume.host_path" field:"host-path,long"`
+	Template    string                                `field:"template,invisible,valueless" create:"set"`
 
 	Timestamps struct {
 		Created types.RelativeTime `mirror:"volume.created_at" field:",short"`
@@ -386,6 +387,14 @@ func (Volume) load(ref *group.Ref, volume platform.Volume, metro *config.Metro) 
 	if err != nil {
 		return Volume{}, fmt.Errorf("could not mirror volume data: %w", err)
 	}
+
+	if result.Size > 0 {
+		result.Usage = types.MeterUsage[types.SizeMebibytes]{
+			Used:  result.Size - result.Free,
+			Total: result.Size,
+		}
+	}
+
 	return result, nil
 }
 
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -98,6 +98,10 @@ type Field struct {
 	// display settings
 	Verbosity FieldVerbosity `json:"verbosity"`
 	Hyperlink string         `json:"hyperlink,omitempty"`
+	// KeepZero indicates the field's zero value is meaningful (e.g. "0B"
+	// free space) and should not be treated as empty/omittable the way an
+	// unset value normally is.
+	KeepZero bool `json:"keep_zero,omitempty"`
 
 	// settings for creating or patching resources
 	Create *Patch `json:"create,omitempty"`
@@ -110,7 +114,7 @@ func (f Field) HasChildren() bool {
 
 func (f Field) IsEmpty() bool {
 	if f.Value != nil {
-		if reflect.ValueOf(f.Value).IsZero() {
+		if reflect.ValueOf(f.Value).IsZero() && !f.KeepZero {
 			return true
 		}
 		if s, ok := f.Value.(fmt.Stringer); ok && s.String() == "" {
--- a/internal/resource/value/compare.go
+++ b/internal/resource/value/compare.go
@@ -41,6 +41,10 @@ func Compare(a, b any) int {
 		return 1
 	}
 
+	if result, ok := compareSelf(a, b); ok {
+		return result
+	}
+
 	// Try type-specific comparisons
 	switch av := a.(type) {
 	case time.Time:
@@ -144,6 +148,31 @@ func Compare(a, b any) int {
 		bStr = fmt.Sprint(b)
 	}
 	return cmp.Compare(aStr, bStr)
+}
+
+// compareSelf detects and invokes a "Compare(T) int" method on a's type,
+// where T is the same type as a, matching the cmp.Compare contract.
+func compareSelf(a, b any) (int, bool) {
+	av := reflect.ValueOf(a)
+	bv := reflect.ValueOf(b)
+	if av.Type() != bv.Type() {
+		return 0, false
+	}
+
+	method := av.MethodByName("Compare")
+	if !method.IsValid() {
+		return 0, false
+	}
+
+	mt := method.Type()
+	if mt.NumIn() != 1 || mt.In(0) != av.Type() {
+		return 0, false
+	}
+	if mt.NumOut() != 1 || mt.Out(0).Kind() != reflect.Int {
+		return 0, false
+	}
+
+	return int(method.Call([]reflect.Value{bv})[0].Int()), true
 }
 
 // compareReflect attempts to compare values using reflection.
--- a/internal/resource/value/compare.go
+++ b/internal/resource/value/compare.go
@@ -158,21 +158,34 @@ func compareSelf(a, b any) (int, bool) {
 	if av.Type() != bv.Type() {
 		return 0, false
 	}
+	if !hasSelfCompareMethod(av) {
+		return 0, false
+	}
 
 	method := av.MethodByName("Compare")
-	if !method.IsValid() {
-		return 0, false
-	}
-
-	mt := method.Type()
-	if mt.NumIn() != 1 || mt.In(0) != av.Type() {
-		return 0, false
-	}
-	if mt.NumOut() != 1 || mt.Out(0).Kind() != reflect.Int {
-		return 0, false
-	}
-
 	return int(method.Call([]reflect.Value{bv})[0].Int()), true
+}
+
+// HasSelfCompare reports whether v has a "Compare(T) int" method where T is
+// v's own type, matching the cmp.Compare contract. Compare uses this same
+// detection internally (see compareSelf) to dispatch to a type's own
+// ordering; callers that need to know upfront whether such ordering exists
+// (e.g. to decide whether a value can be used with an ordering filter) can
+// check it directly.
+func HasSelfCompare(v any) bool {
+	if v == nil {
+		return false
+	}
+	return hasSelfCompareMethod(reflect.ValueOf(v))
+}
+
+func hasSelfCompareMethod(rv reflect.Value) bool {
+	method := rv.MethodByName("Compare")
+	if !method.IsValid() {
+		return false
+	}
+	mt := method.Type()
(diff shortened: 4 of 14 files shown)


internal/types/quota.go
```diff
@@ -46,3 +51,81 @@ func (r Range[T]) String() string {
...
+	case ratio <= 0.8:
+		color = ansi.NewStyle().ForegroundColor(colors.Warning).String()
+	default:
+		color = ansi.NewStyle().ForegroundColor(colors.Error).String()
+	}
+
+	emptyColor := ansi.NewStyle().ForegroundColor(ansi.IndexedColor(238)).String()
+	reset := ansi.NewStyle().ForegroundColor(nil).String()
+
+	dots := []rune{'⣀', '⣄', '⣤', '⣦', '⣶', '⣷', '⣿'}
```
craciunoiuc: I wonder if this format can be found somewhere in `bubbletea` or smth else? I've seen it before

internal/cmd/volumes.go
```diff
@@ -248,7 +248,7 @@ type Volume struct {
 	Tags []string `mirror:"volume.tags" field:",long" create:"set" edit:"set,add,del"`
 
 	State       types.VolumeState                         `mirror:"volume.state" field:",short"`
-	Size        types.SizeMebibytes                       `mirror:"volume.size_mb" field:",short" create:"set" edit:"set"`
+	Size        types.SizeMebibytes                       `mirror:"volume.size_mb" field:",long" create:"set" edit:"set"`
 	Free        types.SizeMebibytes                       `mirror:"volume.free_mb" field:"free,long"`
```
craciunoiuc: Why remove it from `volume ls`? I don't think it took that much space? 🤔

craciunoiuc (commented):
Some comments

khatibomar:
image

Added a custom compare for custom types like sparkling line.

It would be nice to support >,<,>=,<= operators but I seen that it already have an issue 

https://github.com/unikraft-cloud/cli/issues/374

internal/cmd/volumes.go
khatibomar: It was listed in this issue that field should be remove so I did.

It's repeated in the usage cell too, so kinda repetitive information

But yea, I readded it now.

khatibomar: this is currently done
image

altho it makes more sense that free after usage but for the eye it feels wrong

image

craciunoiuc:
@khatibomar can you close the copilot comments that you resolved/ignored so they don't get in the way :pray:

internal/types/quota.go
```diff
@@ -46,3 +51,84 @@ func (r Range[T]) String() string {
...
+
+	var color string
+	switch {
+	case ratio <= 0.5:
+		color = ansi.NewStyle().ForegroundColor(colors.Success).String()
+	case ratio <= 0.8:
+		color = ansi.NewStyle().ForegroundColor(colors.Warning).String()
+	default:
+		color = ansi.NewStyle().ForegroundColor(colors.Error).String()
+	}
```
craciunoiuc: Just checking, did you compare these colors with the quota bars in the quota command? Same for the percentages where these change

internal/types/quota.go
```diff
@@ -46,3 +51,84 @@ func (r Range[T]) String() string {
...
+
+	emptyColor := ansi.NewStyle().ForegroundColor(colors.Slate500).String()
+	reset := ansi.NewStyle().ForegroundColor(nil).String()
+
+	dots := []rune{'⣀', '⣄', '⣤', '⣦', '⣶', '⣷', '⣿'}
+	maxSteps := float64(width * (len(dots) - 1))
+	steps := int(math.Round(ratio * maxSteps))
+
+	fullCells := steps / (len(dots) - 1)
+	rem := steps % (len(dots) - 1)
```
craciunoiuc: I don't know how to count so I trust the math is mathing here w.r.t the dots :D

internal/types/quota.go
```diff
@@ -46,3 +51,84 @@ func (r Range[T]) String() string {
...
+	ratio := float64(u.Used) / float64(u.Total)
+	if ratio > 1 {
+		ratio = 1
+	}
+	if ratio < 0 {
+		ratio = 0
+	}
+
+	const width = 10
+
```
craciunoiuc: could the width be easily made variable? Aka, min 5, max 10 width, like the quotas

craciunoiuc (commented):
Some sanity comments my side. Feel free to comment or ignore

I think it looks fine already 🙏 

I think after we can get @jedevc also in here for a final opinion

internal/types/quota.go
khatibomar: I think the quota have only a blue bar, it doesn't have a color based on percentage.

internal/types/quota.go
khatibomar: yes, it's basically a math to calculate the full bars and empty bars and the rest are the dots in between, scaled to fit the width

internal/types/quota.go
khatibomar: if you mean `metro_quotas.go` they are set to be depending on the screen size, I think same thing can be done in here and limited between [5,10] range

craciunoiuc: Yes exactly, I'm thinking if it would make sense here also to make it variable, so smaller terminals also work

internal/types/quota.go
craciunoiuc: Ah sorry, it means we removed that in the end, all good I think, but I'll tag maybe also @jedevc 

Do we want percentage colouring  or do we remove colors here also?

internal/cmd/testdata/TestOutput/volumes
```diff
@@ -6,6 +6,7 @@ tags:         ["env-prod"]
 state:        available
 size:         50MiB
 free:         10MiB
+usage:        ⣿⣿⣿⣿⣿⣿⣿⣿⣀⣀ 40MiB/50MiB
```
jedevc: nit: I wonder if maybe we should show the number first, then the number first?

```suggestion
usage:        40MiB/50MiB ⣿⣿⣿⣿⣿⣿⣿⣿⣀⣀
```

Or maybe even as a percentage?

```suggestion
usage:        80% ⣿⣿⣿⣿⣿⣿⣿⣿⣀⣀
```

internal/cmd/testdata/TestOutput/volumes
```diff
@@ -19,6 +20,7 @@ tags:         ["env-prod"]
...
```
jedevc:
```
❯ unikraft volume inspect foo -oquiet -f usage
⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀ 0B/10MiB
```

Quiet mode should show the direct underlying value, ideally as a percentage, see above.

internal/resource/value/compare.go
```diff
@@ -20,6 +20,11 @@ func IsZero(v any) bool {
 	return reflect.ValueOf(v).IsZero()
 }
 
+// Comparer is an interface for types that provide a custom comparable value.
+type Comparer interface {
+	CompareValue() any
+}
```
jedevc: I don't think we should do *this* interface. Things should implement `cmp.Compare` to do the comparison.

Somewhere else I had the idea that we should have a custom `Render` implementation for rendering to the screen, but for filtering, we would just use `.String()` or `.MarshalText` or whatever instead of doing the weird post-ASCII filtering.

internal/types/quota.go
```diff
@@ -46,3 +51,84 @@ func (r Range[T]) String() string {
...
+	if u.Total == 0 {
+		return float64(0)
+	}
+	return float64(u.Used) / float64(u.Total)
+}
+
+func (u SparklineUsage[T]) String() string {
+	if u.Total == 0 {
+		return "N/A"
+	}
```
jedevc, line 111: Essentially, as mention above, this method would just become `Render`, while `String` would be implemented to be just the raw percentage.

This would be a bit of a refactor in a few other places, but the code would then feel much neater I think :tada:

internal/types/quota.go
jedevc: Hm, we should have quotas coloring :tada:

craciunoiuc:
What do you mean quotas coloring you told me to remove the colors some time ago :sob: :sob: :sob: :sob:

jedevc:
> What do you mean quotas coloring you told me to remove the colors some time ago 😭 😭 😭 😭


I seem to remember suggesting making the color bars have reasonable boundaries, there were some very magic numbers in there as I remember it :thinking: https://github.com/unikraft-cloud/cli/pull/337#discussion_r3411546380

I don't remember suggesting removing them entirely, but maybe I did - if so, woops.

internal/cmd/testdata/TestOutput/volumes
nderjung: hmmm prefer other way - since the bar is fixed width and looks more aligned in a table

craciunoiuc:
Ah, maybe I got it wrong, eh too late now :D

internal/cmd/testdata/TestOutput/volumes
jedevc: Fair point. We could always " " pad it if it was a percentage.

I think I more care about the percentage than the side it's on, though I would argue that having the most significant info (the number) first is useful.

nderjung: That works, i think separating % and actual usage with the bar in the middle makes it easier to read:

```
METRO  NAME        STATE      FREE   USAGE                           CREATED
fra    my-volume1  available  10MiB   50% ⣿⣿⣿⣿⣿⣀⣀⣀⣀⣀ 50MiB/100MiB   just now
fra    my-volume2  available  10MiB  100% ⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿ 100MiB/100MiB  just now
```

in contrast together feels noisy:

```
METRO  NAME        STATE      FREE   USAGE                           CREATED
fra    my-volume1  available  10MiB  ⣿⣿⣿⣿⣿⣀⣀⣀⣀⣀  50% 50MiB/100MiB   just now
fra    my-volume2  available  10MiB  ⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿ 100% 100MiB/100MiB  just now
```

Edit: would argue we can reduce it further by removing duplicate `MiB` and removing the `FREE` column:

```
METRO  NAME        STATE      USAGE                        CREATED
fra    my-volume1  available   50% ⣿⣿⣿⣿⣿⣀⣀⣀⣀⣀ 50/100MiB   just now
fra    my-volume2  available  100% ⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿ 100/100MiB  just now
```

internal/types/quota.go
nderjung: All these vars should also be defined as `const` / private internal `var`.

internal/types/quota.go
nderjung:
```
var (
  emptyColor = // slate
  safeColor = // green 
  warningColor = // amber > 60%
  dangerColor = // red-rose > 80%
```

Also note light-dark theme variations, e.g.: https://github.com/unikraft/cli/blob/prod-staging/internal/tui/selector/colors.go#L15-L17

internal/types/quota.go
```diff
@@ -46,3 +51,84 @@ func (r Range[T]) String() string {
...
+	case ratio <= 0.5:
+		color = ansi.NewStyle().ForegroundColor(colors.Success).String()
+	case ratio <= 0.8:
+		color = ansi.NewStyle().ForegroundColor(colors.Warning).String()
+	default:
+		color = ansi.NewStyle().ForegroundColor(colors.Error).String()
+	}
+
+	emptyColor := ansi.NewStyle().ForegroundColor(colors.Slate500).String()
+	reset := ansi.NewStyle().ForegroundColor(nil).String()
```
nderjung: when using lipgloss we should not need to do any resets if strings are correctly wrapped (`Render`ed) since it handles it for us.

internal/types/quota.go
```diff
@@ -46,3 +51,84 @@ func (r Range[T]) String() string {
...
+	~int | ~int8 | ~int16 | ~int32 | ~int64 |
+		~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 |
+		~float32 | ~float64
+}
+
+// SparklineUsage represents a used/total pair for quotas, rendered as a colored sparkline.
+type SparklineUsage[T numeric] struct {
+	Used  T `field:",long"`
+	Total T `field:",long"`
+}
```
nderjung: Suggested internal package `internal/tui/sparkline` -> can then be used in other places?

Also actually this is technically not a sparkline (my mistake for calling this in the original issue - sparklines are timeseries data), better to call it "meter" or "guage" or "usagebar" ?

internal/types/quota.go
jedevc: Aha one would think so. I think there's a reason to not do this though - lipgloss sends resets for *everything*, not just the styles you set. I seem to remember that this does super annoying things in combo with the diff viewer when doing `edit`s for example.

internal/resource/value/compare.go
khatibomar: `cmp.Compare` doesn't have an interface unfortunately it's `func Compare[T Ordered](x, y T) int` where `Ordered` is

```go
type Ordered interface {
	~int | ~int8 | ~int16 | ~int32 | ~int64 |
		~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr |
		~float32 | ~float64 |
		~string
}
```

Altho, I did implemente compare-like method on sparkline and it did work!

I believe part of the reflection code catches this method and use it for sorting.

khatibomar: Well it's actually just sorting as string, it just happen to be true, as soon I added percentage at first it broke XD

internal/types/quota.go
khatibomar: I am not sure that I understand this one, I haven't worked on bubbletea before, I think I have to toy around with it.

jedevc: So bubbletea's colors are different from using `ansi` - we do want to actually use `ansi` directly. `lipgloss` also has colors, but the way it does colors interacts badly with our custom diff view that shows on `edit`. You can give it a go, and see what happens 🙂

internal/cmd/testdata/TestOutput/volumes
jedevc: imo, i would keep `usage` as just percent + bar. while the free + size appear as columns after it. makes it more obvious that they can be selected.

khatibomar: @jedevc like that?

image

khatibomar: that empty space in FREE kinda when disk full annoys me a bit

jedevc: Oh that is super odd. Is that because `IsEmpty` is applying…? We should fix that. We should display 0 here.

jedevc: But yes, that is generally correct and what I was hoping for!

khatibomar: Let's hope that none of the users relying on output of this CLI column shape/ordering, their scripts will 🍳

jedevc: Yeah, that's okay 🙂 users who care about this should be using `-ojson`.

khatibomar:
image
Screenshot_20260730_191809
Screenshot_20260730_191911
image
image


@jedevc I addressed the comments, I think it's ready for re-review.

craciunoiuc:
I mean, I like it, so unless someone has an objection, I think we should merge? :eyes:

jedevc:
Will review again on Monday - but yes, we can aim to merge soon.

internal/resource/resource.go
```diff
@@ -98,6 +98,10 @@ type Field struct {
 	// display settings
 	Verbosity FieldVerbosity `json:"verbosity"`
 	Hyperlink string         `json:"hyperlink,omitempty"`
+	// KeepZero indicates the field's zero value is meaningful (e.g. "0B"
+	// free space) and should not be treated as empty/omittable the way an
+	// unset value normally is.
```
jedevc, line 103: Mmm, IMO, this should actually be the default.

Instead, we should be inspired by `encoding/json`, and use an `omitzero` tag.

This will need scanning through all the fields, and determining which of these are worth omitting, but I think probably a worthwhile thing to do anyways.

internal/types/quota.go
```diff
@@ -46,3 +50,79 @@ func (r Range[T]) String() string {
...
+		return 0
+	}
+	return ratio
+}
+
+// String returns the used/total ratio as a raw percentage, e.g. "80%", or
+// "N/A" if Total is 0.
+func (u MeterUsage[T]) String() string {
+	if u.Total == 0 {
+		return "N/A"
```
jedevc: IMO, this should just be `""` - e.g. if parsing this field, users shouldn't have to handle `N/A` as a string.

internal/resource/value/compare.go
jedevc, line 1: Can you rebase - and make sure that this works with `--filter usage>0.5` (or whatever needs to go there)? And ideally add a test for it.

jedevc (commented):
Some very small things. LGTM :tada:

internal/resource/resource.go
khatibomar: Should we do that rn, or keep it as a follow up PR 🤔

internal/resource/value/compare.go
khatibomar: image

It didn't work out of the box, so I pushed https://github.com/unikraft-cloud/cli/pull/423/commits/aae5f9106060daaf4dd8c7db2711b656783fa717 so it works.

khatibomar: image

internal/resource/resource.go
jedevc: Happy for that to be a follow-up.

internal/types/quota.go
```diff
@@ -46,3 +53,135 @@ func (r Range[T]) String() string {
...
+}
+
+// UnmarshalText parses MeterUsage back from the "used/total" or "N/A" form
+// MarshalText produces, round-tripping it. As a convenience for ordering
+// filters (e.g. --filter "usage>0.5"), it also accepts a bare ratio (e.g.
+// "0.5") or percentage (e.g. "50%"), building a used/total pair with a
+// matching ratio at the same 1% resolution String renders at.
+func (u *MeterUsage[T]) UnmarshalText(text []byte) error {
+	s := strings.TrimSpace(string(text))
+	if s == "" || s == "N/A" {
```
jedevc: Likewise about `N/A` here - can we remove it from this entire file?

jedevc (approved):
Nice nice nice :tada:

```
```

PR #429, opened 2026-07-25
ci: Add test monitors and export data to betterstack

jedevc:
Runs the tests at regular intervals, allowing some flakes, and then pushes the data into betterstack.

### DIFF
--- a/.github/workflows/monitor.yaml
+++ b/.github/workflows/monitor.yaml
@@ -1,0 +1,96 @@
+name: monitor
+permissions:
+  contents: read
+
+on:
+  schedule:
+  - cron: '*/30 * * * *'
+  workflow_dispatch: {}
+
+# Automatically cancel in-progress runs of this workflow
+concurrency:
+  group: ${{ github.workflow }}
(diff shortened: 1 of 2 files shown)


.github/workflows/monitor.yaml
```diff
@@ -0,0 +1,96 @@
...
+    name: Export trace to Better Stack
+    runs-on: ubuntu-latest
+    needs: [integration]
+    steps:
+      - name: Export workflow trace
+        uses: dash0hq/otel-cicd-action@v4
+        with:
+          otlpEndpoint: https://${{ secrets.BETTERSTACK_INGESTING_HOST }}/v1/traces
+          otlpHeaders: Authorization=Bearer ${{ secrets.BETTERSTACK_SOURCE_TOKEN }}
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
```
jedevc, line 96: This only exports workflow-level traces. Ideally, we would also be exporting test level traces to monitor which tests exactly are causing failures.

PR #432, opened 2026-07-28
feat(cmd): Accept namespaces for listing

craciunoiuc:
Seems straight-forward

Depends on: https://github.com/unikraft-cloud/console/pull/1065

Closes: TOOL-760

### DIFF
--- a/internal/cmd/images.go
+++ b/internal/cmd/images.go
@@ -37,12 +37,12 @@ import (
 
 type ImagesCmd struct {
 	cmd.ResourceCmd[ImageEntry]
-	cmd.ListableResourceCmd[ImageEntry]
 
 	Build  ImageBuildCmd   `cmd:"" help:"Build a Unikraft project into a container image."`
 	Get    ImagesGetCmd    `cmd:"" help:"Inspect an image." aliases:"inspect,show"`
 	Delete ImagesDeleteCmd `cmd:"" help:"Remove an image." aliases:"rm,remove"`
 	Copy   ImagesCopyCmd   `cmd:"" help:"Copy images."`
+	List   ImagesListCmd   `cmd:"" help:"List images." aliases:"ls"`
 }
 
 type Image struct {
@@ -66,6 +66,8 @@ type ImageConfig struct {
 	Cmd []string          `field:",long"`
 	Env map[string]string `field:",long"`
 }
+
+type imageNamespacesKey struct{}
 
 type ImageMetadata struct {
 	Author  string     `field:",long"`
@@ -300,7 +302,11 @@ func (ImageEntry) List(ctx context.Context) ([]resource.Resource, error) {
 	eg, ctx := joinerrgroup.WithContext(ctx)
 	eg.Go(func() error {
 		log.G(ctx).Trace().Msg("listing images from controlplane")
-		resp, err := client.ListImages(ctx, controlplane.ListImagesOpts{Details: new(true)})
+		namespaces, _ := ctx.Value(imageNamespacesKey{}).([]string)
+		resp, err := client.ListImages(ctx, controlplane.ListImagesOpts{
+			Details:   new(true),
+			Namespace: namespaces,
+		})
 		if err != nil {
 			return err
 		}
@@ -348,6 +354,7 @@ func listPlatformImages(ctx context.Context) ([]resource.Resource, error) {
 	if err != nil {
 		return nil, err
 	}
+	namespaces, _ := ctx.Value(imageNamespacesKey{}).([]string)
 
 	return group.CollectAllSlices(ctx, g, func(ctx context.Context, c multimetro.MetroClient) ([]resource.Resource, error) {
 		log.G(ctx).Trace().Msg("listing images from image-store")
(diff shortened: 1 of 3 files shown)


internal/cmd/images.go
```diff
@@ -373,6 +380,9 @@ func listPlatformImages(ctx context.Context) ([]resource.Resource, error) {
 					continue
 				}
 				for _, entry := range entries {
+					if len(namespaces) > 0 && !slices.Contains(namespaces, entry.Namespace) {
+						continue
+					}
```
jedevc, line 385: 🤔 Why is this needed? Is the response including namespaces that weren't selected? If so, that feels more like a server-side bug. If not, then this check isn't needed.

internal/cmd/images.go
```diff
@@ -67,6 +67,8 @@ type ImageConfig struct {
 	Env map[string]string `field:",long"`
 }
 
+type imageNamespacesKey struct{}
```
jedevc, line 70: 😢 very sad, I don't love it, but it gets the functionality done. Such is life 🤷

jedevc (commented):
Tests pls 🙏

internal/cmd/images.go
craciunoiuc: :magic_wand:

internal/cmd/images.go
craciunoiuc: Ok so apparently you were correct to doubt it, but not fully :D

The controlplane API filters namespaces correctly  so this `c.GetImageStore`

But if that fails the fallback doesn't (it returns everything that you can see) `c.GetImages`. Aka the code for old metros

There's two options: I add the if here, or I duplicate some code and add the for+if in the error branch together with the fallback

I'm actually fine either way, the end result is the same (and we can get rid of the code when we update nodes)

craciunoiuc: I can also add a note to it so we don't forget (I don't think we will :triumph: )

jedevc: Hm, doesn't the fallback support it? [https://github.com/unikraft-cloud/platform/blob/de19d04f99ce09404f9391ac4e7a61c7352a8d4c/src/api/http/v1/images/list.c#L80-L84](<https://github.com/unikraft-cloud/platform/blob/de19d04f99ce09404f9391ac4e7a61c7352a8d4c/src/api/http/v1/images/list.c#L80-L84>)

craciunoiuc: doesn't hurt to try I guess though there might be a chance that that was not the case 1-2 years ago when that version is from

jedevc: Can we just that one? Instead of doing client-side filtering.

craciunoiuc: yup, I have time now so I can try it out

craciunoiuc: Ok so apparently the endpoint `/v1/images` and `/v1/images/list` accept a namespace argument and parse it

But apparently it is never used in the code after parsing :dizzy: 
<https://github.com/unikraft-cloud/platform/blob/87f2b99238008b6f573fec83ba50c18084d4534f/src/api/http/v1/images/list.c#L67>

craciunoiuc: So I think it's safe to assume that it never worked server side? That field is not used anywhere and it was never used since I checked in ~`v0.8`

jedevc: Huh, okay. Can you add a comment here, explaining that the server-side `namespace` for the fallback doesn't work.

We also shouldn't do `strings.Contains` - we should actually parse the reference in the same way we do the rest of the CLI, and filter it out properly. `strings.Contains` will include too much (e.g. listing `--namespace official` would include `jedevc/official`)

craciunoiuc: dayum, nice catch

will do

jedevc (changes requested):
Still have comments, and also still needs tests.

craciunoiuc:
yeah, I didn't get to it yet ;_;

internal/cmd/images.go
craciunoiuc: Hmm, so your argument is correct, but at the same time wrong :dizzy: 

Because the code has `slices.Contains` not `strings.Contains` and the slices one does an exact match per element :eyes:

craciunoiuc:
comment was addressed (actually nothing to do)

tests successfully vibrated

craciunoiuc:
(lemme know if you want a single test for these kind of things, aka the golden path, or trying everything)

jedevc:
Yeah, we should just have this integrated into the rest of the tests - no need to do a new test setup or a custom http server.

jedevc (changes requested):
Needs changes.

craciunoiuc:
boom

jedevc:
Boom 
image

craciunoiuc:
Robot result:
```
Harbor's robot tokens only grant access to the user's own project plus official, so for any other namespace Harbor answers 403 FORBIDDEN. Console's listImages treated any Harbor error as fatal and wrapped it in errz.NewErrorInternalError, so the whole listing came back as a 500
```

So something to fix in the controlplane also, but for now I'm doing a small fix here also. @jedevc

jedevc:
Okay, prod-stable console is deployed with that fix.

jedevc:
Hm, the issue is still there.

craciunoiuc:
Ok, wrote on discord. I had some uncomitted code

Linked and tested. Looks fine now

jedevc (approved):
```
```

PR #433, opened 2026-07-28
fix(config): Strip urls only if fully qualified

craciunoiuc:
So if you look at the parse it used string parsing. This means that the `/v1` suffix stayed in the host variable aka the name of the node.

Or maybe you like the AI explanation better:
```
URL parser derived name before removing /v1 → client registered as localhost:8080/v1; resource requests select localhost:8080 → missing client.
```

Closes: TOOL-1140

### DIFF
--- a/internal/config/profile.go
+++ b/internal/config/profile.go
@@ -117,16 +117,14 @@ func profileFromEnv(name string) Profile {
 
 // metroFromEnv parses the UKC_METRO environment variable into a Metro.
 func metroFromEnv(raw string) Metro {
-	endpoint := raw
-	var name string
-	if strings.Contains(endpoint, "://") {
-		_, host, _ := strings.Cut(endpoint, "://")
-		host = strings.TrimPrefix(host, "api.")
-		name, _, _ = strings.Cut(host, ".")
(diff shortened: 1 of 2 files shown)


internal/config/profile.go
```diff
@@ -119,9 +119,8 @@ func profileFromEnv(name string) Profile {
 func metroFromEnv(raw string) Metro {
 	endpoint := raw
 	var name string
-	if strings.Contains(endpoint, "://") {
-		_, host, _ := strings.Cut(endpoint, "://")
-		host = strings.TrimPrefix(host, "api.")
+	if u, err := url.Parse(endpoint); err == nil && u.IsAbs() && u.Host != "" {
+		host := strings.TrimPrefix(u.Host, "api.")
 		name, _, _ = strings.Cut(host, ".")
 		endpoint = strings.TrimSuffix(strings.TrimSuffix(endpoint, "/"), "/v1")
```
jedevc: Instead of doing this then, can't we just reformat the URL?

Or alternatively, just share the `/v1` stripping between name + endpoint.

I like the idea of moving away from just string - based parsing, but we should then do it for the endpoint as well.

jedevc (approved):
```
```

PR #436, opened 2026-08-04
feat: Add sandbox integration

dragosgheorghioiu:


internal/cmd/shell.go
jedevc, line 1: when i type the screen flickers 🤔

```
❯ gnome-terminal --version
# GNOME Terminal 3.60.0 using VTE 0.84.0 +BIDI +GNUTLS +ICU +SYSTEMD
```

jedevc, line 1: If the plugin isn't loaded, it should give a reasonable error, not:
```
error: failed on "staging" client: failed to start command: performing the request: request failed: 404 Not Found
parsing response: invalid character '<' looking for beginning of value
```

jedevc:
Can you update the description with info on how to run the plugin? Also probably worth looking at https://linear.app/unikraft/issue/TOOL-1319/publish-plugins-images.

internal/cmd/shell.go
```diff
@@ -0,0 +1,461 @@
...
+		if line == "" {
+			continue
+		}
+
+		if line == "exit" {
+			return nil
+		}
+
+		if line == "clear" {
+			print("\033[H\033[2J")
```
jedevc: We should use `ansi` / `lipgloss`, instead of manually constructed escape sequences.

internal/cmd/shell_builtins.go
jedevc, line 1: Let's move this and all *other* shell stuff into `internal/shell`. `shell` is a complicated command, so let's keep all of it together.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,581 @@
...
+	"unikraft.com/cloud/sdk/sandbox"
+	"unikraft.com/x/kingkong"
+	"unikraft.com/x/log"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/multimetro"
+)
+
+type (
+	SandboxInstanceCmd []string
```
jedevc: Why a separate type for this?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,581 @@
...
+
+	if len(env) > 0 {
+		var envBuf strings.Builder
+		envBuf.WriteString("env ")
+		for k, v := range env {
+			escapedVal := "'" + strings.ReplaceAll(v, "'", "'\\''") + "'"
+			fmt.Fprintf(&envBuf, "%s=%s ", k, escapedVal)
+		}
+		prefix += envBuf.String()
+	}
```
jedevc: I thought the API was updated so we didn't need this anymore.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,581 @@
...
+				return
+			}
+			eofReq := sandbox.FeedInstanceCommandStdinRequestBody{Eof: true}
+			_, _ = c.Sandbox.FeedInstanceCommandStdin(ctx, instanceUUID, pluginName, cmdUUID, eofReq)
+			return
+		}
+	}
+}
+
+type WriteSandboxInstanceCmd struct {
```
jedevc, line 423: Maybe we should call this `cp`? We could make it follow `scp` semantics if you really wanted to. So `name:path`.

Since that's what it feels closer to. Maybe at least an alias.

Same with `cat` for read.

internal/cmd/sandbox_test.go
jedevc, line 1: Generally, we're missing full-on integration tests. I don't really think we need tests that the parsing is correct, I'm more interested that this works against a deployed instance with a sandbox.

internal/cmd/shell.go
```diff
@@ -0,0 +1,461 @@
...
+	"get": true, "help": true, "restart": true, "start": true,
+	"stop": true, "suspend": true, "mount": true, "unmount": true,
+	"edit": true, "volumes": true, "history": true,
+}
+
+func (s *shellState) handleBuiltin(sctx *ShellContext, line string) bool {
+	fields := shellFields(line)
+	if len(fields) == 0 {
+		return false
+	}
```
jedevc: Because we're not using the evaluator things in `mvdan.sh`, it means that things like `volumes mounted > out` don't work. Also `echo $(volumes mounted)` won't work, because it's not present in the shell.

If we really don't want to use the evaluator (and want to directly rely on the host's shell), we need to be clear that these are special commands (some sort of prefix, maybe `/`, like `/volumes mounted`)?

internal/cmd/shell.go
```diff
@@ -0,0 +1,461 @@
...
+		env:     env,
+		running: running,
+	}
+}
+
+var knownBuiltins = map[string]bool{
+	"get": true, "help": true, "restart": true, "start": true,
+	"stop": true, "suspend": true, "mount": true, "unmount": true,
+	"edit": true, "volumes": true, "history": true,
+}
```
jedevc: Can't we derive this from `shell_builtins.go`? Or determine it based on error?

internal/cmd/shell.go
```diff
@@ -0,0 +1,461 @@
...
+	parser, err := kong.New(&cmd,
+		kong.Name(""),
+		kong.NoDefaultHelp(),
+		kong.Exit(func(int) {}),
+		kong.Writers(sctx.Out, sctx.ErrOut),
+	)
+	if err != nil {
+		fmt.Fprintf(sctx.ErrOut, "%s failed to initialize shell parser: %v\n", shell.ShellErrorStyle.Render("error:"), err)
+		return true
+	}
```
jedevc: I think we may be able to cache this between calls. It's the same each time we run it right? (except for `Out` and `ErrOut`, but I think you can set those after)

Just trying to avoid needing to do reflection on every single command run.

internal/cmd/shell_builtins.go
```diff
@@ -0,0 +1,714 @@
...
+
+// builtinHelp prints the list of available shell builtins.
+func builtinHelp(out io.Writer) {
+	fmt.Fprintln(out, shell.ShellTitleStyle.Render("Builtins:"))
+	fmt.Fprintln(out)
+	fmt.Fprintf(out, "  %s %s\n", shell.ShellValueStyle.Render("cd"), shell.ShellHintStyle.Render("change the current remote directory"))
+	fmt.Fprintf(out, "  %s %s\n", shell.ShellValueStyle.Render("get"), shell.ShellHintStyle.Render("inspect the current instance"))
+	fmt.Fprintf(out, "  %s %s\n", shell.ShellValueStyle.Render("edit"), shell.ShellHintStyle.Render("edit instance fields (env, args, memory, vcpus, tags)"))
+	fmt.Fprintf(out, "  %s %s\n", shell.ShellValueStyle.Render("volumes"), shell.ShellHintStyle.Render("list volumes mounted on this instance (alias for volumes mounted)"))
+	fmt.Fprintf(out, "  %s %s\n", shell.ShellValueStyle.Render("volumes mounted"), shell.ShellHintStyle.Render("list volumes mounted on this instance"))
```
jedevc: Let's build this as a table. Or even use a help printer, similar to how kingkong does it, instead of repeating it over and over.

internal/multimetro/client.go
```diff
@@ -21,7 +22,11 @@ import (
 
 type MetroClient struct {
 	platform.Client
-	Metro config.Metro
+	// Sandbox serves the sandbox plugin API of the same metro.  It cannot be
+	// embedded alongside platform.Client, since both name their interface
+	// Client.
```
jedevc: excess comment.

internal/cmd/sandbox.go
dragosgheorghioiu: just a personal preference here 

i will remove it

internal/cmd/sandbox.go
dragosgheorghioiu: it was not it's on separate branch

internal/cmd/sandbox.go
dragosgheorghioiu: i am not sure i see the point here
would scp semantincs be useful if we already know that for read the source is the sandbox and for write the source is local

internal/cmd/sandbox_test.go
dragosgheorghioiu: agree i can add them

internal/cmd/sandbox.go
jedevc: I was told by Michalis that it's ready to merge as soon as we're (me, you, alex) are happy with the API. Can you confirm @nurof3n?

jedevc: I would get it ready to use that.

internal/cmd/sandbox.go
jedevc: Mmm, you're right. `cp` semantics would be nice, but it would be good to keep lower-level read/write.

With `cp`, you could do `unikraft instance copy foo instance-a:bar` or `unikraft instance copy instance-b:foo bar` or even `unikraft instance cp instance-c:foo instance-d:bar`, all in one command.

internal/cmd/shell.go
dragosgheorghioiu: the builtins are documented in the help menu but i can make the fact clear that they cannot run on the sandbox directly

maybe some different syntax highlighting for them can also be a plus

jedevc: Indeed, but we should be very clear that they are evaluated differently - different syntaxing is a plus, but also additional syntax to really clarify is useful too. Otherwise, users will attempt to use it like I showed, and will be surprised that it doesn't.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,607 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+
+package cmd
```
nderjung, line 5: can we move this file to `instance_exec.go` ?

nderjung: Nvm.

I'd like to start breaking these files up they are getting quite long and it would be good to match the file name with subcommand. We can leave that as separate work item.

internal/cmd/shell.go
dragosgheorghioiu: maybe a special "builtin mode" that is triggered by keybind where you can only run builtins
maybe that would be clear enough for the user

jedevc: That's also possible. I've seen some cool stuff before where you type something like `!` at the beginning of the line, and the prompt changes - but if you backspace, it goes back to the normal prompt. So it's not like a "mode", it's more like a nice display on top of a prefix.

dragosgheorghioiu: that sounds even better i can try that

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,607 @@
...
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+
+	"mvdan.cc/sh/v3/syntax"
+
+	"unikraft.com/cloud/sdk/platform"
+	"unikraft.com/cloud/sdk/platform/group"
+	"unikraft.com/cloud/sdk/sandbox"
```
nderjung: Boom! `unikraft.com/cloud/plugins/sandbox` <3

internal/cmd/shell.go
dragosgheorghioiu: builtins can now be called only by using `:` and can only be at the start of the line
they are highlighted differently

internal/cmd/sandbox.go
dragosgheorghioiu: huuuuge :tada:

internal/cmd/instances.go
```diff
@@ -1350,6 +1362,16 @@ func (Instance) Examples() map[cmd.CmdType][]kingkong.Example {
...
 				},
 			},
+			{
+				Description: "Create an instance with a plugin",
+				Commands: []string{
+					`unikraft instance create \
+	  --name demo-instance \
+	  --metro fra \
+	  --image nginx:latest \
+	  --plugin name=sandbox,rom=sandbox:latest`,
```
nderjung:
```
--plugin name=sandbox,rom=plugins/sandbox:latest
```
?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,797 @@
...
+
+	ExecOpts
+}
+
+func (cmd ExecSandboxInstanceCmd) Examples() []kingkong.Example {
+	return []kingkong.Example{
+		{
+			Description: "Run a command on a sandbox instance",
+			Commands: []string{
+				"unikraft instance exec my-instance --plugin sandbox -- echo hello",
```
nderjung, line 57: I wonder if we can derive the default `--plugin <name>` for sandbox from https://github.com/unikraft-cloud/openapi/blob/prod-staging/plugins/sandbox.yaml#L838-L839 (and image)

internal/cmd/sandbox_test.go
dragosgheorghioiu: don't know what base image to use for testing the plugins because they need to use linux

i have some tests using base-compat but they still don't cover exec should i also push a base alpine or debian image to the plugins repo (or even somewhere else)?

internal/cmd/sandbox.go
dragosgheorghioiu: done

i think the sdk exports only the name for now, i can also add suport for it to export the image as well
happy to add it as it may also enable stuff like `--sandbox` or `--plugin sandbox` in the create instance command

internal/cmd/shell_builtins.go
dragosgheorghioiu: i would keep the shell_builtins in cmd because they are closer to their equivalents from `instances.go`

i did move some shell specific logic out of it though

jedevc: I still think they're worth moving. Importantly - imagine if we add the shell to the TUI? Or want to put it in the console?

Then, it makes sense to keep it separate.

dragosgheorghioiu: that's fair will try to move them out then

internal/cmd/sandbox_test.go
dragosgheorghioiu: now building a small busybox based image for testing exec

internal/cmd/shell.go
dragosgheorghioiu: not sure how to handle cd and export because they technically run locally

for example:
`:cd` - runs locally and changes the cwd for the following commands
`cd`  - runs on the instance itself and effects only the current command

jedevc: oh 🙁 hm, `cd` then needs to be different. although, argh, this works really weirdly for things like `echo foo; cd bar; pwd`.<br><br>make an exception for `cd` for now, but as I've mentioned before, we need to do some real shell parsing to properly solve this.

dragosgheorghioiu: but then it would be really unclear where the cd is being run

jedevc: Yeah, exactly, but that's the problem with this approach.

IMO, just pick something that feels sensible for now, this is experimental anyways — we should make sure there's a log message on all sandbox commands warning that this is experimental. I'm confident these problems *are* solvable, but the core problem is that we are trying to handle some builtins ourselves, and sending others to the remote `sh` — this split responsibility is what's causing all this. The solution is to remove the split:

* Either we use mvdan/sh and have all the commands + builtins evaluated directly by us.
* Or we open a login shell on the remote, and just stream all our commands there, so there's only one command running, then there's `cd` persistence there (but we then lose the ability to know the working directory at all, and a bunch of other shiny things).

dragosgheorghioiu: maybe the "different mode" idea would work better for now

PR #437, opened 2026-08-04
fix(images): Accept http(s)+oci image references

craciunoiuc:
Accepts `http+oci://` / `https+oci://` image references, so an OCI layout served over HTTP works wherever a registry reference does. The node is what fetches it, so the CLI carries the URI through verbatim.

The reference grammar and normalization move to `unikraft.com/x/image-spec/imageref`, replacing the CLI's `internal/x/reference` fork.

Depends on: https://github.com/unikraft-cloud/x/pull/368

GitHub-Fixes: TOOL-1093


internal/images/images.go
```diff
@@ -80,18 +81,58 @@ func WithInsecureRegistries() AccessorOpt {
...
+		return ParsedRef{Named: named, Scheme: imagespec.URISchemeOCI}, err
+	}
+
+	switch uri.Scheme {
+	case imagespec.URISchemeOCI:
+		named, err := parseNamedMetro(metro, uri.Path)
+		return ParsedRef{Named: named, Scheme: uri.Scheme}, err
+
+	case imagespec.URISchemeHTTPOCI, imagespec.URISchemeHTTPSOCI:
+		ref, err := uri.HTTPRef()
```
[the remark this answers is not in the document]
craciunoiuc: Will be rebased

cmd/unikraft/integration/instance_http_oci_test.go
```diff
@@ -0,0 +1,120 @@
...
+		assert.Regexp(t, `unexpected terminating '/'`, out)
+	})
+
+	// The end-to-end case needs a layout server the metro itself can reach, so
+	// it cannot be stood up from here. Point it at one to run it.
+	t.Run("live", func(t *testing.T) {
+		image := os.Getenv("UKC_TEST_HTTP_OCI_URI")
+		if image == "" {
+			t.Skip("set UKC_TEST_HTTP_OCI_URI to an http+oci:// image reachable from the metro")
+		}
```
jedevc: We should use the approach as in https://github.com/unikraft-cloud/integration-tests/pull/109.

Where we host the built image itself *inside* a VM, and then the agent pulls it from that place.

Then we don't need to be dependent on setting this here.

internal/types/image.go
```diff
@@ -42,9 +48,17 @@ func (ir ImageRef[T]) Render(opts value.RenderOpts) (string, error) {
...
+		return "", err
+	}
+	s := string(text)
 	if opts.Short {
-		s, _, _ = strings.Cut(s, "@")
+		if i := strings.LastIndex(s, "/@"); i >= 0 {
+			s = s[:i]
+		} else {
+			s, _, _ = strings.Cut(s, "@")
+		}
```
jedevc: Hmm. Despite having written the original version here, this does not spark joy.

We shouldn't be doing string manipulation, instead, we should be actually manipulating the `Reference`.

internal/types/image.go
```diff
@@ -54,21 +68,43 @@ func (ir ImageRef[T]) Value() any {
...
+func (ir ImageRef[T]) WireURL() string {
+	if ir.URI != "" {
+		return ir.URI
+	}
+	var zero T
+	if ir.Reference == zero {
+		return ""
+	}
+	return ir.Reference.String()
+}
```
jedevc: Replace with `MarshalText`?

internal/images/images.go
```diff
@@ -80,18 +81,58 @@ func WithInsecureRegistries() AccessorOpt {
...
+	Named reference.Named
+
+	// Scheme is the URI scheme the identifier carried, defaulting to oci for a
+	// bare reference.
+	Scheme imagespec.URIScheme
+
+	// URI is the verbatim identifier when Scheme is not oci, so that such
+	// identifiers round-trip back to the platform unchanged. It is empty for
+	// plain OCI references, which are sent as a bare reference.
+	URI string
```
jedevc: Hm, struggling to articulate what's wrong with this. The first thing is that this type and parsing probably be in `image-spec/reference`. So that it's centralized, instead of needing it custom in the agent as well eventually (so we can split out https://github.com/unikraft-cloud/agent/blob/7301eb51d13162c1b05ff6a57853f4b4c1c22da2/internal/oci/uri.go#L70-L86).

We should also really be attempting to parse the magic tarball reference into constituent parts here, not just treating it as opaque.

I'm also divided on which of these approaches we might want to take:

- Try and make an `httpRef` implementation. Then it'll work with `reference.WithDigest`, `reference.Domain` and the whole host of *things*. It would behave entirely as a normal reference, except that `String` would return the scheme prefix as well.

  Some helpers just wouldn't really work with this, especially the way this all is manipulated.

- (more likely) Make a single new `Reference` struct. Then it's very type-safe, very demure. Then we define a bunch of helpers on in, so that you can easily get the `Digest`, or remove the `Digest` for the short display above. Something in the style of https://github.com/dagger/dagger/blob/8c6516e3257650b055360a7a19672a674fa34714/internal/buildkit/util/gitutil/git_url.go#L33-L56 (with similar helpers like `fromURL`/`fromSCPStyleURL` instead being `fromReference` and `fromHTTPReference`).

  Then across x and the CLI, we'd *just* use this new `Reference` struct, and avoid using `reference.Named` and similar everywhere.

I am really struggling to think of the best approach, but importantly:
- As much of the handling code should be shared, callers ideally shouldn't need to distinguish between whether it's an HTTP reference or a real OCI reference (although they *should* be able to somehow).
- All the parsing logic for references should live in `x` so we can share it in the `agent` at some point.

craciunoiuc: :+1: on the last two bullets

:+1: on using `demure`

Yup, of the two the second bullets seems more plausible

craciunoiuc:
All comments legitimate, will address

craciunoiuc:
addressed

PR #438, opened 2026-08-05
feat(filters): Allow better relative/absolute time comparisons

jedevc:
Fixes https://linear.app/unikraft/issue/TOOL-1294/support-relative-time-comparisons.

See https://github.com/unikraft-cloud/docs/pull/251 for docs.

### DIFF
--- a/internal/resource/cmd/cmd_test.go
+++ b/internal/resource/cmd/cmd_test.go
@@ -41,6 +41,7 @@ func setupTestEnv() *resourcet.TestEnv {
 		URL:       "https://example.com",
 		Hidden:    "hidden-test1",
 		Invisible: "invisible-test1",
+		Created:   time.Now().Add(-3 * 24 * time.Hour),
 		Settings: resourcet.TestSettings{
 			Foo:   42,
 			Bar:   "hello",
@@ -60,6 +61,7 @@ func setupTestEnv() *resourcet.TestEnv {
 		URL:       "https://example.org",
 		Hidden:    "hidden-test2",
 		Invisible: "invisible-test2",
+		Created:   time.Now().Add(-40 * 24 * time.Hour),
 		Settings: resourcet.TestSettings{
 			Foo:   7,
 			Bar:   "world",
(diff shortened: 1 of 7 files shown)


internal/resource/cmd/cmd_test.go
```diff
@@ -2214,4 +2217,87 @@ func TestFilterComparisonOperators(t *testing.T) {
...
+
+	t.Run("relative_time_greater_equal_weeks", func(t *testing.T) {
+		output, err := runFilter(t, "created>=5w")
+		require.NoError(t, err)
+		assert.Contains(t, output, "test2")
+		assert.NotContains(t, output, "test1")
+	})
+
+	t.Run("relative_time_less_equal_compound_units", func(t *testing.T) {
+		output, err := runFilter(t, "created<=3d1h")
```
khatibomar, line 2244: Maybe add a test with `<=3d` exactly

internal/x/time/time.go
```diff
@@ -0,0 +1,53 @@
...
+// be more lenient about the input formats they accept.
+package time
+
+import (
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/araddon/dateparse"
```
khatibomar, line 16: last time this was updated 5 years ago...

jedevc: Indeed, but IMO, it's feature complete. Is it missing anything else?

jedevc: If there's a better alternative, happy to try that.

khatibomar: Not sure about replacements with some search, there is this one https://github.com/itlightning/dateparse which is updated last year.

internal/resource/cmd/cmd_test.go
```diff
@@ -2214,4 +2217,87 @@ func TestFilterComparisonOperators(t *testing.T) {
...
+	cutoff := time.Now().Add(-10 * 24 * time.Hour).UTC()
+
+	t.Run("absolute_time_rfc3339", func(t *testing.T) {
+		output, err := runFilter(t, "created>"+cutoff.Format(time.RFC3339))
+		require.NoError(t, err)
+		assert.Contains(t, output, "test1")
+		assert.NotContains(t, output, "test2")
+	})
+
+	t.Run("absolute_time_date_time_no_zone", func(t *testing.T) {
```
khatibomar, line 2268: I wonder if we should go crazy with datetime, like what date format we accept.

Someone could use UTC for example, others may use different time formats, should we go crazy and test some of these cases 🤔

internal/resource/cmd/cmd_test.go
jedevc: I suspect this will not work - given tiny variations in run durations, isn't that going to be impossible to hit exactly?

khatibomar: Ham, yea.

When test reaches the Parse method the duration would have shifted.

internal/resource/cmd/cmd_test.go
khatibomar: I wonder how we are dealing with time in database, are the created_at stored as UTC by default?

I think there should be some conversion happening between the time passed in by cli/frontend when hitting the API into UTC, then backend query the db in UTC and return the result as what ever timezone the client is in.

So the client can then specify what ever they want in time filters

khatibomar: Actually I think what is happening here is different, we filtering the frontend itself, which means we return all the data, then we filter them before rendering.

So all what I commented above is not applicable anw, we just filtering against assuming what is local to the client seeing data.

craciunoiuc (approved):
Change actually makes sense

But I see you've been putting them in `internal/x` would we have a use for this in other places? Aka moving it to the X repository.

Does the console or maybe kraftlet parse correctly times that are like these? If yes then no sense to move I guess?

```
```

internal/x/time/time.go
jedevc: Mm, significantly less popular. I would rather stick with the old one, we can always move to a fork if we find issues.

internal/resource/cmd/cmd_test.go
jedevc: Yes, this is all just client-side filtering for now.

Some filters can be turned into server-side filters (e.g. if filtering by tag), we just currently aren't :sweat:

jedevc:
> But I see you've been putting them in `internal/x` would we have a use for this in other places? Aka moving it to the X repository.

The things in `internal/x` absolutely could go in `x` - generally I've been using them as extensions/copies of other libraries that are missing functionality. So maybe they should go in `x/ext`? Although I'm not sure how widely useful they are outside of the CLI rn.

I'll leave this to a follow-up discussion.

PR #445, opened 2026-08-12
fix(tools/gendocs): Correctly escape mdx

jedevc:
See the thread here: https://github.com/unikraft-cloud/docs/pull/252#discussion_r3766794234.

We should fix the CLI not to generate invalid mdx.

### DIFF
--- a/tools/gendocs/mdx.go
+++ b/tools/gendocs/mdx.go
@@ -85,12 +85,12 @@ func generateMarkdown(ctx context.Context, node *kong.Node, dir string) error {
 
 	if node.Parent == nil {
 		if help != "" {
-			buf.WriteString(help + "\n\n")
+			buf.WriteString(escapeMdx(help) + "\n\n")
 		}
 	} else if node.Detail != "" {
-		buf.WriteString(node.Detail + "\n\n")
+		buf.WriteString(escapeMdx(node.Detail) + "\n\n")
 	} else if help != "" {
-		buf.WriteString(help + "\n\n")
+		buf.WriteString(escapeMdx(help) + "\n\n")
 	}
 
 	if IsRunnable(node) {
@@ -125,7 +125,7 @@ func generateMarkdown(ctx context.Context, node *kong.Node, dir string) error {
 		if err != nil {
 			return err
 		}
-		fmt.Fprintf(buf, "* [`%s`](%s): %s\n", relatedName, link, related.Help)
+		fmt.Fprintf(buf, "* [`%s`](%s): %s\n", relatedName, link, escapeMdx(related.Help))
 	}
 	if hasSeeAlso {
 		buf.WriteString("\n")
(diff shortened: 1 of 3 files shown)


tools/gendocs/mdx.go
```diff
@@ -85,12 +85,12 @@ func generateMarkdown(ctx context.Context, node *kong.Node, dir string) error {
...
 		if help != "" {
-			buf.WriteString(help + "\n\n")
+			buf.WriteString(escapeMdx(help) + "\n\n")
 		}
 	} else if node.Detail != "" {
-		buf.WriteString(node.Detail + "\n\n")
+		buf.WriteString(escapeMdx(node.Detail) + "\n\n")
 	} else if help != "" {
-		buf.WriteString(help + "\n\n")
+		buf.WriteString(escapeMdx(help) + "\n\n")
```
craciunoiuc, line 93: Just a wild thought, would it break smth if we were to make the `bufWriteString` always call `escapeMdx` somehow?

Like that we don't need to think if escaping applies there also

tools/gendocs/mdx_escape.go
```diff
@@ -0,0 +1,80 @@
...
+	ch := fence[0]
+	n := 0
+	for n < len(trimmed) && trimmed[n] == ch {
+		n++
+	}
+	return n == len(trimmed) && n >= len(fence)
+}
+
+// escapeMdxLine escapes a single line, skipping over inline code spans.
+func escapeMdxLine(line string) string {
```
craciunoiuc, line 57: Normally I'll say to inline this, but the code change is pretty big so not worth it no?

tools/gendocs/mdx_escape.go
```diff
@@ -0,0 +1,80 @@
...
+
+// closesFence reports whether trimmed closes the given fence.
+func closesFence(trimmed, fence string) bool {
+	ch := fence[0]
+	n := 0
+	for n < len(trimmed) && trimmed[n] == ch {
+		n++
+	}
+	return n == len(trimmed) && n >= len(fence)
+}
```
craciunoiuc, line 54: same here, I like them more as functions but normally we would inline them

tools/gendocs/mdx_escape_test.go
```diff
@@ -0,0 +1,62 @@
...
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEscapeMdx(t *testing.T) {
+	tests := []struct {
```
craciunoiuc, line 15: There's quite a bit of parsing going on in the escape file, maybe some unit tests for the subfunctions used also? 🤔

craciunoiuc (commented):
Just some comments

tools/gendocs/mdx_escape_test.go
jedevc: Ideally not. It's only 80 lines of parsing code, we don't need more tests right? Testing all of the subfunctions just makes it more annoying to refactor the code.

tools/gendocs/mdx_escape.go
jedevc: I've kept them as functions, I think it reads easier.

tools/gendocs/mdx_escape.go
jedevc: Done.

tools/gendocs/mdx.go
jedevc: Maybe at some point? I think this would likely break something, and is a much larger change - happy for you to attempt this later, but I want to unblock the docs deploys.

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #447, opened 2026-08-13
feat: Add arm version of build and publish roms

dragosgheorghioiu:


internal/builder/build.go
```diff
@@ -53,8 +53,33 @@ type FSOpts struct {
...
+// SetArch overrides the architecture of every platform to build for. When no
+// platform is set yet, one is added for the default OS.
+func (o *BuildOpts) SetArch(arch string) {
+	if len(o.Platform) == 0 {
+		o.Platform = []ocispec.Platform{{Architecture: arch, OS: DefaultOS}}
+		return
+	}
+	for i := range o.Platform {
+		o.Platform[i].Architecture = arch
+	}
```
jedevc: IMO, instead of this, it should act as a filter.

So:
- No `--arch` specified, use `o.Platform` as is
- If no `o.Platform` exists, fallback to `DefaultPlatform` (maybe even with an info log line)

Otherwise:
- If `--arch` specified, filter `o.Platform` to only the targets that have that architecture. So if you're building both `fc/x86_64` and `qemu/x86_64` and you specify `--arch x86_64`, then you get both of em, but it would exclude `fc/arm64`.
- If no `o.Platform` exists, fallback to `DefaultPlatform` (with specified `--arch`).

Does that make sense 😅

internal/builder/build.go
```diff
@@ -53,8 +53,33 @@ type FSOpts struct {
...
+func NormalizeArch(arch string) (string, error) {
+	switch strings.ToLower(arch) {
+	case "x86_64", "x86-64", "amd64":
+		return "x86_64", nil
+	case "arm64", "aarch64":
+		return "arm64", nil
+	default:
+		return "", fmt.Errorf("unsupported architecture %q: expected x86_64 or arm64", arch)
+	}
+}
```
jedevc: We can use https://github.com/unikraft-cloud/x/pull/390.

internal/cmd/build.go
```diff
@@ -23,8 +23,10 @@ import (
 )
 
 type ImageBuildCmd struct {
-	Input  string `arg:"" default:"." help:"Path to the input directory."`
-	Output string `short:"o" help:"Output destination"`
+	Input   string `arg:"" default:"." help:"Path to the input directory."`
+	Output  string `short:"o" help:"Output destination"`
+	Arch    string `help:"Architecture to build for (x86_64 or arm64). Defaults to the Kraftfile targets, or x86_64."`
+	Replace bool   `help:"Publish only the architecture that was built, dropping any others already published under the destination tag."`
```
craciunoiuc: If you ask me I would reverse this?

By default it should replace and if asked it should append to an artifact

internal/builder/build.go
dragosgheorghioiu: oh yeah i think i understand so it should filter depending on what it can find in the kraftfile targets otherwise default to kraftcloud

internal/cmd/build.go
dragosgheorghioiu: so i replaced with an append flag but a multiarch build with a Dockerfile rootfs would look like this
```
unikraft build . --arch x86_64 -o org/debian-ssh:latest
unikraft build . --arch arm64  -o org/debian-ssh:latest --append
```
not sure if replace should be the default

craciunoiuc: hmm, then why not just `unikraft build . -o org/debian-ssh:latest`

Under the hood it would fetch all permutations and then build for each (and append each time)

internal/builder/build.go
jedevc: Exactly.

internal/builder/rootfs.go
```diff
@@ -106,6 +106,11 @@ func DetectSourceType(path string) (kraftfile.SourceType, error) {
 }
 
 func BuildRoms(ctx context.Context, opts BuildOpts) (_ []imagespec.File, rerr error) {
+	plats := opts.Platform
+	if len(plats) == 0 {
+		plats = []ocispec.Platform{DefaultPlatform}
+	}
```
jedevc, line 118: Add a log here too...

internal/cmd/build.go
```diff
@@ -92,6 +100,16 @@ func (c *ImageBuildCmd) Run(ctx context.Context, cfg *config.Config, sandbox *re
 		return err
 	}
 
+	if c.Arch != "" {
+		arch, err := imagespec.NormalizeArch(c.Arch)
+		if err != nil {
+			return fmt.Errorf("%w: expected %s or %s", err, imagespec.ArchitectureIntel, imagespec.ArchitectureArm)
```
jedevc, line 106:
```suggestion
			return err
```

internal/cmd/build.go
```diff
@@ -25,6 +25,8 @@ import (
 type ImageBuildCmd struct {
 	Input  string `arg:"" default:"." help:"Path to the input directory."`
 	Output string `short:"o" help:"Output destination"`
+	Arch   string `help:"Only build the Kraftfile targets of this architecture (x86_64 or arm64). Defaults to every target, or x86_64 when none are declared."`
+	Append bool   `help:"Keep the platforms already published under the destination tag instead of publishing only what was built."`
```
jedevc: This is racy right? If I run two jobs here, and they both try to push at the same time, they're gonna overwrite each other.

Dunno the solution, but maybe we allow `unikraft image copy` to take multiple sources? Then we can push to separate tags, and then create a multi-platform image. This is a reasonably common approach.

(or we make multi-platform builds actually work, so you just need one `unikraft build` to solve it)

jedevc: Mm, I will make the latter work. So `unikraft build` just does it.

dragosgheorghioiu: my idea was to just run buildkit for every arch maybe that can avoid **that** bug

jedevc: Okay, here we go. https://github.com/unikraft-cloud/cli/pull/448

(yeah it's exactly that solution)

craciunoiuc (commented):
this is okay now my side

leaving you two to resolve your last comments

internal/cmd/build.go
```diff
@@ -23,8 +24,10 @@ import (
 )
 
 type ImageBuildCmd struct {
-	Input  string `arg:"" default:"." help:"Path to the input directory."`
-	Output string `short:"o" help:"Output destination"`
+	Input  string   `arg:"" default:"." help:"Path to the input directory."`
+	Output string   `short:"o" help:"Output destination"`
+	Arch   []string `help:"Only build the Kraftfile targets of these architectures (x86_64, arm64). Defaults to every target, or x86_64 when none are declared."`
+	Append bool     `help:"Keep the platforms already published under the destination tag instead of publishing only what was built."`
```
jedevc: Do we need this now that we just support doing `unikraft build` with multiple platforms?

dragosgheorghioiu: i see it more as a "nice to have" but it's not mandatory now
i can see some specific situation where you just want to append instead of overwriting it completely

i can remove it but i don't see a problem with leaving it here

jedevc: Let's remove it, since it's racy. I don't think it's something we shouldn't have ever, but at least should be a separate PR, I want to think about it.

dragosgheorghioiu: removed the flags from `build` and `image copy`

internal/cmd/build.go
```diff
@@ -92,6 +100,19 @@ func (c *ImageBuildCmd) Run(ctx context.Context, cfg *config.Config, sandbox *re
...
+	arches := make([]string, 0, len(c.Arch))
+	for _, a := range c.Arch {
+		arch, err := imagespec.NormalizeArch(a)
+		if err != nil {
+			return fmt.Errorf("%w: expected %s or %s", err, imagespec.ArchitectureIntel, imagespec.ArchitectureArm)
+		}
+		arches = append(arches, arch)
+	}
+	slices.Sort(arches)
+	if err := buildOpts.FilterArch(slices.Compact(arches)...); err != nil {
```
jedevc: Hm, `slices.Compact` without `slices.Sort` doesn't remove all duplicates...

I don't think we actually need this call though?

internal/builder/build.go
```diff
@@ -105,16 +157,24 @@ func Build(ctx context.Context, opts BuildOpts) ([]*imagespec.Image, error) {
 	// set of platforms we're building for.
 	var kernels []*imagespec.Image
 	if opts.Runtime != "" {
-		var err error
-		kernels, err = BuildKernel(ctx, opts)
+		// An empty platform set means neither the Kraftfile nor the command line
+		// narrowed the build, so the runtime alone decides what we get.
+		asked := len(opts.Platform) > 0
```
jedevc: inline this.

internal/builder/build.go
```diff
@@ -105,16 +157,24 @@ func Build(ctx context.Context, opts BuildOpts) ([]*imagespec.Image, error) {
...
 		defer func() {
-			for _, kernel := range kernels {
+			for _, kernel := range loaded {
 				kernel.Close()
 			}
 		}()
+
+		kernels = loaded
+		if !asked && len(kernels) > 1 {
+			kernels = defaultArchKernels(ctx, loaded)
```
jedevc: inline the helper?

also this needs an explanation of *why* it's done like this. it's because we don't want to silently start building a multiarch build right? so that when `base-compat` gets updated, that we still only build for x86_64 instead of also silently including arm64 right? Which would be a regression?

I had to spend quite a chunk of time trying to decipher why we needed this.

internal/builder/build.go
```diff
@@ -54,7 +55,58 @@ type FSOpts struct {
...
+		return err != nil || !slices.Contains(arches, norm)
+	})
+	if len(o.Platform) == 0 {
+		return fmt.Errorf("no target with architecture %s: targets are %s",
+			strings.Join(arches, ", "), strings.Join(available, ", "))
+	}
+	return nil
+}
+
+func defaultArchKernels(ctx context.Context, kernels []*imagespec.Image) []*imagespec.Image {
```
jedevc: Maybe `filterKernels(kernels, arch)`.

internal/builder/kernel.go
```diff
@@ -39,24 +40,51 @@ func BuildKernel(ctx context.Context, opts BuildOpts) ([]*imagespec.Image, error
...
-		if _, ok := imgPlatforms[platforms.Format(platforms.Normalize(p))]; !ok {
-			missingPlatforms = append(missingPlatforms, platforms.Format(p))
+
+	return imgs, nil
+}
+
+// missingPlatforms reports the requested platforms that the runtime does not
+// provide, naming the ones it does so that both a partial and an empty match
+// read the same way.
+func missingPlatforms(runtime string, requested []ocispec.Platform, available []*imagespec.Image) error {
```
jedevc: I'm struggling with this one, sorry 😓

Why do we need the changes in this file? A commit message or something explaining this would be useful :(

Is it just about making the error messages consistent? Or is there an actual functionality change?

internal/builder/kernel.go
```diff
@@ -39,24 +40,51 @@ func BuildKernel(ctx context.Context, opts BuildOpts) ([]*imagespec.Image, error
 
 	imgs, err := access.LoadAll(ctx, runtime, platform)
 	if err != nil {
+		if len(opts.Platform) > 0 {
```
jedevc: Only go down this flow if the error is containerd's `errdefs.ErrNotFound`, we shouldn't go here if we failed to pull images due to a network error.

internal/builder/build.go
```diff
@@ -105,16 +157,24 @@ func Build(ctx context.Context, opts BuildOpts) ([]*imagespec.Image, error) {
 	// set of platforms we're building for.
 	var kernels []*imagespec.Image
 	if opts.Runtime != "" {
-		var err error
-		kernels, err = BuildKernel(ctx, opts)
+		// An empty platform set means neither the Kraftfile nor the command line
+		// narrowed the build, so the runtime alone decides what we get.
```
jedevc: ai comment 😢

internal/cmd/build.go
dragosgheorghioiu: removed

internal/builder/build.go
dragosgheorghioiu: it was meant to not start more than one build by default unless specified with the --arch flag 
also inlined

internal/builder/build.go
dragosgheorghioiu: inlined

internal/builder/kernel.go
dragosgheorghioiu: on single load_all now
and apply the filters after that

internal/builder/kernel.go
dragosgheorghioiu: it was meant for the situation in which the runtime does not provide the requested arch for the build to print a clean message to indicate the mismatch

jedevc (approved):
```
```

PR #450, opened 2026-08-14
fix(build)!: Remove default target

jedevc:
We're now a multi-platform shop: we support amd64 and also aarch64.

So, to support them equally - no more build defaults. We now *require* that when building an image the user has specified an architecture:
- For building an image, very little changes if users have a `runtime`, we still default to using all of them. Users can still override it using the `targets:` key (or the new `--arch` flag).
- For building a rom, users will now need to set the `targets:` key (or `--arch`) *explicitly*.

This is a breaking change, but this was originally done under my previous (wrong) assumption that ROMs would be platform-independent. They really aren't, so I think the breaking change is justified, and creates a simpler-to-understand config.

Before:

```
spec: v0.7
roms:
  - ./Dockerfile
```

After:

```
spec: v0.7
roms:
  - ./Dockerfile
targets:
  - kraftcloud/x86_64
```

### DIFF
--- a/cmd/unikraft/integration/build_test.go
+++ b/cmd/unikraft/integration/build_test.go
@@ -71,7 +71,7 @@ roms:
 			).Apply(dir))
 
 			r.Run(t, []string{"unikraft", "build", "base", "--output", baseImage}, integ.WithWorkDir(dir))
-			r.Run(t, []string{"unikraft", "build", "rom", "--output", romImage}, integ.WithWorkDir(dir))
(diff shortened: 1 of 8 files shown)


dragosgheorghioiu (dismissed):
LGTM

```
```

craciunoiuc (dismissed):
All good here. Thanks!

```
```

craciunoiuc (dismissed):
I agree even harder now.

```
```

jedevc:
@dragosgheorghioiu I will merge this, but we also need docs updates right?

dragosgheorghioiu:
the examples are all updated but yeah should do another pass through all the docs

PR #456, opened 2026-08-17
feat: Add build context support

### LINKED ISSUE #371: Add support for `--build-context` argument to `unikraft build`
We already support for most of the same args as `docker buildx build` - we should make it complete, I think the big remaining one is `unikraft build --build-context`. See **https://docs.docker.com/build/concepts/context/.

For implementation details, this is originally done in buildx around here: https://github.com/docker/buildx/blob/3e413a05dfef87849f252fc85be7e07f32d75b99/build/build.go#L112-L112

We'd need to grab some of this parsing logic (maybe via imports, maybe vendoring if we can't), then making sure we support setting build contexts.

Also we should have tests around the rest of the build tests.

Internal issue: https://linear.app/unikraft/issue/TOOL-831/add-support-for-build-context-to-unikraft-build

Omkar-Ugal:
## Summary

Add support for `--build-context` to `unikraft images build`.

This allows named build contexts to be passed to BuildKit and used with
`COPY --from=<context>` in Dockerfiles.

Closes #371

### DIFF
--- a/internal/builder/context.go
+++ b/internal/builder/context.go
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package builder
+
+import (
+	"strings"
+
+	"github.com/distribution/reference"
+	"github.com/pkg/errors"
+)
+
+// ParseContextNames parses --build-context values in the form name=value.
+func ParseContextNames(values []string) (map[string]string, error) {
+	if len(values) == 0 {
+		return nil, nil
+	}
+
+	result := make(map[string]string, len(values))
+	for _, value := range values {
+		if value == "" {
+			continue
+		}
+
+		kv := strings.SplitN(value, "=", 2)
+		if len(kv) != 2 {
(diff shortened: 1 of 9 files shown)


internal/builder/context.go
```diff
@@ -0,0 +1,49 @@
...
+
+		named, err := reference.ParseNormalizedNamed(kv[0])
+		if err != nil {
+			return nil, errors.Wrapf(err, "invalid context name %s", kv[0])
+		}
+
+		name := strings.TrimSuffix(
+			reference.FamiliarString(named),
+			":latest",
+		)
```
jedevc: Buildkit also supports build-contexts that are local directories? Or git repositories? We shouldn't just support images.

internal/builder/context_test.go
```diff
@@ -0,0 +1,70 @@
...
+			want:   map[string]string{},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := ParseContextNames(tt.values)
+
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("ParseContextNames() error = %v, wantErr %v", err, tt.wantErr)
```
jedevc: Why no use of `require`/`assert`?

Omkar-Ugal: updated

jedevc:
This needs build integration tests that these do work :pray:

PR #458, opened 2026-08-19
feat: Sandbox integration

dragosgheorghioiu:
Closes: TOOL-762


cmd/unikraft/examples_test.go
```diff
@@ -40,6 +41,16 @@ func TestExamples(t *testing.T) {
...
 
+	// Create files used by the write example (type:existingfile) in a temp dir.
+	tmp := t.TempDir()
+	origDir, err := os.Getwd()
+	require.NoError(t, err)
+	require.NoError(t, os.Chdir(tmp))
+	t.Cleanup(func() { require.NoError(t, os.Chdir(origDir)) })
+	for _, name := range []string{"config.json", "data.bin"} {
+		require.NoError(t, os.WriteFile(filepath.Join(tmp, name), nil, 0o644))
+	}
```
jedevc: 🤔 bit weird that these are hardcoded.

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,582 @@
...
+
+const (
+	// sandboxPlugin is the plugin name the sandbox commands address by default,
+	// and sandboxPluginRom is the ROM serving it.
+	sandboxPlugin = "sandbox"
+
+	// TODO: the ROM the plugin itself publishes, once it carries the cwd and
+	// environment support the exec subtests need. The one it serves today
+	// accepts both and ignores them.
+	sandboxPluginRom = "dragosgheorghioiu/sandbox-rom"
```
jedevc: Is this merged yet?

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,582 @@
...
+	// Copy specifications are resolved before anything is sent to an instance,
+	// so these need no API.
+	t.Run("copy-spec", func(t *testing.T) {
+		t.Run("neither-names-an-instance", func(t *testing.T) {
+			r := runner(t, false, []string{staging, stable, prod})
+
+			out := r.Run(t, []string{"unikraft", "instance", "copy", "./a.txt", "./b.txt"}, integ.ExpectFail())
+			assert.Regexp(t, `neither "\./a\.txt" nor "\./b\.txt" names an instance`, out)
+			assert.Regexp(t, `written as <instance>:<path>`, out)
+		})
```
jedevc: IMO, I don't think we really need all these negative tests. Let's just focus on the positive tests that make sure that the behavior correctly works.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+	"unikraft.com/cli/internal/multimetro"
+)
+
+type ExecOpts struct {
+	Cmd []string `arg:"" name:"command" help:"Command to pass to the instance." placeholder:"cmd"`
+
+	Plugin      string            `name:"plugin" help:"Plugin name from the instance to run the command onto"`
+	Dir         string            `name:"dir" help:"Directory to execute the command from"`
+	Env         map[string]string `name:"env"     help:"Environment variables to set (KEY=VALUE)" mapsep:","`
+	TimeoutMsec int               `name:"cmd-timeout" help:"Timeout for waiting the result of the command"`
```
jedevc: Instead of setting this, let's just rely on the global `--timeout` (which propagates to the context cancellation).

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/multimetro"
+)
+
+type ExecOpts struct {
+	Cmd []string `arg:"" name:"command" help:"Command to pass to the instance." placeholder:"cmd"`
+
+	Plugin      string            `name:"plugin" help:"Plugin name from the instance to run the command onto"`
+	Dir         string            `name:"dir" help:"Directory to execute the command from"`
+	Env         map[string]string `name:"env"     help:"Environment variables to set (KEY=VALUE)" mapsep:","`
```
jedevc: formatting.

jedevc: Let's make this work exactly the same as the instance create command. So alias to `-e` as well.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+}
+
+func (c *WriteSandboxInstanceCmd) Run(ctx context.Context, stdio config.Stdio) error {
+	target, err := resolveSandboxTarget(ctx, c.Target, c.Plugin)
+	if err != nil {
+		return err
+	}
+
+	remote, err := uploadSandboxFile(ctx, target.g, target.key, target.plugin, c.Local, c.Remote, c.Append, c.Parents)
```
jedevc: Too many args.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+	g, err := multimetro.NewClient(ctx)
+	if err != nil {
+		return sandboxTarget{}, err
+	}
+
+	return sandboxTarget{
+		key:    instance.Key().(multimetro.Key),
+		g:      g,
+		plugin: plugin,
+	}, nil
```
jedevc: Couldn't we just directly return the plugin `sandbox.Client`? Plugin names can be set there. It's already fully resolved.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+	return fmt.Errorf("%s: %w", msg, err)
+}
+
+func quoteShellArg(s string) string {
+	if quoted, err := syntax.Quote(s, syntax.LangBash); err == nil {
+		return quoted
+	}
+	return "'" + strings.ReplaceAll(s, "'", "'\\''") + "'"
+}
```
jedevc: Why the fallback?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+func decodeSandboxPayload(s string) []byte {
+	decoded, err := base64.StdEncoding.DecodeString(s)
+	if err != nil {
+		return []byte(s)
+	}
+	return decoded
+}
+
+func wrapSandboxErr(ctx context.Context, key multimetro.Key, plugin, msg string, err error) error {
+	if err != nil && strings.Contains(err.Error(), "request failed: 404") {
```
jedevc: What if the sandbox legitimately returns 404?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+		if execResp.Data == nil {
+			return struct{}{}, fmt.Errorf("failed to start command: the %q plugin did not report a command UUID", pluginName)
+		}
+		cmdUUID := execResp.Data.Uuid
+
+		defer func() {
+			if ctx.Err() == nil {
+				return
+			}
+			signalCtx, cancelSignal := context.WithTimeout(context.Background(), 5*time.Second)
```
jedevc: Still use `ctx`, but you can do `context.WithTimeout(context.WithoutCancel(ctx))`.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+		}()
+
+		if in != nil {
+			feedCtx, cancelFeed := context.WithCancel(ctx)
+			defer cancelFeed()
+			go feedSandboxStdin(feedCtx, c, instance, pluginName, cmdUUID, in)
+		}
+
+		var stdoutOffset, stderrOffset uint64
+		fetchAndPrint := func() error {
```
jedevc: This could be a separate helper.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,758 @@
...
+	target, filePath = spec[:i], spec[i+len(copyPathSeparator):]
+
+	// A target that opens like a filesystem path is a local file whose name
+	// happens to carry a colon.
+	if target == "" || strings.HasPrefix(target, "/") || strings.HasPrefix(target, ".") || strings.HasPrefix(target, "~") {
+		return "", spec
+	}
+
+	return target, filePath
+}
```
jedevc, line 323: This could use some tests.

jedevc (dismissed):

cmd/unikraft/integration/sandbox_test.go
dragosgheorghioiu: fixed this it was some stale change from the old pr where i was testing the new version of the sandbox

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,376 @@
...
+	"unikraft.com/cloud/plugins/sandbox"
+
+	integ "unikraft.com/cli/internal/integration"
+)
+
+const (
+	sandboxPlugin = sandbox.PluginName
+
+	sandboxPluginRom = "plugins/sandbox:staging"
+
```
craciunoiuc: Is this a public project la "official"?

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,376 @@
...
+	sandboxKraftfile = `
+spec: v0.7
+name: sandbox-e2e
+runtime: base-compat:latest
+rootfs:
+  format: erofs
+  source: ./Dockerfile
```
craciunoiuc, line 31: Add `type: dockerfile` for good measure to have the test focused/not flake in case of anything

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,376 @@
...
+	t.Helper()
+
+	dir := t.TempDir()
+	require.NoError(t, fstest.Apply(
+		fstest.CreateFile("Dockerfile", []byte("FROM busybox:latest\n"), 0o644),
+		fstest.CreateFile("Kraftfile", []byte(sandboxKraftfile), 0o644),
+	).Apply(dir))
+
+	// The image is registered in the test's resource sandbox, which deletes it
+	// when the test ends.
```
craciunoiuc: remove this I think

This is true for all resources in the framework thankfully

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,376 @@
...
+			instName := newSandboxInstance(t, r)
+			remote := "/sb-append-" + uniq() + ".txt"
+			dir := t.TempDir()
+			require.NoError(t, fstest.Apply(
+				fstest.CreateFile("first.txt", []byte("first line\n"), 0o644),
+				fstest.CreateFile("second.txt", []byte("second line\n"), 0o644),
+			).Apply(dir))
+
+			r.Run(t, []string{"unikraft", "instance", "write", instName, "./first.txt", remote}, integ.WithWorkDir(dir))
+			r.Run(t, []string{"unikraft", "instance", "write", instName, "./second.txt", remote, "--append"}, integ.WithWorkDir(dir))
```
craciunoiuc, line 158: I see a lot of tests for the sandbox itself, I wonder if they should sit in the integration framework in the future

Nothing for now I think, but we can migrate them some time later

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,376 @@
...
+			r.Run(t, []string{"unikraft", "instance", "read", instName, remote}, integ.WithWorkDir(dir))
+			fetched, err := os.ReadFile(filepath.Join(dir, filepath.Base(remote)))
+			require.NoError(t, err)
+			assert.Equal(t, "remote contents\n", string(fetched))
+
+			r.Run(t, []string{"unikraft", "instance", "read", instName, remote, "--force"}, integ.WithWorkDir(dir))
+
+			// A local path naming a directory is written into.
+			require.NoError(t, os.Mkdir(filepath.Join(dir, "into"), 0o755))
+			r.Run(t, []string{"unikraft", "instance", "read", instName, remote, "./into"}, integ.WithWorkDir(dir))
```
craciunoiuc, line 206: I've recently read a post where they discovered that more tools might actually degrade agent performance. And they saw an increase in scores from ~80->100% and from ~50%->60% just by cutting tools from 21 to 9. Basically the agent knew how to use `exec` and that's all it needed.

What I'm getting at: we should be careful to not add too many commands apart from `upload/download/read/write/execute/edit`, aka if the sandbox has `bash` it will work just fine.

cmd/unikraft/testdata/TestHelp/instances
```diff
@@ -34,6 +34,16 @@ Resources:
...
+  exec
+          Exec a command on a sandbox instance.
+  copy, cp
+          Copy a file to or from a sandbox instance.
+  write
+          Write a file to a sandbox instance.
+  read
+          Read a file from a sandbox instance.
+  mkdir
+          Create a directory on a sandbox instance.
```
craciunoiuc: To continue the yap above: I find mkdir a bit useless tbh

cmd/unikraft/testdata/TestHelp/instances
```diff
@@ -365,6 +379,13 @@ Examples:
   	  --metro fra \
   	  --template my-template
 
+  # Create an instance with a plugin
+  unikraft instance create \
+  	  --name demo-instance \
+  	  --metro fra \
+  	  --image nginx:latest \
+  	  --plugin name=sandbox,rom=plugins/sandbox:latest
+
```
craciunoiuc, line 390: Would it make sense to split into two separate commits:
 * add plugin support
 * add sanbox support

Maybe even stacked PRs

internal/cmd/testdata/TestOutput/instances
```diff
@@ -81,6 +81,7 @@ roms:
 - name:          my-rom
   image:         myuser/my-rom:latest
   at:            /rom
+plugins:
```
craciunoiuc, line 94: Is it ok for this to be empty here?

internal/cmd/instances.go
```diff
@@ -65,6 +66,12 @@ type InstancesCmd struct {
 	Restart InstancesRestartCmd `cmd:"" help:"Restart one or more instances."`
 	Tunnel  InstancesTunnelCmd  `cmd:"" aliases:"port-forward" help:"Forward a local port to an unexposed instance."`
 	History InstanceHistoryCmd  `cmd:"" help:"Show checkpoint history for an instance."`
+
+	Exec  ExecSandboxInstanceCmd  `cmd:"" help:"Exec a command on a sandbox instance."`
+	Copy  CopySandboxInstanceCmd  `cmd:"" aliases:"cp" help:"Copy a file to or from a sandbox instance."`
+	Write WriteSandboxInstanceCmd `cmd:"" help:"Write a file to a sandbox instance."`
+	Read  ReadSandboxInstanceCmd  `cmd:"" help:"Read a file from a sandbox instance."`
+	Mkdir MkdirSandboxInstanceCmd `cmd:"" help:"Create a directory on a sandbox instance."`
```
craciunoiuc: Should these maybe sit one level below?

`unikraft instance sandbox ...`

craciunoiuc: Bonus, I realised that this need to apply to other plugins also in this case, because you add it with `--plugin sandbox`

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+			Description: "Run a command on a sandbox instance",
+			Commands: []string{
+				"unikraft instance exec my-instance --plugin sandbox -- echo hello",
+			},
+		},
+		{
+			Description: "Run a command in a specific working directory",
+			Commands: []string{
+				"unikraft instance exec my-instance --plugin sandbox --dir /var/lib/app -- ls -la",
+			},
```
craciunoiuc: Continuing the point in the other comment. It makes things complcated because each plugin has a custom set of command right?

If yes, I think we should only support "official" plugins and everything else can be done through the API

As such this would become `"unikraft instance saandbox exec my-instance --dir /var/lib/app -- ls -la"`

Opinions @jedevc ?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+			return nil
+		}
+		loaded = append(loaded, p.Name)
+	}
+
+	if len(loaded) == 0 {
+		return fmt.Errorf("instance %q has no plugins loaded, so nothing answers to %q", instance.Name, plugin)
+	}
+	return fmt.Errorf("instance %q has no plugin named %q; it has: %s", instance.Name, plugin, strings.Join(loaded, ", "))
+}
```
craciunoiuc: These can be inlined no? Are they used anywhere else?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+
+const sandboxLogPollInterval = 100 * time.Millisecond
+
+func decodeSandboxPayload(s string) []byte {
+	decoded, err := base64.StdEncoding.DecodeString(s)
+	if err != nil {
+		return []byte(s)
+	}
+	return decoded
+}
```
craciunoiuc: This can also be inlined I think

Also, shouldn't we error out if the output is not decodable?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+
+func feedSandboxStdin(ctx context.Context, target sandboxTarget, cmdUUID string, in io.Reader) {
+	write := func(data string, eof bool) error {
+		req := sandbox.CommandStdinRequest{Data: data, Eof: &eof}
+		_, err := target.client.WriteCommandStdin(ctx, target.instance, cmdUUID, &req, target.opts...)
+		return err
+	}
+
+	buf := make([]byte, 32*1024)
```
craciunoiuc: Hardcoded value?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+	return nil
+}
+
+type downloadOpts struct {
+	remote string
+	local  string
+	force  bool
+}
+
+func downloadSandboxFile(ctx context.Context, target sandboxTarget, opts downloadOpts) (string, int, error) {
```
craciunoiuc: Maybe you could put all of these helpers into an `internal/sandbox/file.go` or smth and we would import that? What do you think?

You would create a sandbox object with a single element `target sandboxTarget` which is common to all of them. You could then do all operations on a sandbox target

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+		return nil, fmt.Errorf("failed to read file: the %q plugin returned no contents", target.plugin)
+	}
+
+	return decodeSandboxPayload(resp.Data.Contents), nil
+}
+
+const copyPathSeparator = ":"
+
+func parseCopyPath(spec string) (target, filePath string) {
+	i := strings.Index(spec, copyPathSeparator)
```
craciunoiuc, line 299: This needs tests for sure if it doesn't have it

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+				return "", spec
+			}
+			i += len(copyPathSeparator) + j
+			break
+		}
+	}
+
+	target, filePath = spec[:i], spec[i+len(copyPathSeparator):]
+
+	if target == "" || strings.HasPrefix(target, "/") || strings.HasPrefix(target, ".") || strings.HasPrefix(target, "~") {
```
craciunoiuc: try to use `filepath` and `os` for some of these?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+}
+
+type CopySandboxInstanceCmd struct {
+	Source      string `arg:"" name:"source" help:"File to copy from, either a local path or <instance>:<path>."`
+	Destination string `arg:"" name:"destination" help:"Where to copy it to, either a local path or <instance>:<path>."`
+
+	Plugin  string `name:"plugin" help:"Plugin name from the instance to copy through."`
+	Force   bool   `name:"force" help:"Overwrite the local file if it already exists."`
+	Parents bool   `name:"parents" help:"Create parent directories on the remote path if they don't already exist."`
+}
```
craciunoiuc: should have a short also of `-p`

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+			Description: "Copy a file to an instance in a specific metro, creating parent directories as needed",
+			Commands: []string{
+				"unikraft instance copy ./data.bin fra0/my-instance:/var/lib/app/data.bin --parents",
+			},
+		},
+	}
+}
+
+func (c *CopySandboxInstanceCmd) Run(ctx context.Context, stdio config.Stdio) error {
+	srcTarget, srcPath := parseCopyPath(c.Source)
```
craciunoiuc: how does download/upload/copy handle symlinks btw?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,774 @@
...
+	}
+}
+
+type MkdirSandboxInstanceCmd struct {
+	Target string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instance to create the directory on."`
+	Path   string `arg:"" name:"path" help:"Remote directory path to create."`
+
+	Plugin  string `name:"plugin" help:"Plugin name from the instance to create the directory on."`
+	Parents bool   `name:"parents" short:"p" help:"Create parent directories as needed."`
+}
```
craciunoiuc: `-p` here also

internal/integration/env.go
```diff
@@ -69,6 +70,11 @@ func WithWorkDir(dir string) CmdOption {
 	return func(c *cmdConfig) { c.workDir = dir }
 }
 
+// WithStdin feeds in to the command's standard input, which is otherwise empty.
+func WithStdin(in string) CmdOption {
+	return func(c *cmdConfig) { c.stdin = in }
+}
+
```
craciunoiuc, line 78: Separate commit for this change & co?

go.mod
```diff
@@ -1,6 +1,6 @@
 module unikraft.com/cli
 
-go 1.26.2
+go 1.26.4
```
craciunoiuc: Just bump it to 1.27.0 if it's not been done already

internal/resource/value/parse.go
```diff
@@ -18,6 +18,53 @@ import (
 	xmaps "unikraft.com/cli/internal/x/maps"
 )
 
+// splitTopLevel splits s on commas, keeping those inside balanced {} or []
+// pairs and inside JSON string literals, so a value carrying JSON survives.
+func splitTopLevel(s string) []string {
+	var parts []string
+	var buf strings.Builder
```
craciunoiuc: Separate commit for this change? Is this a fix? what is it for?

internal/multimetro/client.go
```diff
@@ -21,7 +23,33 @@ import (
...
+	// are made with. The plugin client is built from the platform client's
+	// options, which carry no HTTP client of their own, so it is handed to the
+	// plugin per call instead.
+	sandboxHTTPClient *http.Client
+}
+
+// SandboxOpts returns the options for a call against the named plugin on this
+// metro. An empty plugin name leaves the plugin the client was built with.
+func (c MetroClient) SandboxOpts(plugin string) []sandbox.Option {
+	opts := make([]sandbox.Option, 0, 2)
```
craciunoiuc: I am not 100% sure these should sit here, but not sure what to suggest either/ what you discussed already?

craciunoiuc (changes requested):
I left you a lot of comments which I think need to be discussed at least

They are general comments related to the whole work + refactorings

craciunoiuc:
you need to rebase also :flushed:

cmd/unikraft/integration/sandbox_test.go
dragosgheorghioiu: yeah they are but should probably merge to prod-stable the plugins (not sure when to do that though)

cmd/unikraft/integration/sandbox_test.go
dragosgheorghioiu: there should be only `exec/read/write/copy`
copy being just a wrapper over read and write using scp semantics

cmd/unikraft/testdata/TestHelp/instances
dragosgheorghioiu: yeah i can remove mkdir

internal/cmd/instances.go
dragosgheorghioiu: the idea to have them here is to hot load the plugin run the command and unload it on any instance so i think these can sit here

also the plugin name is optional because the default name is derived from the sdk

internal/cmd/sandbox.go
dragosgheorghioiu: this is how it works now i can update the help instruction but the flag is there only for the case where someone wants to give it a different name for some reason

jedevc: No, we should keep it at the top-level. If the sandbox plugin isn't installed, we should warn, and then dynamically auto-inject it (the platform still needs support for this, but that's the plan), then remove it when we're done.

This *is* a special case, and it's worth supporting as one.

internal/cmd/sandbox.go
dragosgheorghioiu: any non regular file is ignored (i think) will have to check

internal/cmd/sandbox.go
dragosgheorghioiu: removed mkdir completely

internal/resource/value/parse.go
dragosgheorghioiu: this is for the plugin flag to not split inside a json for example the config section

internal/integration/env.go
dragosgheorghioiu: it's used only for exec now i can split it for that one only but not really sure if it's worth it

internal/multimetro/client.go
dragosgheorghioiu: don't think it was really discussed 
i split this in a separate file for now

internal/cmd/sandbox.go
dragosgheorghioiu: inlinedd

internal/cmd/sandbox.go
craciunoiuc: well, softlinks at least we should support, they are not that complex. Only thing to decide is if we should move them as symlinks, or the path they point to. I can't really decide

hardlinks we can duplicate for now

internal/integration/env.go
craciunoiuc: I think it makes sense: first you add a commit to allow custom stdin -> then you use it

(you could put them together also, but that commit I think will be already quite big)

internal/cmd/sandbox.go
craciunoiuc: Hmm, if there's a plan for this already, then sure, we go with that

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	"unikraft.com/cloud/plugins/sandbox"
+
+	integ "unikraft.com/cli/internal/integration"
+)
+
+const (
+	sandboxPlugin = sandbox.PluginName
+
+	sandboxPluginRom = "plugins/sandbox:staging"
```
craciunoiuc: Will there be a staging/stable plugin for this? If yes, will there be a way to pick?

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	sandboxKraftfile = `
+spec: v0.7
+name: sandbox-e2e
+runtime: base-compat:latest
+rootfs:
+  format: erofs
+  source: ./Dockerfile
+type: dockerfile
```
craciunoiuc: I think the indent is not correct? `type` should be under `rootfs`, unless?

internal/multimetro/sandbox.go
```diff
@@ -0,0 +1,60 @@
...
+	return platform.Instance{Uuid: key.Ref().UUID}
+}
+
+// newSandboxClient builds a metro's plugin client from the same options its
+// platform client is built from.
+//
+// sandbox.NewClient ignores the HTTP client those options carry and installs
+// one of its own, so ours reaches the plugin per call in Opts instead. A call
+// made without Opts silently uses the plugin's own client, which knows
+// nothing of this metro's insecure setting.
```
craciunoiuc: Drop this part of the comment

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+	"unikraft.com/cli/internal/multimetro"
+	"unikraft.com/cli/internal/resource"
+	"unikraft.com/cli/internal/resource/value"
+	xio "unikraft.com/cli/internal/x/io"
+)
+
+type ExecOpts struct {
+	Cmd []string `arg:"" name:"command" help:"Command to pass to the instance." placeholder:"cmd"`
+
+	Plugin string   `name:"plugin" help:"Plugin name from the instance to run the command onto"`
```
craciunoiuc: Why the space?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+	Dir    string   `name:"dir" help:"Directory to execute the command from"`
```
craciunoiuc: Maybe `-w` as a short aka short for "workdir"?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+}
+
+type ExecSandboxInstanceCmd struct {
+	Target string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instances to run the command on."`
+
+	ExecOpts
+}
+
+func (cmd ExecSandboxInstanceCmd) Help() string {
+	return "The environment given with --env is the whole environment the command sees. Standard input is forwarded to the command only when it is redirected from a file or a pipe; run from a terminal, the command sees an empty standard input and cannot be typed at."
```
craciunoiuc: Does this mean it has no environment defaults?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+	target, err := resolveSandboxTarget(ctx, rsb, c.Target, c.Plugin)
+	if err != nil {
+		return err
+	}
+
+	in := stdio.Stdin
+	fd, isFile := in.(interface{ Fd() uintptr })
+	if in == nil || (isFile && term.IsTerminal(fd.Fd())) {
+		in = strings.NewReader("")
+	}
```
craciunoiuc, line 87: Is this a terminal check? I am pretty sure we have smth like this in `xio`, can you check?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+		return sandboxTarget{}, fmt.Errorf("ambiguous instance: %s (found %v)", target, keys)
+	}
+
+	instance, ok := resources[0].(Instance)
+	if !ok {
+		return sandboxTarget{}, fmt.Errorf("%q is not an instance", target)
+	}
+
+	if !instance.State.IsRunning() {
+		return sandboxTarget{}, fmt.Errorf("instance %q is not running (state: %s); start it with \"unikraft instance start %s\"", instance.Name, string(instance.State), instance.Name)
```
craciunoiuc: I think no need for `start it with \"unikraft instance start %s\"`

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+		}
+		if p.Name == plugin {
+			hasPlugin = true
+			break
+		}
+		loaded = append(loaded, p.Name)
+	}
+	if !hasPlugin {
+		if len(loaded) == 0 {
+			return sandboxTarget{}, fmt.Errorf("instance %q has no plugins loaded, so nothing answers to %q", instance.Name, plugin)
```
craciunoiuc: Weird error wording?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+}
+
+const (
+	sandboxLogPollInterval    = 100 * time.Millisecond
+	sandboxLogPollMaxInterval = 1 * time.Second
+)
+
+func shQuote(arg string) string {
+	return "'" + strings.ReplaceAll(arg, "'", `'\''`) + "'"
+}
```
craciunoiuc: Inline

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+	cmdUUID string
+	out     io.Writer
+
+	stdoutOffset, stderrOffset       uint64
+	stdoutAvailable, stderrAvailable uint64
+}
+
+func (l *commandLogs) pending() bool {
+	return l.stdoutAvailable > l.stdoutOffset || l.stderrAvailable > l.stderrOffset
+}
```
craciunoiuc: inline

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+		written += uint64(len(decoded))
+	}
+	return written, nil
+}
+
+// sandboxStdinChunkSize is the maximum number of bytes read from the local
+// standard input per iteration, and so the most that a single stdin request
+// carries to the command once base64-encoded.
+const sandboxStdinChunkSize = 32 * 1024 // 32 KiB
+
```
craciunoiuc: in another place u used the units package

maybe worth using here also aka `32 * units.kib`

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,882 @@
...
+			return "", 0, fmt.Errorf("creating local parent directories for %q: %w", local, err)
+		}
+	}
+
+	flags := os.O_WRONLY | os.O_CREATE | os.O_EXCL
+	if opts.force {
+		flags = os.O_WRONLY | os.O_CREATE | os.O_TRUNC | syscall.O_NOFOLLOW
+	}
+	f, err := os.OpenFile(local, flags, 0o644)
+	if err != nil {
```
craciunoiuc: Should we maybe keep permissions of the sandbox file? worth a discussion at least

At least users/groups are not worth because the sandbox will not have them

craciunoiuc (changes requested):
Some more comments

I saw you pushed meanwhile so maybe some stuff is now out of date

internal/cmd/sandbox.go
dragosgheorghioiu: yes

cmd/unikraft/integration/sandbox_test.go
dragosgheorghioiu: yeah there will you can pick by the tag latest will default to stable version

internal/cmd/sandbox.go
craciunoiuc: Do you think maybe it makes sense to keep the busybox defaults or others? like `USER`/`HOME`/etc.?

This in the actual sandbox, not the CLI stuff

dragosgheorghioiu: i can really see this working but at the same time maybe you want something that is as close to a "clean" environment as possible and give full control over the env and having some default would muddy things a bit

craciunoiuc: Yeah, this is more I guess for Andrei or someone else who worked on the original sandbox to decide

either way, it's a basic change

(philosophically: one could argue you should have a setup as common as possible so the agent doesn't realise it's in a sandbox)

internal/multimetro/sandbox.go
```diff
@@ -0,0 +1,62 @@
...
+// Sandbox returns this metro's plugin client, targeting the named plugin. An
+// empty name leaves the plugin the client was built with.
+func (c MetroClient) Sandbox(plugin string) SandboxClient {
+	if c.sandbox == nil {
+		return SandboxClient{}
+	}
+
+	sb := SandboxClient{
+		Client: c.sandbox.Client,
+		Opts:   make([]sandbox.Option, 0, len(c.sandbox.Opts)+1),
```
craciunoiuc: Wait what, why is there a +1 here?

internal/multimetro/sandbox.go
```diff
@@ -0,0 +1,62 @@
...
+	return platform.Instance{Uuid: key.Ref().UUID}
+}
+
+// newSandboxClient builds a metro's plugin client from the same options its
+// platform client is built from, plus that client's own *http.Client.
+//
+// The second argument is load-bearing: sandbox.NewClient never reads the HTTP
+// client those options carry, installing one of its own instead, so ours has to
+// reach the plugin per call through Opts. Drop it and every sandbox call
+// silently stops honouring this metro's Insecure setting.
```
craciunoiuc: Delete the bottom half of the comment

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,898 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
```
craciunoiuc: 2026 😩

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,898 @@
...
+	sandboxLogPollMaxInterval = 1 * time.Second
+
+	// sandboxLogChunkSize bounds one poll's response: without a limit the
+	// plugin returns each stream in its entirety, which a single 10s fetch
+	// cannot carry once a command has produced enough output.
+	sandboxLogChunkSize = 1 * units.MiB
+
+	// sandboxLogPollMaxFailures is how many consecutive failed polls are
+	// tolerated before exec gives up, so a transient error does not end the
+	// session of a command that is running fine.
```
craciunoiuc: Minify these comments

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,898 @@
...
+	}
+
+	contents, err := base64.StdEncoding.DecodeString(resp.Data.Contents)
+	if err != nil {
+		return nil, fmt.Errorf("decoding file contents: %w", err)
+	}
+	return contents, nil
+}
+
+const copyPathSeparator = ":"
```
craciunoiuc: I'm a bit conflicted about all these consts sitting above the first place they are used. Maybe worth putting them all up at the start of the file? @jedevc Opinions?

craciunoiuc (changes requested):
Final set of comments my side, if this works fine then the shape at least of the code is acceptable 

I will let @jedevc do a round now (at least) and then i do a final pass at the end

internal/cmd/sandbox.go
dragosgheorghioiu: moved them at the top of the file for now

internal/resource/sandbox.go
jedevc, line 1: Unrelated changes? Might be worth renaming this functionality though - just to avoid confusion.

Happy to brainstorm for other names.

internal/multimetro/sandbox.go
```diff
@@ -0,0 +1,54 @@
...
+		sb.Opts = append(sb.Opts, sandbox.WithPluginName(plugin))
+	}
+	return sb
+}
+
+// SandboxInstance addresses the plugin endpoint of the instance a key refers
+// to: a plugin client reaches its plugin through the instance's UUID alone.
+func SandboxInstance(key Key) platform.Instance {
+	return platform.Instance{Uuid: key.Ref().UUID}
+}
```
jedevc: This has a code smell.
1. It's only called in one place.
2. It doesn't use `Name` semantics if `UUID` isn't set.
3. It's not actually populated with everything that `platform.Instance` is expected to have when returned from the API.

I would drop this func.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,885 @@
...
+	sandboxLogChunkSize       = 1 * units.MiB
+	sandboxLogPollMaxFailures = 3
+	sandboxStdinChunkSize     = 32 * units.KiB
+	sandboxMaxFileSize        = 32 * units.MiB
+	copyPathSeparator         = ":"
+)
+
+type ExecOpts struct {
+	Cmd    []string `arg:"" name:"command" help:"Command to pass to the instance." placeholder:"cmd"`
+	Plugin string   `name:"plugin" help:"Plugin name from the instance to run the command onto"`
```
jedevc: Wonder if we should have a default? e.g. `sandbox`?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,885 @@
...
+	Dir    string   `name:"dir" short:"w" help:"Directory to execute the command from"`
+	Env    []string `name:"env" short:"e" sep:"none" help:"Environment variable." placeholder:"<key>=<value>" example:"DEBUG=true"`
+}
```
jedevc: Why a separate struct?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,885 @@
...
+		{
+			Description: "Run a command with environment variables set",
+			Commands: []string{
+				"unikraft instance exec my-instance --plugin sandbox -e DEBUG=true -e PORT=8080 -- ./start.sh",
+			},
+		},
+	}
+}
+
+func (c *ExecSandboxInstanceCmd) Run(ctx context.Context, stdio config.Stdio, rsb *resource.Sandbox) error {
```
jedevc: Yeah, the `resource.Sandbox` makes things suuuuper confusing. We should definitely rename it.

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,885 @@
...
+	log.G(ctx).Trace().Msg("executing command")
+
+	cmdline, err := buildExecCommand(opts.Cmd)
+	if err != nil {
+		return err
+	}
+
+	req := sandbox.RunCommandRequest{
+		Cmd: cmdline,
+	}
```
jedevc: Ah. That's annoying. I would have expected this to just directly pass all the args, and not mess around.

Like so: https://github.com/unikraft-cloud/plugins/pull/23

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,885 @@
...
+	for _, arg := range cmdArgs {
+		if strings.ContainsRune(arg, 0) {
+			return "", fmt.Errorf("cannot quote command argument %q: it contains a NUL byte", arg)
+		}
+		quoted = append(quoted, "'"+strings.ReplaceAll(arg, "'", `'\''`)+"'")
+	}
+	return strings.Join(quoted, " "), nil
+}
+
+func (t sandboxTarget) exec(ctx context.Context, out io.Writer, in io.Reader, opts ExecOpts) error {
```
jedevc: While this is kinda fine for now... 

I wonder if a more "go-primitive" way of doing this would be to provide a `Cmd`. (like in `os/exec.Cmd`).

Because the problem is, this calling code is all littered with all the details of the API. We have to handle stdin/stdout/signal plumbing all here. e.g. when we try and add all this for the shell, we'll end up having to reproduce all this right? https://github.com/unikraft-cloud/cli/pull/463

I wonder if we should create a little wrapper - the caller here would just create a `Cmd`, and then call `Start`. And have that handle signals + stdin + stdout + polling. Then all the calling code for the plugins becomes a lot simpler.

internal/resource/sandbox.go
dragosgheorghioiu: hmm maybe scope or lease can work?

internal/cmd/sandbox.go
dragosgheorghioiu: i was setting them in the function proper to the name exported by the sdk
i find the kong var alternative a bit weird but maybe it's better

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,437 @@
...
+
+const copyPathSeparator = ":"
+
+var sandboxKongVars = kong.Vars{"sandbox_plugin": sandbox.PluginName}
+
+type ExecSandboxInstanceCmd struct {
+	Target string   `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instance to run the command on."`
+	Cmd    []string `arg:"" name:"command" help:"Command to pass to the instance." placeholder:"cmd"`
+
+	Plugin string   `name:"plugin" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
```
craciunoiuc, line 37: I hope that default value is correct 👀

internal/cmd/sandbox.go
jedevc: Mm, let's put it in kong, because then they appear as a default flag with `--help`.

internal/resource/sandbox.go
jedevc: What about `Partition`?

internal/cmd/sandbox.go
```diff
@@ -0,0 +1,437 @@
...
+	in := stdio.Stdin
+	if in == nil || xio.IsTTYReader(in) {
+		in = strings.NewReader("")
+	}
+
+	cmd := target.CommandLine(ctx, c.Cmd)
+	cmd.Dir = c.Dir
+	cmd.Env = env
+	cmd.Stdin = in
+	cmd.Stdout = xio.Unwrap(stdio.Stdout)
```
jedevc, line 93: Beautiful :chefskiss:

craciunoiuc (commented):
I actually did not find anything else significant

internal/cmd/sandbox.go
dragosgheorghioiu: it's set from the sdk https://github.com/unikraft-cloud/cli/pull/458/changes#diff-4058e480aa874118f6965bc58392b3a953c7cbeab37d8982b33cf8f4f0579640R30
it defaults to sandbox for now

jedevc (approved):
```
```

craciunoiuc (approved):
All good here. Thanks!

```
```

PR #460, opened 2026-08-20
fix(volume-templates): Create a template per volume

jedevc:
Fallout from https://github.com/unikraft-cloud/cli/pull/395 — `unikraft volume template create a b` comma-joined its positional volumes into one `--set volumes` value and relied on the parser splitting it back apart, so it now goes looking for a volume named `a,b`.

PR #403, opened 2026-07-07
feat!: Add instance branching + checkpointing

jedevc:
TODO:
- [x] Quick design once-over

PR #461, opened 2026-08-20
fix(volume-templates)!: Restrict create to a single volume

jedevc:
Volume templates got missed when we made this call for instance templates and checkpoints in https://github.com/unikraft-cloud/cli/pull/403#discussion_r3719782435 — same reasoning applies, the platform API only creates one at a time and returning multiple resources from create was only ever meant for `--replicas`.

This also drops the multi-volume path that https://github.com/unikraft-cloud/cli/pull/460 repaired, since there is nothing left to repair once the command takes a single volume.

Breaking: `unikraft volume template create a b` now takes a single volume, and the create field is `volume` rather than `volumes`.

### DIFF
--- a/cmd/unikraft/integration/volume_template_test.go
+++ b/cmd/unikraft/integration/volume_template_test.go
@@ -10,7 +10,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


craciunoiuc (approved):
All good here. Thanks!

```
```

PR #462, opened 2026-08-21
fix(gomod): Pin 'ultraviolet' to older version

craciunoiuc:
Turns out there is a regression causing our header to duplicate every time a row is removed or added. Pinning reverts it.

Closes: TOOL-1360

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -196,3 +196,8 @@ require (
 	google.golang.org/protobuf v1.36.12-0.20260120151049-f2248ac996af // indirect
 	tailscale.com v1.94.1 // indirect
 )
+
+// Pin ultraviolet to the last revision before it started misbehaving.
+// Concretely every screen resize currently duplicates our table header.
+// See: https://github.com/charmbracelet/ultraviolet/issues/155
+replace github.com/charmbracelet/ultraviolet => github.com/charmbracelet/ultraviolet v0.0.0-20260703014108-f5a850f9c2b7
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -196,3 +196,7 @@ require (
 	google.golang.org/protobuf v1.36.12-0.20260120151049-f2248ac996af // indirect
 	tailscale.com v1.94.1 // indirect
 )
+
+// Pin ultraviolet to the last revision before it started resizing its screen.
+// Concretely every screen resize currently duplicates our table header.
+replace github.com/charmbracelet/ultraviolet => github.com/charmbracelet/ultraviolet v0.0.0-20260703014108-f5a850f9c2b7
```
jedevc: This needs a link to an issue in this repo, it looks like maybe there are some that fit the bill? Like, https://github.com/charmbracelet/ultraviolet/issues/155, this sounds very similar.

If we don't have that, then there's no way of knowing when we can safely remove this and can upgrade.

jedevc: Also, is resizing the bug? That doesn't seem the same as what's described in the issue/discord.

craciunoiuc: hmm looks like it's something with resize, just not exactly what I wrote there

But yeah, that issue looks like the correct one, will edit now

craciunoiuc: Pushed :saluting_face:

jedevc (approved):
```
```

PR #463, opened 2026-08-21
feat: Sandbox shell command

dragosgheorghioiu:
Closes: TOOL-914


internal/cmd/shell.go
```diff
@@ -0,0 +1,328 @@
...
+	}, shell.Streams{In: stdio.Stdin, Out: stdio.Stdout, Err: stdio.Stderr})
+}
+
+// shellBuiltins answers the lines opening with ":".
+type shellBuiltins struct {
+	key       string
+	partition *resource.Partition
+}
+
+type shellBuiltinCmds struct {
```
dragosgheorghioiu, line 129: maybe should also move the builtins out of cmd

jedevc:
Question - how much of this is now replaced by the `x` stack of PRs?

dragosgheorghioiu:
excluding the base sandbox integration it's around 700 lines (from which 250 are tests) now

internal/cmd/shell_sandbox.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+const (
+	// sandboxReapTimeout is how long a probe nobody waits for is given to die.
+	sandboxReapTimeout = 30 * time.Second
+
+	sandboxSignalTimeout = 5 * time.Second
+
+	// sandboxInterruptGrace is how long an interrupted command is given to report what it died of.
+	sandboxInterruptGrace = 10 * time.Second
```
nderjung, line 30: can we make these flags?

internal/cmd/shell.go
```diff
@@ -0,0 +1,328 @@
...
+	return nodes
+})
+
+type ShellSandboxInstanceCmd struct {
+	Target string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instance to open a shell on."`
+
+	Plugin  string            `name:"plugin" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
+	Dir     string            `name:"dir" help:"Directory to start the shell in." default:"/"`
+	Env     map[string]string `name:"env" short:"e" help:"Environment variables." placeholder:"<key>=<value>" example:"DEBUG=true,PORT=8080" mapsep:","`
+	Command string            `name:"command" short:"c" help:"Run a single command line and exit."`
```
nderjung, line 54: Can we have a `--no-autoload` flag and have the default functionality to attempt to autoload/inject the sandbox plugin if it is not present on the instance? This would make using a sandbox more seamless and require less setup by the user. We can use `defer` to automatically remove it after the program exits.

```go
import "unikraft.com/cloud/plugins/sandbox"

if !noAutoload && !alreadyAttached() {
	defer detachPlugin()
	attachPlugin()
}

// ...

// attachPlugin mounts the plugin (backed by the given ROM image) on
// the instance so its authenticated plugin endpoint becomes reachable.
func attachPlugin(ctx context.Context, c multimetro.MetroClient, instanceUuid, pluginName, pluginRom string) error {
	log.G(ctx).
		Trace().
		Str("uuid", instanceUuid).
		Str("rom", pluginRom).
		Str("name", pluginName).
		Msg("attaching plugin")

	value := any([]platform.InstancePlugin{{
		Name: pluginName,
		Rom:  pluginRom,
	}})
	_, err := c.UpdateInstanceByUUID(ctx, instanceUuid, platform.UpdateInstanceByUUIDRequestBody{
		Prop:  platform.MutableInstancePropertyPlugins,
		Op:    platform.MutableInstanceOperationAdd,
		Value: &value,
	})
	return err
}
```

internal/cmd/shell.go
```diff
@@ -0,0 +1,328 @@
...
+
+	nodes := slices.Clone(parser.Model.Children)
+	slices.SortFunc(nodes, func(a, b *kong.Node) int { return strings.Compare(a.Name, b.Name) })
+	return nodes
+})
+
+type ShellSandboxInstanceCmd struct {
+	Target string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instance to open a shell on."`
+
+	Plugin  string            `name:"plugin" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
```
nderjung, line 51: Would be useful to be able to also use a different plugin ROM (e.g. during testing/dev). Suggesting two separate fags instead: `--plugin-name` (to rename the plugin on the instance, replacing `--plugin` now) and `--plugin-rom` to set an alternative ROM image.

craciunoiuc:
@nderjung too late old man, this PR needs rebasing to use the work from X

@dragosgheorghioiu could you do a preemptive update to the ~700 line one you were mentioning? Also see if the comments from Alex still apply to the X version

dragosgheorghioiu:
about 50% are tests and yeah all the comments still apply

internal/cmd/shell.go
dragosgheorghioiu: i would do this in a separate pr because i would need to update all the commands and this pr is only related to the shell

internal/cmd/shell.go
dragosgheorghioiu: this would work once the platform supports full hotplug of plugins would make sense to add this then

internal/cmd/shell.go
jedevc: `--plugin-rom` would only apply for "autoload" right?

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -70,6 +77,12 @@ func TestSandbox(t *testing.T) {
 		assert.Contains(t, out, "to-stdout")
 		assert.Contains(t, out, "to-stderr")
 
+		// A command that fails on the instance is not this CLI failing: the
+		// status it exited with is reported as our own, with nothing said over
+		// whatever the command printed.
+		r.Run(t, []string{"unikraft", "instance", "exec", instName, "--", "sh", "-c", "exit 3"},
+			integ.ExpectFail())
```
jedevc: Might be good to actually check the exact exit code?

internal/cmd/shell.go
dragosgheorghioiu: i think so, but i am not sure what this can help with outside of picking different versions of the sandbox plugin

on that same idea maybe also a `--plugin-config` flag would work

cmd/unikraft/testdata/TestHelp/instances
```diff
@@ -2510,6 +2512,83 @@ Global flags:
...
+`Do you want to continue? [Y/n]` can be answered — and anything typed ahead
+goes to that command rather than to the next prompt.
+
+Ctrl-C interrupts the command on the instance, and the shell waits for it
+to stop before reporting what it stopped of in `$?`; the rest of the line
+is abandoned.
+
+The instance offers no terminal, so programs that need one — `vim`, `top`,
+`less` — will not work there yet.
+A command that fails on the instance sets `$?` but does not fail the shell.
```
jedevc: I think a lot of these are leaked implementation details - this is just how users expect this to work right?

internal/cmd/sandbox.go
```diff
@@ -95,7 +101,13 @@ func (c *ExecSandboxInstanceCmd) Run(ctx context.Context, stdio config.Stdio, pa
...
 	cmd.Stderr = out
 
-	return cmd.Run()
+	var exit *sandbox.ExitError
+	if err := cmd.Run(); errors.As(err, &exit) {
+		return ExitStatus(exit.Code)
+	} else if err != nil {
+		return err
+	}
+	return nil
```
jedevc: Could use `shell.ExitStatus` here?

internal/cmd/shell.go
```diff
@@ -0,0 +1,122 @@
...
+	"unikraft.com/x/shell"
+	xsignal "unikraft.com/x/signal"
+	xstdio "unikraft.com/x/stdio"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/resource"
+)
+
+var shellBanner = strings.Join([]string{
+	"this shell is experimental",
```
jedevc: maybe a little warning icon `⚠︎`?

internal/cmd/shell.go
```diff
@@ -0,0 +1,122 @@
...
+	"no terminal, so vim, top and less will not run",
+	"no job control, so no ctrl-z, bg or fg",
```
jedevc: Do we need this here? I think it's okay if it's just listed in the `--help` and `docs`.

internal/cmd/shell.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+
+	code, err := shell.Run(ctx, shell.Config{
+		Instance:       c.Target,
+		Dir:            c.Dir,
+		Env:            c.Env,
+		Command:        c.Command,
+		Transport:      newSandboxTransport(target, sandboxTimeouts{InterruptGrace: c.InterruptGrace, Reap: c.ReapTimeout}),
```
jedevc, line 90: Hehe, glorious :tada:

internal/cmd/shell.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil || code == 0 {
+		return err
+	}
+	return ExitStatus(code)
+}
+
+// shellStreams is the CLI's stdio as the shell takes it.
+func shellStreams(stdio config.Stdio) xstdio.Stdio {
+	return xstdio.Stdio{Stdin: stdio.Stdin, Stdout: stdio.Stdout, Stderr: stdio.Stderr}
+}
```
jedevc: Inline?

internal/cmd/shell_sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+	if len(env) == 0 {
+		return nil
+	}
+	m := make(map[string]string, len(env))
+	for _, record := range env {
+		name, value, _ := strings.Cut(record, "=")
+		m[name] = value
+	}
+	return m
+}
```
jedevc: I would make `Env` on `Cmd` a `[]string` (to really mirror `os/exec.Cmd`). Then we do this transformation in the `internal/sandbox` package.

internal/sandbox/cmd.go
```diff
@@ -306,6 +313,7 @@ func (c *Cmd) Wait() error {
 				}
 				return err
 			}
+			c.Forget(c.ctx)
```
jedevc: bug: Only on the success path, so every command that exits non-zero leaves its record on the instance — and in a shell session that's every `false`, every typo, every `grep` that matched nothing. Move it to a `defer` after the `c.done` receive?

```
$ # same as TestCmdForget, but fake.exit(1)
=== RUN   TestForgetLeakOnNonZero
    Error:      Should be true
    Messages:   a command that failed should still be dropped
--- FAIL: TestForgetLeakOnNonZero (0.00s)
```

internal/cmd/shell.go
```diff
@@ -0,0 +1,122 @@
...
+	"no terminal, so vim, top and less will not run",
+	"no job control, so no ctrl-z, bg or fg",
+}, "\n")
+
+type ShellSandboxInstanceCmd struct {
+	Target string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instance to open a shell on."`
+
+	Plugin  string            `name:"plugin" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
+	Dir     string            `name:"dir" help:"Directory to start the shell in." default:"/"`
+	Env     map[string]string `name:"env" short:"e" help:"Environment variables." placeholder:"<key>=<value>" example:"DEBUG=true,PORT=8080" mapsep:","`
```
jedevc: bug: `mapsep:","` means we reject any value with a comma in it, and `exec` right next door doesn't — same flag, same short, different rules. Can we use `[]string` with `sep:"none"` to match?

https://github.com/unikraft-cloud/cli/blob/88ea3587b05befbab422b1e6cf8c6b23cecf38a0/internal/cmd/sandbox.go#L40

```
$ unikraft instance shell my-inst -e 'NO_PROXY=localhost,127.0.0.1' -c true
│ error:
│   parsing arguments: --env: expected "<key>=<value>" but got "127.0.0.1"

$ unikraft instance exec my-inst -e 'NO_PROXY=localhost,127.0.0.1' -- true
│ error:
│   references not found: [my-inst]
```

internal/sandbox/pipe.go
```diff
@@ -0,0 +1,56 @@
...
+	closed func()
+	dead   bool
+}
+
+func (w *deadPipe) Write(p []byte) (int, error) {
+	if w.dead {
+		return len(p), nil
+	}
+	if _, err := w.out.Write(p); err != nil {
+		w.dead = true
```
jedevc: suggestion: Any write error counts as a dead reader here, so a full disk on `> file` — or a genuinely broken terminal — gets swallowed and we report the command as having succeeded. `errors.Is(err, syscall.EPIPE) || errors.Is(err, io.ErrClosedPipe)`?

`deafWriter` returns a plain `errors.New("broken pipe")`, so the tests would need a real `syscall.EPIPE` — which is rather the point.

cmd/unikraft/integration/sandbox_test.go
```diff
@@ -313,4 +326,64 @@ func TestSandbox(t *testing.T) {
...
+			require.Eventually(t, func() bool {
+				_, err := r.RunRaw(t, []string{
+					"unikraft", "instance", "read", instName, started, "./probe",
+				}, integ.WithWorkDir(t.TempDir()))
+				return err == nil
+			}, 90*time.Second, 2*time.Second, "the command never reached the instance")
+
+			proc.Interrupt()
+			out, _ := proc.Wait()
+			assert.NotContains(t, out, marker, "an interrupt abandons the rest of the line")
```
jedevc, line 385: nit: This passes just as well if the shell died instantly and printed nothing at all. We also drop the status from `proc.Wait()`, which is the bit that says it was interrupted rather than crashed — assert on that too?

internal/cmd/shell.go
```diff
@@ -0,0 +1,122 @@
...
+	"this shell is experimental",
+	"no terminal, so vim, top and less will not run",
+	"no job control, so no ctrl-z, bg or fg",
+}, "\n")
+
+type ShellSandboxInstanceCmd struct {
+	Target string `arg:"" name:"target" completion-predictor:"resource-key-instance" help:"Target instance to open a shell on."`
+
+	Plugin  string            `name:"plugin" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
+	Dir     string            `name:"dir" help:"Directory to start the shell in." default:"/"`
```
jedevc: nit: No placeholders on `--dir` or `--command`, so help renders them as a bare `--dir` with no hint of what goes there. `exec` also gives `--dir` a `-w`. Worth matching both?

internal/flush/flush.go
```diff
@@ -3,7 +3,7 @@
 // Licensed under the BSD-3-Clause License (the "License").
```
jedevc: nit: :eyes: A whole package for two interface declarations. Everything else in `internal/x/io` went upstream — any reason these two couldn't go with them?

internal/cmd/shell_sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+	timeouts sandboxTimeouts
+}
+
+// newSandboxTransport reaches target; a zero timeout is the default one.
+func newSandboxTransport(target sandbox.Target, timeouts sandboxTimeouts) sandboxTransport {
+	t := sandboxTransport{target: target, timeouts: sandboxTimeouts{
+		InterruptGrace: cmp.Or(timeouts.InterruptGrace, defaultSandboxTimeouts.InterruptGrace),
+		Reap:           cmp.Or(timeouts.Reap, defaultSandboxTimeouts.Reap),
+	}}
+	t.ExecTransport = t.exec
```
jedevc, line 42: nit: `t.exec` binds a *copy* of `t` taken here, so any field we add later and set after this line will silently be invisible to `exec`. Fine today, but it'll bite someone.

jedevc (approved):
```
```

PR #465, opened 2026-08-21
gomod(deps): Update to sdk@487e048643d5

jedevc:
Picks up https://github.com/unikraft-cloud/go-sdk/pull/256, so `NewHTTPClient` clones `DefaultTransport` rather than building a bare one — clients get the 90s idle connection timeout and HTTP/2 back instead of holding their idle sockets open for the lifetime of the process.

Same bump as https://github.com/unikraft-cloud/console/pull/1109, which is where the fd exhaustion this fixes actually showed up.

The integration suite is red against staging, but `prod-staging` fails identically without this patch — same seven subtests, same causes: `sched_priority` is rejected by the backend there, and the five `unikraft build` tests 403 because the registry namespace is at 963 MiB of its 1.0 GiB limit. `./internal/builder` is green either way.


khatibomar (approved):
```
```

PR #477, opened 2026-08-25
feat(instances): Add annotations

jedevc:
Fixes https://linear.app/unikraft/issue/TOOL-851/add-instance-annotations-to-cli.

Annotations are like tags, but key-value and visible to the guest — they go into the instance's startdata, so a workload can read what it was deployed with. Settable on create and edit, read-only on templates and checkpoints.

Spec landed in https://github.com/unikraft-cloud/proto/pull/344, so this also bumps the SDK to pick up the typed field.


khatibomar (approved):
```
```

PR #478, opened 2026-08-25
fix(instance-templates): Link volumes to volume templates

jedevc:
Templates and checkpoints were linking their attached volumes at the `volume` resource type, but converting an instance turns its volumes into volume templates — and the two platform lookups are mutually exclusive, so that link could only ever miss. Broke TUI drill-down and left the volumes untracked by the sandbox at teardown.

While auditing the two resources against the platform status handlers, `type` and `snapshot` turned out to be populated for both and `boot_time_us`/`net_time_us` for templates, so those are surfaced too. Autokill was in scope as well but is left out — #355 already covers it.

Schedules are also emitted and still missing, but instances have no schedule support to be consistent with yet.


khatibomar (approved):
```
```

PR #480, opened 2026-08-31
feat: Add --sandbox flag

dragosgheorghioiu:
added to the create command
```
--sandbox                        # persistent, all defaults
--sandbox=path=/data,size=2GiB   # comma-separated key=value, like --scale-to-zero
--sandbox=on / =off              # bare-word shorthand for persist=true/false
```


dragosgheorghioiu:
marking as draft until https://github.com/unikraft-cloud/plugins/pull/22 is merged

PR #481, opened 2026-08-31
feat(instances): Add network interfaces and relays

jedevc:
Instances now report their network interfaces properly - name, tap device, autoconfig - and can configure them on create, including a relay that routes all of an instance's traffic through another instance's interface. The platform has had relays since https://github.com/unikraft-cloud/platform/pull/846 but there was no CLI surface for them at all.

Reaching the relay from the flat `key=value` form needed dotted keys in `internal/resource/value`, so `--network relay.name=my-router-eth0,relay.dns=false` addresses the nested field directly instead of every nested type having to invent a text form that packs its subfields into one value.

One caveat worth knowing: `network_interfaces` is gated behind the net-manager permission, and the API rejects the whole field as an *unknown member* rather than a permission error, so `--network` reads as broken rather than forbidden on an account without it.

Fixes https://linear.app/unikraft/issue/TOOL-1409/relays-in-the-cli.

### DIFF
--- a/cmd/unikraft/integration/instance_test.go
+++ b/cmd/unikraft/integration/instance_test.go
@@ -234,6 +234,62 @@ cmd: ["cat", "/sys/class/uio/uio0/device/startdata"]
 		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + instName})
 	})
 
+	t.Run("create-relay", func(t *testing.T) {
+		r := runner(t, true, []string{staging, stable})
+		routerName, clientName, optOutName := uniq(), uniq(), uniq()
+
+		// A relay points at an interface, and the generated interface name is
+		// not predictable - it falls back to eth-<suffix> whenever
+		// network_interfaces is given explicitly - so the router needs one
+		// named up front to aim at.
+		iface := "test-" + routerName + "-eth0"
+		create := func(name string, opts ...string) []string {
+			return append([]string{
+				"unikraft", "instance", "create",
+				"--name", "test-" + name,
+				"--metro", r.Config.MetroName,
+				"--image", "nginx:latest",
+				"--memory", "128",
+				"--vcpus", "1",
+				"--set", "autostart=false",
+			}, opts...)
+		}
+
+		r.Run(t, create(routerName, "--network", "name="+iface, "--output", "quiet"))
+
+		out := r.Run(t, create(clientName, "--network", "relay.name="+iface))
+		assert.Regexp(t, `relay:`, out)
+		assert.Regexp(t, `name:\s+`+regexp.QuoteMeta(iface), out)
+		assert.Regexp(t, `dns:\s+true`, out)
+
+		// relay.dns is a dotted key rather than a nested value because relay=
+		// would take the whole value as the interface name.
+		out = r.Run(t, create(optOutName, "--network", "relay.name="+iface+",relay.dns=false"))
+		assert.Regexp(t, `dns:\s+false`, out)
+
+		out = r.Run(t, create(uniq(), "--network", "relay.name=test-"+routerName+"-nonexistent"), integ.ExpectFail())
+		assert.Regexp(t, `Invalid relay`, out)
+
+		// Relay chains are rejected, so the client's own interface cannot
+		// itself be relayed through.
+		clientIface := strings.TrimSpace(r.Run(t, []string{
+			"unikraft", "instance", "get", "test-" + clientName,
+			"--output", "template={{ (index .networks 0).name }}",
+		}))
+		require.NotEmpty(t, clientIface)
+		out = r.Run(t, create(uniq(), "--network", "relay.name="+clientIface), integ.ExpectFail())
+		assert.Regexp(t, `Invalid relay`, out)
+
+		r.Run(t, []string{"unikraft", "instance", "delete", "test-" + clientName, "test-" + optOutName})
+
+		// The relay's datapath is torn down asynchronously after its last
+		// client goes, and until it is the target instance deletes as -EBUSY.
+		require.Eventually(t, func() bool {
+			_, err := r.RunRaw(t, []string{"unikraft", "instance", "delete", "test-" + routerName}, integ.WithoutCancel())
+			return err == nil
+		}, 2*time.Minute, 5*time.Second, "relay target never became deletable")
+	})
+
 	t.Run("create-oom", func(t *testing.T) {
 		r := runner(t, true, []string{staging, stable})
 		instName := uniq()
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -93,6 +93,8 @@ type InstanceCreateCmd struct {
 	Volume []InstanceVolume `group:"flag-create" shortcut:"volumes" short:"v" sep:"none" help:"Attach volume." placeholder:"<name>:<path>[:<options>]" example:"my-vol:/data,cache:/tmp:ro,data:/mnt:size=10GiB"`
 	Rom    []InstanceRom    `group:"flag-create" shortcut:"roms" sep:"none" help:"Attach ROM." placeholder:"image=<ref>,at=<path>" example:"image=myuser/my-rom:latest\\,at=/rom0\\,name=my-rom,dir=./mydata\\,at=/rom"`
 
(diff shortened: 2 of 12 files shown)


internal/cmd/instances.go
```diff
@@ -244,9 +244,22 @@ type Instance struct {
...
+	TapName    string                `mirror:"tap_name" field:"tap-name,long"`
+	Relay      *InstanceNetworkRelay `mirror:"relay" field:",embed"`
+	Autoconfig *bool                 `mirror:"autoconfig" field:",long"`
+}
+
+// InstanceNetworkRelay is the interface all of this interface's traffic is
+// routed through. The target is another instance's interface, which has no
+// API of its own, so this holds plain identifiers rather than a Link.
+type InstanceNetworkRelay struct {
+	Name string `mirror:"name" field:",long"`
```
craciunoiuc, line 260: This PR will conflict with the other refactor, so I'll let you decide the order in which you want to merge

internal/cmd/instances.go
```diff
@@ -244,22 +246,50 @@ type Instance struct {
...
 // InstanceNetworkRelay is the interface all of this interface's traffic is
 // routed through. The target is another instance's interface, which has no
 // API of its own, so this holds plain identifiers rather than a Link.
 type InstanceNetworkRelay struct {
-	Name string `mirror:"name" field:",long"`
-	UUID string `mirror:"uuid" field:",long"`
-	DNS  *bool  `mirror:"relay_dns" field:",long"`
+	Name string `name:"name" mirror:"name" json:"name,omitempty" field:",long"`
+	UUID string `name:"uuid" mirror:"uuid" json:"uuid,omitempty" field:",long"`
+	DNS  *bool  `name:"dns" mirror:"relay_dns" json:"dns,omitempty" field:",long"`
```
craciunoiuc, line 267: I assume the mirror here is correct? 🤔

cmd/unikraft/integration/instance_test.go
```diff
@@ -234,6 +234,62 @@ cmd: ["cat", "/sys/class/uio/uio0/device/startdata"]
...
 	})
 
+	t.Run("create-relay", func(t *testing.T) {
+		r := runner(t, true, []string{staging, stable})
+		routerName, clientName, optOutName := uniq(), uniq(), uniq()
+
+		// A relay points at an interface, and the generated interface name is
+		// not predictable - it falls back to eth-<suffix> whenever
+		// network_interfaces is given explicitly - so the router needs one
+		// named up front to aim at.
```
craciunoiuc, line 244: No need to explain this I think?

craciunoiuc (commented):
Some nits here also, code looks fine

internal/cmd/instances.go
```diff
@@ -1162,6 +1205,35 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
 				}
 				req.Roms = append(req.Roms, reqRom)
 			}
+		case "networks":
+			for _, net := range field.Create.Set.([]*InstanceNetwork) {
```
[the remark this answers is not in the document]
jedevc: Fixed - `uuid` and `private-ip` are now `json:"-"` as well as `name:"-"`, so neither spelling advertises them. The compact form still errors on `uuid=`; the JSON form now drops it as an unknown member, which is what it does for every other unknown member on every other type here. Covered by a new `TestCreatePatches` case.

internal/cmd/instances.go
jedevc: Rebased onto the refactor - the `--network` flag is now declared on `Instance.Networks` with the `flag:`/`help:`/`example:` tags rather than as a shortcut on `InstanceCreateCmd`.

internal/cmd/instances.go
jedevc: It is - the API really does name that member `relay_dns` inside the `relay` object itself (`platform.NetworkInterfaceRelay.RelayDns`, `platform.InstanceNetworkInterfaceRelay.RelayDns`), so the path reads `relay.relay_dns`. Added a one-liner saying so, since it reads like a typo otherwise.

cmd/unikraft/integration/instance_test.go
jedevc: Dropped.

internal/cmd/instances.go
```diff
@@ -93,6 +93,8 @@ type InstanceCreateCmd struct {
 	Volume []InstanceVolume `group:"flag-create" shortcut:"volumes" short:"v" sep:"none" help:"Attach volume." placeholder:"<name>:<path>[:<options>]" example:"my-vol:/data,cache:/tmp:ro,data:/mnt:size=10GiB"`
 	Rom    []InstanceRom    `group:"flag-create" shortcut:"roms" sep:"none" help:"Attach ROM." placeholder:"image=<ref>,at=<path>" example:"image=myuser/my-rom:latest\\,at=/rom0\\,name=my-rom,dir=./mydata\\,at=/rom"`
 
+	Network []InstanceNetwork `group:"flag-create" shortcut:"networks" sep:"none" help:"Attach network interface.\n  name: interface name\n  relay.name: interface to route all traffic through\n  relay.uuid: same, by uuid\n  relay.dns: whether the relay forwards DNS (default true)\n  ip: address in CIDR notation, requires tap-name\n  mac: address, requires tap-name\n  tap-name: TAP device to bring your own interface\n  autoconfig: whether the guest configures the interface itself" placeholder:"<key>=<value>" example:"relay.name=my-router-eth0,relay.name=my-router-eth0\\,relay.dns=false,name=eth1\\,tap-name=tap0\\,ip=10.0.0.5/24"`
```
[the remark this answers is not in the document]
jedevc: Partly. Added a `relay.uuid` case (3) - it reads the router interface uuid back out and creates against it. Cleanup (4) was already there: every instance that creates successfully is deleted, and the cases that assert failure never create one.

Not adding (1) and (2). `tap-name`/`ip`/`mac` bring your own interface, which needs a TAP device on the host and the `net-manager` permission the integration account does not have, so those cases would fail for reasons unrelated to the CLI. The parsing and request-building for all four keys is covered in `TestCreatePatches`.

internal/resource/value/fields.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+	}
+}
+
+// structValue returns f as a struct for a dotted path to descend into. An
+// embedded field is skipped, having no key of its own to sit under.
+//
+// alloc fills in nil pointers on the way, for parsing; rendering leaves them
+// alone and so finds nothing to descend into.
```
craciunoiuc, line 59: Remove?

internal/cmd/instances.go
```diff
@@ -128,8 +128,8 @@ type Instance struct {
...
 	Plugins []*InstancePlugin `mirror:"instance.plugins" field:",embed" create:"set" edit:"set,add,del=strings" flag:"plugin" sep:"none" help:"Load plugin into the instance." placeholder:"name=<name>,image=<ref>[,config=<json>]" example:"name=sandbox\\,image=plugins/sandbox:latest,name=sandbox\\,image=plugins/sandbox:latest\\,config={\"persist_path\":\"/data\"}"`
 
-	Networks []InstanceNetwork `mirror:"instance.network_interfaces" field:",embed"`
-	Gpus     []InstanceGpu     `mirror:"instance.gpus" field:"gpus,embed"`
+	Networks []*InstanceNetwork `mirror:"instance.network_interfaces" field:",embed" create:"set" flag:"network" sep:"none" help:"Attach network interface.\n  name: interface name\n  relay.name: interface to route all traffic through\n  relay.uuid: same, by uuid\n  relay.dns: whether the relay forwards DNS (default true)\n  ip: address in CIDR notation, requires tap-name\n  mac: address, requires tap-name\n  tap-name: TAP device to bring your own interface\n  autoconfig: whether the guest configures the interface itself" placeholder:"<key>=<value>" example:"relay.name=my-router-eth0,relay.name=my-router-eth0\\,relay.dns=false,name=eth1\\,tap-name=tap0\\,ip=10.0.0.5/24"`
```
craciunoiuc, line 131: Is there no better way to represent sub-structs? this is ugly 😔

internal/cmd/instances.go
```diff
@@ -1363,6 +1391,35 @@ func (Instance) Create(ctx context.Context, fields []resource.Field) ([]resource
...
+						Name:     ptr.NilIfZero(net.Relay.Name),
+						Uuid:     ptr.NilIfZero(net.Relay.UUID),
+						RelayDns: net.Relay.DNS,
+					}
+				}
+				if net.MAC != "" {
+					// HACK: the spec's network interface omits mac, though
+					// /v1/instances has accepted it since MAC-only custom
+					// interfaces landed.
+					macJSON, _ := json.Marshal(net.MAC)
```
craciunoiuc, line 1416: Wait, why is the mac field not in the spec? 🤔

craciunoiuc (commented):
Last comments.

Can you also reword the commits with the new rules cause this is older and has very long comments.

internal/cmd/instances.go
jedevc: See https://discord.com/channels/879723098881028167/1448722316782735463/1549777441785122928. Please do weight in :grin:

craciunoiuc: My weigh is in

jedevc: So, is this okay then? Or do you have better suggestions :sweat:

craciunoiuc: Ok for now, as I have no better suggestion :pensive:

PR #482, opened 2026-08-31
feat(instances): Add plugins flag

dragosgheorghioiu:
Add --plugin to instance create, mirroring the API's instance.plugins: a name, the ROM to load it from, and an optional JSON config blob.

value.Parse learns JSON object syntax for struct values and splits only on top-level commas, so a plugin's config survives being passed inline alongside the other key=value pairs.

### DIFF
--- a/internal/cmd/instances.go
+++ b/internal/cmd/instances.go
@@ -26,6 +26,7 @@ import (
 	"github.com/distribution/reference"
 	"github.com/go-json-experiment/json/jsontext"
 	"mvdan.cc/sh/v3/shell"
+
 	"unikraft.com/cloud/sdk/platform"
 	"unikraft.com/cloud/sdk/platform/group"
 	"unikraft.com/cloud/sdk/platform/logs"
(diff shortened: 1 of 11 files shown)


internal/cmd/instances.go
```diff
@@ -452,6 +458,61 @@ func (r *InstanceRom) UnmarshalJSON(data []byte) error {
 	return json.Unmarshal(data, (*romJSON)(r))
 }
 
+// PluginConfig is a plugin's arbitrary JSON configuration, carried as raw JSON
+// text rather than decoded so that any JSON value survives both the --plugin
+// flag and the API round trip.
```
craciunoiuc:
```suggestion
// PluginConfig is a plugin's arbitrary JSON configuration, carried as raw JSON
```

craciunoiuc (commented):
Looks good actually, only one nit. Also hope the parse tests are comprehensive enough

craciunoiuc:
This part of the stack looks good to me now, over to you @jedevc 

Ignore the test fail, that will be fixed separately by me

internal/resource/value/parse.go
```diff
@@ -18,6 +18,76 @@ import (
 	xmaps "unikraft.com/cli/internal/x/maps"
 )
 
+// splitTopLevel splits s on commas, skipping those inside a JSON value. Only a
+// value that opens with {, [ or " immediately after its key's "=" is treated as
+// JSON, so a brace or quote appearing mid-value stays literal and the item
+// separators around it still split.
+func splitTopLevel(s string) ([]string, error) {
```
jedevc: Argh. This is really hacky. But also, I don't have better ideas right now.

internal/cmd/instances_test.go
```diff
@@ -0,0 +1,198 @@
...
+			require.NoError(t, mirror.Mirror(result, &result))
+			require.Len(t, result.Plugins, 1)
+			assert.Equal(t, "logger", result.Plugins[0].Name)
+			assert.Equal(t, "plugins/logger:latest", result.Plugins[0].Rom)
+			assert.Equal(t, tt.wantConfig, result.Plugins[0].Config)
+		})
+	}
+}
+
+func TestInstancePluginConfigPrecision(t *testing.T) {
```
jedevc: Wut? What is this test. (same applies to the rest).

internal/resource/value/parse.go
jedevc: Maybe add a `HACK` to this for now...

The correct way of solving this involves rethinking how we pass options like this.

dragosgheorghioiu: yeah i agree, marked as hack for now

internal/cmd/instances_test.go
dragosgheorghioiu: these were meant to test if the json values are never decoded inside some go struct and they are passed as they are to the api but i agree they are not really useful so i removed them

jedevc (approved):
```
```

PR #484, opened 2026-09-01
refactor(cmd)!: Generate create and edit flags from the schema

jedevc:
The shortcut flags were a second, hand-maintained copy of very create and edit field, and their values reached the patch as strings — `--rom '{"name":"ab","at":"/r,x"}'` errored with `unknown fields: [x]` where the identical `--set roms=…` worked. Flags are now generated from the field schema and land typed, so the field's tag is the only place a flag is declared.

Contributes to [TOOL-686](https://linear.app/unikraft/issue/TOOL-686/updates-to-field-structures-for-vastly-improved-editing) — it doesn't solve that, but it removes the parallel flag declarations and the string round-trip that were in the way.

Flag sets are unchanged across all 18 affected commands; help output is byte-identical apart from ordering, which now follows schema order.

Two behaviour changes worth knowing. `-e FOO` with no `=` is now an error rather than silently setting `FOO=`, since `--env` is the schema's `map[string]string`. And `resources create --type=instance` no longer also sets the instance's VM type to `instance` — `platform.InstanceType` doesn't validate, so that one was going to the API unnoticed.


craciunoiuc (commented):
I will review this later, but for now:

Can you minify the commit message text

Also:
`Two behaviour changes worth knowing. -e FOO with no = is now an error rather than silently setting FOO=`

Should this be marked as a breaking commit?

internal/cmd/root.go
```diff
@@ -366,7 +366,13 @@ func NewParser(cli *UnikraftCLI) (*kong.Kong, error) {
 			},
 		}),
 		kong.NamedMapper("optional", xkong.Optional()),
-	)
+	}
+
+	// The donor parser that builds schema-derived flags needs the same
+	// mappers and groups as this one.
```
craciunoiuc: What is a donor parser? 👀

internal/cmd/volumes.go
```diff
@@ -243,22 +199,22 @@ func (c *VolumesCloneCmd) Run(ctx context.Context, stdio config.Stdio, sandbox *
 }
 
 type Volume struct {
-	Metro LinkName[Metro] `field:"metro,short" create:"set,required"`
-	Name  string          `mirror:"volume.name" field:",short" create:"set"`
+	Metro LinkName[Metro] `field:"metro,short" create:"set,required" flag:"metro" help:"Metro to create in." placeholder:"metro" example:"fra,sfo,nyc"`
```
craciunoiuc: nyc doesn't exist 😔

internal/resource/patch/patch.go
```diff
@@ -120,7 +159,23 @@ func PatchedFields(fields []resource.Field, spec PatchSpec) ([]resource.Field, e
 		}
 	}
 
+	// A typed value that fitted no field at all is a bug in the flag, not an
+	// ambiguity: report it rather than dropping the value silently.
```
craciunoiuc: damn, a vanilla one, haven't seen one of these in a long time 😳

craciunoiuc (commented):
Some small nits my side

I guess I can approve after, it's anyway mostly a refactor

jedevc:
Should be ready again :pray:

craciunoiuc (dismissed):
Just doe it. Thanks!

```
```

PR #485, opened 2026-09-01
fix(integration): Increase OOM test memory

craciunoiuc:
After recent platform and kernel updates, the oom test started failing with an 'unknown error'.

This is because probably some components started using up more memory, making the kernel not boot up at all to report the page fault.

Fix is to slightly increase memory to allow for the crash to happen, but not to much or it will start passing.


jedevc (approved):
```
```

PR #486, opened 2026-09-02
feat: Link user console profiles to analytics

craciunoiuc:
First bumps the sdk.

Links user accounts to the data they were previously sending anonymously.

If a user does not have an uuid saved, it will use the unique machine id instead for tracking.

This needs merging and redeployment to stable first:
Depends-on: https://github.com/unikraft-cloud/console/pull/1150

Closes: TOOL-1131

### DIFF
--- a/cmd/unikraft/main.go
+++ b/cmd/unikraft/main.go
@@ -154,12 +154,12 @@ func run(ctx context.Context, args []string, stdio config.Stdio) (context.Contex
 
 	// Initialize analytics if telemetry is enabled.
 	//
-	// These are anonymous usage analytics, and no personally identifiable
-	// information is collected.  This information is used to help us understand
-	// how the CLI is being used and to improve it over time.  We may collect
-	// information such as which commands are used, how often they are used, and
-	// any errors that occur.  This data is aggregated and analyzed to identify
-	// trends and areas for improvement.
+	// These usage analytics are keyed to the logged-in account when there is
+	// one, and anonymous otherwise.  This information is used to help us
+	// understand how the CLI is being used and to improve it over time.  We may
+	// collect information such as which commands are used, how often they are
+	// used, and any errors that occur.  This data is aggregated and analyzed to
+	// identify trends and areas for improvement.
 	//
 	// Unikraft is committed to user privacy and data protection; visit[0] for
 	// more information.
@@ -173,7 +173,8 @@ func run(ctx context.Context, args []string, stdio config.Stdio) (context.Contex
 	// for our CLI.
 	_, doNotTrack := os.LookupEnv("DO_NOT_TRACK")
 	if !doNotTrack && opts.Telemetry && !isSendAnalytics {
-		if err := telemetry.Init(); err != nil {
+		profile, _ := config.G(ctx).CurrentProfile()
+		if err := telemetry.Init(profile); err != nil {
 			log.G(ctx).
 				Debug().
 				Err(err).
@@ -181,7 +182,7 @@ func run(ctx context.Context, args []string, stdio config.Stdio) (context.Contex
 		} else {
 			log.G(ctx).
 				Debug().
-				Msg("collecting anonymous usage analytics, set `UNIKRAFT_TELEMETRY=false` to disable")
+				Msg("collecting usage analytics, set `UNIKRAFT_TELEMETRY=false` to disable")
 		}
 	}
 
--- a/cmd/unikraft/testdata/TestHelp/instances
+++ b/cmd/unikraft/testdata/TestHelp/instances
@@ -104,7 +104,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
@@ -185,7 +185,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
@@ -272,7 +272,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
@@ -353,7 +353,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
@@ -464,7 +464,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
@@ -661,7 +661,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
@@ -869,7 +869,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
           Set the current profile.
       --[no-]telemetry ($UNIKRAFT_TELEMETRY)
-          Toggle anonymous usage analytics.
+          Toggle usage analytics.
           [default: true]
       --timeout=<duration> ($UNIKRAFT_TIMEOUT)
           Set a deadline for the command (e.g. 30s, 5m, 1h).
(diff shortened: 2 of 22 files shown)


cmd/unikraft/testdata/TestHelp/instances
```diff
@@ -104,7 +104,7 @@ Global flags:
       --profile=<name> ($UNIKRAFT_PROFILE)
```
jedevc, line 104: `task golden-update` 🙏

internal/cmd/login/login.go
```diff
@@ -120,6 +130,8 @@ func (cmd *LoginCmd) Run(ctx context.Context, cfg *config.Config) error {
 	}
 	profile.Token = token
 	profile.Organization = organization
+	profile.UserUUID = userUUID
+	profile.OrganizationUUID = orgUUID
```
jedevc: bug: `--token` logins never set `userUUID` — only the browser path does, and `getOrg` can't help because `GetAuthorizationResponseData` has no `user_uuid` on it. So `findOrCreateProfile` finds an existing profile and this *wipes* a UUID we'd already stored.

internal/telemetry/telemetry.go
```diff
@@ -74,8 +79,17 @@ func Init() error {
 		return fmt.Errorf("no API key set for PostHog; use UNIKRAFT_POSTHOG_API_KEY environment variable")
 	}
 
-	// Generate anonymous distinct ID from machine fingerprint
+	// Use the user UUID when known, otherwise the machine fingerprint.
 	distinctID = generateDistinctID()
```
jedevc: suggestion: Can we set `$process_person_profile: false` when we fall back to the fingerprint? Right now PostHog builds a Person per *machine*, which is the one thing we don't want persisted — and on a shared box or CI runner that Person is several humans.

> [!NOTE]
>
> Maybe we should wipe all the old CLI people?

internal/telemetry/track.go
```diff
@@ -75,6 +76,7 @@ func TrackCommandStart(cmdPath string) {
 		Properties: posthog.NewProperties().
 			Set("command", cmdPath).
 			Set("session_id", sessionID),
+		Groups: groups,
```
jedevc: suggestion: Can we set an explicit `authenticated` bool while we're in here? Otherwise splitting the two populations in a query means guessing whether a `distinct_id` looks like a UUID.

Probably wants a small shared helper rather than four copies of it.

(also at `internal/telemetry/track.go:38`, `internal/telemetry/track.go:101`, `internal/telemetry/track.go:133`)

internal/telemetry/telemetry.go
```diff
@@ -55,11 +58,13 @@ type EventPayload struct {
 	DistinctID string         `json:"distinct_id"`
 	SessionID  string         `json:"session_id"`
```
jedevc: ai nit: `SessionID` is dead — `detach.go` doesn't populate it and `SendEvent` doesn't read it, the real one rides in `Properties`. Every payload just carries a top-level `"session_id": ""`.

`Timestamp` is the same story: no `Track*` func sets it, so posthog-go backfills it in the subprocess.

internal/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,85 @@
...
+				assert.Nil(t, groups)
+				return
+			}
+			assert.Equal(t, tt.org, groups["organization"])
+		})
+	}
+}
+
+func TestEventPayloadCarriesGroups(t *testing.T) {
+	data, err := json.Marshal(EventPayload{
```
jedevc: nit: Is this testing much past `encoding/json`? It builds the `EventPayload` by hand, so it can't catch a field dropped from the literal in `detach.go` — which is where the risk actually is, and where `SessionID` already went missing.

internal/config/profile.go
```diff
@@ -73,6 +73,10 @@ type Profile struct {
 	Token string `json:"token,omitempty" field:",long"`
 	// Organization is the organization associated with the profile.
 	Organization string `json:"organization,omitempty" field:",short"`
+	// UserUUID is the UUID of the user who approved the login, if known.
+	UserUUID string `json:"user_uuid,omitempty" field:",long"`
+	// OrganizationUUID is the UUID of the organization of the token.
+	OrganizationUUID string `json:"organization_uuid,omitempty" field:",long"`
```
jedevc: note: Only a fresh browser login can ever fill these in: `profileFromEnv` has no UUID fields so `UNIKRAFT_PROFILE_ENV`/`UKC_TOKEN` profiles stay anonymous, and existing profiles have nothing to backfill from — `GetAuthorization` is our only refresh hook and it only runs at login.

Not blocking, but worth expecting the identified share to ramp slowly as people re-auth, rather than reading it as growth.

cmd/unikraft/main.go
```diff
@@ -143,12 +143,12 @@ func run(ctx context.Context, args []string, stdio config.Stdio) (context.Contex
...
-	// any errors that occur.  This data is aggregated and analyzed to identify
-	// trends and areas for improvement.
+	// These usage analytics are keyed to the logged-in account when there is
+	// one, and anonymous otherwise.  This information is used to help us
+	// understand how the CLI is being used and to improve it over time.  We may
+	// collect information such as which commands are used, how often they are
+	// used, and any errors that occur.  This data is aggregated and analyzed to
+	// identify trends and areas for improvement.
 	//
 	// Unikraft is committed to user privacy and data protection; visit[0] for
```
jedevc: note: We dropped *No personally identifiable information is collected* from the README, but this still points at the privacy policy — does that page need the same edit before this ships? Especially once the console starts pushing email/name onto the same Persons these events key on.

internal/cmd/login/login.go
```diff
@@ -293,22 +305,22 @@ func (cmd *LoginCmd) getAuth(ctx context.Context, profile *config.Profile) (*con
 	}
 }
 
-func (cmd *LoginCmd) getOrg(ctx context.Context, profile *config.Profile) (string, error) {
+func (cmd *LoginCmd) getOrg(ctx context.Context, profile *config.Profile) (*controlplane.GetAuthorizationResponseData, error) {
```
jedevc: nit: `getOrg` doesn't get an org any more. `getAuthorization`?

internal/telemetry/telemetry_test.go
jedevc, line 1: IMO, these tests don't add much?

jedevc:
Also: follow-up. We should add [`identify`](https://posthog.com/docs/getting-started/identify-users) calls to the console backend when users+orgs are created/updated. This way, we make sure that all users get this info, and isn't reliant on the frontend to do it (e.g. users might deny/block these scripts).

We might need to run some process to batch import all the existing users/orgs to posthog?

internal/telemetry/telemetry.go
craciunoiuc: I guess we discuss the wiping part on discord

internal/telemetry/telemetry_test.go
craciunoiuc: Sure, I guess, I kept them cause they were smol :flushed:

craciunoiuc: Will remove them

internal/config/config.go
```diff
@@ -68,18 +68,37 @@ func (c *Config) Save() error {
...
 	if err := f.Close(); err != nil {
 		return jujuerrors.Annotate(err, "closing config file")
 	}
+
+	if err := os.Chmod(tmp, mode); err != nil {
+		return jujuerrors.Annotate(err, "setting config file permissions")
+	}
+	if err := os.Rename(tmp, c.Path); err != nil {
+		return jujuerrors.Annotate(err, "replacing config file")
+	}
```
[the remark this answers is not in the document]
craciunoiuc: Just make a note for now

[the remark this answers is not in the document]
craciunoiuc: Actually wrong

internal/telemetry/detach.go
```diff
@@ -26,7 +26,19 @@ func spawnDetachedAnalytics(event posthog.Capture) {
...
+		event.Properties.Set("$process_person_profile", false)
+	}
+	if machine {
+		event.Properties.Set("$set", posthog.Properties{"is_machine": true})
+	}
+
+	payload, err := json.Marshal(EventPayload{
+		Event:      event.Event,
+		DistinctID: event.DistinctId,
+		Properties: event.Properties,
```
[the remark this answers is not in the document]
craciunoiuc: Wrong

internal/cmd/login/login.go
jedevc, line 34: Follow-up material maybe... what's the use case of setting `--organization`? I think we should remove this.

internal/telemetry/telemetry.go
```diff
@@ -74,8 +79,25 @@ func Init() error {
...
+			groups = posthog.Groups{"organization": profile.OrganizationUUID}
+		}
+		switch {
+		case profile.UserUUID != "":
+			distinctID = profile.UserUUID
+			anonymous = false
+		case profile.OrganizationUUID != "":
+			distinctID = "machine-" + distinctID
+			anonymous = false
+			machine = true
```
jedevc: Something I've realized. In this scenario, a machine distinct id can belong to multiple groups. Does that cause problems?

jedevc: Random thought. In this scenario, maybe we could set the distinct id to the organization *itself*? Instead of trying to id it to a specific machine.

Then within the context of the console, we identify *this* user as a special "generic" user.

internal/cmd/login/login.go
craciunoiuc: I'm not even sure that field is used

I assume the flag just sets `organization` in the config

I know it's used for displaying the quotas, but I don't think it's used in much else

internal/telemetry/telemetry.go
craciunoiuc: No, it shouldn't cause any problems, at least how I see it, I'll check

On the second comment: I think it's something we clear up with Julian. If you ask me maybe it's better to have a single "Person" for this to cut on the spam and the costs. Especially since these are more like "power users"

internal/telemetry/telemetry.go
```diff
@@ -74,8 +79,25 @@ func Init() error {
...
+			groups = posthog.Groups{"organization": profile.OrganizationUUID}
+		}
+		switch {
+		case profile.UserUUID != "":
+			distinctID = profile.UserUUID
+			anonymous = false
+		case profile.OrganizationUUID != "":
+			distinctID = profile.OrganizationUUID
+			anonymous = false
+			machine = true
```
jedevc: IMO, we should remove this `machine`.

craciunoiuc: Agreed to change to `userless`

internal/telemetry/telemetry.go
```diff
@@ -74,8 +80,26 @@ func Init() error {
 		return fmt.Errorf("no API key set for PostHog; use UNIKRAFT_POSTHOG_API_KEY environment variable")
 	}
 
-	// Generate anonymous distinct ID from machine fingerprint
-	distinctID = generateDistinctID()
+	// Use the user UUID when known, otherwise the machine fingerprint.
+	machineID = generateDistinctID()
```
jedevc: nit: Rename to `generateMachineID`?

internal/telemetry/telemetry.go
```diff
@@ -74,8 +80,26 @@ func Init() error {
...
+	distinctID = machineID
+	groups = nil
+	anonymous = true
+	userless = false
```
jedevc: nit: `userless` defaults to `true` right? `anonymous` implies `userless` right?

jedevc (approved):
```
```

PR #491, opened 2026-09-04
fix: Follow-ups to the sandbox integration

jedevc:
Review follow-ups on #458, one fix per commit.

The user-visible ones: `unikraft instance exec` now exits with the status its command exited with, so a script can check it, and Ctrl-C waits for that command to die of the interrupt instead of giving up on it at once. Underneath, the two output streams are kept apart — folding them onto stdout is the exec command's choice, not something baked in below it — and standard input that fails part way through closes the command's input rather than leaving a `cat` waiting on it forever.

### DIFF
--- a/cmd/unikraft/main.go
+++ b/cmd/unikraft/main.go
@@ -23,6 +23,7 @@ import (
 	"unikraft.com/cli/internal/cmd"
 	"unikraft.com/cli/internal/config"
 	"unikraft.com/cli/internal/logfmt"
+	"unikraft.com/cli/internal/sandbox"
 	"unikraft.com/cli/internal/telemetry"
 	"unikraft.com/x/colors"
 	"unikraft.com/x/log"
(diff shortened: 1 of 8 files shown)


cmd/unikraft/main.go
```diff
@@ -53,6 +54,13 @@ func main() {
 		err = ctx.Err()
 	}
 
+	// a command that ran on an instance and failed isn't an error of ours, so
+	// exit with the status it exited with and print nothing over its output
+	exitCode := 0
+	if exited, ok := errors.AsType[*sandbox.ExitError](err); ok {
+		exitCode, err = exited.ExitCode(), nil
+	}
+
```
dragosgheorghioiu, line 63: this does make more sense than what i had before 
i had the impression it should just swallow the command completely including the exit code

internal/sandbox/cmd.go
```diff
@@ -229,10 +284,7 @@ func (c *Cmd) Wait() error {
 			if err := c.cancel(); err != nil {
 				return err
 			}
-			switch {
-			case c.WaitDelay == 0:
-				return c.interrupted()
-			case c.WaitDelay > 0:
+			if c.WaitDelay > 0 {
```
dragosgheorghioiu, line 287: I also have the same exact fix in the shell pr haha https://github.com/unikraft-cloud/cli/pull/463/changes#diff-4058e480aa874118f6965bc58392b3a953c7cbeab37d8982b33cf8f4f0579640R239

dragosgheorghioiu (approved):
```
```

PR #493, opened 2026-09-07
feat: Support building the CLI for js/wasm

craciunoiuc:
The CLI did not compile for GOOS=js GOARCH=wasm, so the dashboard could not run it in the browser. The build graph reached BuildKit, containerd, Docker, Bubble Tea, a machine fingerprint and a large generated IATA table, none of which the wasm backend accepts.

Commands that need a local daemon, terminal or filesystem now split into `!js` and `js` files. The js side keeps the kong structs, so help output is unchanged, and reports that the command belongs to the native CLI. A `wasm` task builds and compresses the module.

Not yet ready for review

### DIFF
--- a/.github/workflows/check-static.yaml
+++ b/.github/workflows/check-static.yaml
@@ -57,3 +57,11 @@ jobs:
       shell: bash
       run: |
           task go:ci/check
+
+    # The web console runs this CLI as a WebAssembly module.  Build it here so
+    # a new dependency on a local daemon, a terminal or a filesystem is caught
+    # in review rather than at release time.
+    - name: Build for js/wasm
+      shell: bash
+      run: |
+          GOOS=js GOARCH=wasm go build -o /dev/null ./cmd/unikraft
--- a/internal/cmd/build_js.go
+++ b/internal/cmd/build_js.go
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+//go:build js
+
+package cmd
+
+import (
+	"context"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/resource"
+)
+
+// Run reports that image builds need a local container builder.
+func (c *ImageBuildCmd) Run(ctx context.Context, cfg *config.Config, partition *resource.Partition) error {
+	return notAvailable("image build")
+}
+
+// Run reports that image builds need a local container builder.
+func (c *BuildCmd) Run(ctx context.Context, cfg *config.Config, partition *resource.Partition) error {
+	return notAvailable("build")
+}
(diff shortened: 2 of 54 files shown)


craciunoiuc:
Summary of the call:
* Do changes in X first
* Preference order: stub packages in x -> stub packages internal -> stub commands -> partially reimplement for each version
* Remove iata from the CLI
* Limit image-spec only for reading stuff
* Figure out why kraftfile is replaced
* Stub fingerprint in X
* Completely disable TUI things

Bonus:
* Check if https://github.com/rhysd/vim.wasm can be used in the console or any other basic editor
* Do the changes with the CLI in mind not the console(generic, not a single consumer of the feature)

internal/cmd/build_js.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	"unikraft.com/cli/internal/config"
+	"unikraft.com/cli/internal/resource"
+)
+
+// Run reports that image builds need a local container builder.
+func (c *ImageBuildCmd) Run(ctx context.Context, cfg *config.Config, partition *resource.Partition) error {
+	return notAvailable("image build")
```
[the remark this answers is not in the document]
craciunoiuc: I don't see how this would be testable here :thinking:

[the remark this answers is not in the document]
jedevc: Of course the AI wants even more insane and useless tests.

internal/cmd/images_registry_js.go
```diff
@@ -0,0 +1,65 @@
...
+	"unikraft.com/cli/internal/resource/cmd"
+	"unikraft.com/cli/internal/types"
+)
+
+// Image identifies a registry image.  The web console has no local image
+// store, so only the fields which name an image are kept.
+type Image struct {
+	Ref    types.ImageRef[reference.Named] `field:",short"`
+	Digest digest.Digest                   `field:",long"`
+}
```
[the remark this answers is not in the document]
jedevc: Hm, this is a good point. Why can't we get this info?

Is it cause of the containerd dependencies? What's in there...

.github/workflows/check-static.yaml
```diff
@@ -57,3 +57,8 @@ jobs:
       shell: bash
       run: |
           task go:ci/check
+
+    - name: Build for js/wasm
+      shell: bash
+      run: |
+          GOOS=js GOARCH=wasm go build -o /dev/null ./cmd/unikraft
```
jedevc: Maybe should go in `build.yaml`?

internal/cmd/images_registry_js.go
jedevc: What's the blocker for these not being available?

internal/tui/watcher/watcher_native.go
```diff
@@ -0,0 +1,44 @@
...
+
+	tea "charm.land/bubbletea/v2"
+)
+
+func watchOutputPretty(ctx context.Context, interval time.Duration, out io.Writer, render func(io.Writer) error) error {
+	program := tea.NewProgram(
+		watchModel{underlying: out, render: render, interval: interval},
+		tea.WithOutput(out),
+		tea.WithContext(ctx),
+	)
```
jedevc: No bubbletea in wasm?

internal/cmd/build_native.go
jedevc: Instead of calling these files `_native.go`, I would just leave them as is. Only the `_js.go` variants need the suffix.

PR #494, opened 2026-09-08
fix(integration): Remove leaked test binary dir

craciunoiuc:
BuildUnikraft put the 51 MB test binary in a temporary directory that nothing removed. Two test packages call it, so every run left two directories behind in /tmp, which is a tmpfs and holds them in RAM.

CleanupUnikraftBinary deletes the directory, in the manner of CleanupSharedImages. TestMain of each package that builds the binary calls it. The package cmd/unikraft has a new TestMain.

### DIFF
--- a/cmd/unikraft/main_test.go
+++ b/cmd/unikraft/main_test.go
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, Unikraft GmbH and The Unikraft CLI Authors.
+// Licensed under the BSD-3-Clause License (the "License").
+// You may not use this file except in compliance with the License.
+
+package main
+
+import (
+	"os"
+	"testing"
+
+	integ "unikraft.com/cli/internal/integration"
+)
+
+func TestMain(m *testing.M) {
+	code := m.Run()
+	integ.CleanupUnikraftBinary()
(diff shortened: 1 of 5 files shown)


cmd/unikraft/main_test.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"os"
+	"testing"
+
+	integ "unikraft.com/cli/internal/integration"
+)
+
+func TestMain(m *testing.M) {
+	code := m.Run()
+	integ.CleanupUnikraftBinary()
```
jedevc: Hm. Wonder if we should `CleanupSharedImages` here as well? Or maybe even have a general `integration.Cleanup`, that calls `cleanupSharedImages` and `cleanupUnikraftBinary`, so nothing falls through the cracks in the future?

jedevc (changes requested):

cmd/unikraft/main_test.go
craciunoiuc: Yeah, real, load bearing, material, word

I created a smol refactor commit on top so we have a single Cleanup function everywhere. If the subfunctions don't apply, they nop

Should be fine now

jedevc (approved):
Magnificent.

```
```

PR #497, opened 2026-09-09
fix(cmd): Use the stdlib for the command timeout

jedevc:
Drops the hand-rolled timeout context in favour of `context.WithTimeoutCause`. Behaviour is unchanged apart from the message, which now names the deadline that fired — `timed out after 30s` rather than `Get "https://…": timed out` — since a global `--timeout` bounds the whole command, not one request.

This came out of looking at TOOL-1426, which I could not reproduce: `--timeout` *is* enforced on a hung plugin route, over HTTP/1 and HTTP/2, on a connection held open and on a body that stalls mid-response. What I found instead was that `Err()` returned a bare `timed out`, so `errors.Is(err, context.DeadlineExceeded)` was false everywhere, and the returned cancel never cancelled anything — which was quietly load-bearing, since making it real turns `ctx.Err()` into a non-nil error on every successful run that set `--timeout`.

Relates to https://linear.app/unikraft/issue/TOOL-1426/unikraft-api-timeout-not-enforced-on-hung-plugin-route-requests.

### DIFF
--- a/cmd/unikraft/main.go
+++ b/cmd/unikraft/main.go
@@ -48,11 +48,6 @@ func main() {
 	)
 
 	ctx, err = run(ctx, args, stdio)
-	if err == nil {
-		// catch context cancellation errors, and make sure we show them, even if
-		// the command succeeded
-		err = ctx.Err()
-	}
 
 	// a command that ran on an instance and failed isn't an error of ours, so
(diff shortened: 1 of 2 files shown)


craciunoiuc (approved):
All good here. Thanks!

```
```

PR #500, opened 2026-09-11
feat: Write command help text as Markdown

danielvallance:
The api description was written for the terminal, with ALL-CAPS headings and indented tables. Both generators emitted it verbatim, so the docs page showed pseudo-headings and the man page reflowed the aligned columns into ragged text.

It is now Markdown: kingkong renders it for the terminal, the docs pass it through, and man generation demotes the headings for md2man. gendocs gains golden coverage, and one example which documented the wrong key order is corrected.

This x PR should be merged first as this PR depends on this one: https://github.com/unikraft-cloud/x/pull/462

The original approach was to infer the headings/code tabs from the plaintext content as in (https://github.com/unikraft-cloud/cli/pull/496) however this was too fragile so changing the help text to markdown adds structure and is more reliable


danielvallance:
- [ ] Update go.mod/go.sum once x PR is merged

PR #503, opened 2026-09-16
feat: Pass Dockerfiles to unikraft build

danielvallance:
Now that the platform supports kernel-less images, it is not strictly necessary to pass a Kraftfile
to unikraft build - instead it can accept a
Dockerfile.

Passing a Dockerfile will build a image which
contains a rootfs based off the Dockerfile,
and no kernel.

The filename is used to determine which
type of file has been passed. If a directory
is passed, the Kraftfile takes precedence
over the Dockerfile.

When a Dockerfile is passed, the --arch argument
must be set so the CLI knows which architecure
to build for.

Kernel-less images can only be run on metros
with a default kernel installed.

Closes: TOOL-1439


internal/builder/build_test.go
```diff
@@ -324,6 +324,28 @@ CMD ["/dockerfile-cmd"]
 	require.Contains(t, imgs[0].Image.Config.Env, "DOCKERFILE_ENV=from-dockerfile")
 }
 
+func TestBuildDockerfileLabelsIntegration(t *testing.T) {
+	ctx := integrationContext(t)
+	dockerfile := `
+FROM scratch
+LABEL org.unikraft.source=dockerfile org.unikraft.only=dockerfile
```
nderjung: where is `org.unikraft.source` coming from?

IMHO we should use `com.` || `cloud.` for anything specifically cloud-based - so as not to mess with stuff in the OSS/KraftKit.

craciunoiuc:
do let me know when I should check it out :wink:

internal/builder/build_test.go
danielvallance: a similar label was in `TestBuildCmdEnvLabelsIntegration` - however that test and the tests I added do not rely on the meaning of the label at all, so am just going to change them to use com.example.* type labels

danielvallance:
@craciunoiuc stable integration test is failing because of a platform issue fixed in latest version

danielvallance:
@craciunoiuc they pass now!

internal/builder/input.go
```diff
@@ -0,0 +1,106 @@
...
+
+// kraftfileNames are the Kraftfile names that a project directory can contain,
+// in order of precedence. The order is the name order of the directory listing.
+var kraftfileNames = []string{
+	"Kraftfile",
+	"Kraftfile.yaml",
+	"Kraftfile.yml",
+	"kraft.yaml",
+	"kraft.yml",
+}
```
craciunoiuc, line 34: If this isn't already in `x/kraftfile` then it should be probably 🤔

internal/builder/input.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+	}
+}
+
+// isDockerfileName reports whether a file name denotes a Dockerfile, such as
+// "Dockerfile", "app.Dockerfile" or "Dockerfile.dev".
+func isDockerfileName(base string) bool {
+	return base == DefaultDockerfileName ||
+		slices.Contains(strings.Split(base, "."), DefaultDockerfileName)
+}
```
craciunoiuc, line 106: inline

internal/builder/input.go
```diff
@@ -0,0 +1,106 @@
...
+			Str("min", kraftfile.SpecVersionMin).
+			Msg("Kraftfile spec version is older than minimum; parsing is best-effort")
+	} else if semver.Compare(kf.Spec, kraftfile.SpecVersionMax) > 0 {
+		log.G(ctx).Warn().
+			Str("spec", kf.Spec).
+			Str("max", kraftfile.SpecVersionMax).
+			Msg("Kraftfile spec version is newer than maximum; parsing is best-effort")
+	}
+	return KraftfileToBuildOpts(dir, kf)
+}
```
craciunoiuc, line 88: just one use place, so inline

internal/builder/input.go
```diff
@@ -0,0 +1,106 @@
...
+// DockerfileToBuildOpts returns build options that use the Dockerfile at path
+// as the only rootfs source. The directory of the Dockerfile is the context.
+func DockerfileToBuildOpts(path string) BuildOpts {
+	return BuildOpts{
+		Rootfs: FSOpts{
+			Path: path,
+			Type: kraftfile.SourceTypeDockerfile,
+		},
+	}
+}
```
craciunoiuc, line 99: Inline also 😳

internal/cmd/build.go
```diff
@@ -110,6 +97,9 @@ func (c *ImageBuildCmd) Run(ctx context.Context, cfg *config.Config, partition *
 	if err := buildOpts.FilterArch(arches...); err != nil {
 		return err
 	}
+	if buildOpts.Runtime == "" && len(buildOpts.Platform) == 0 {
+		return fmt.Errorf("no target architecture: use --arch when neither declared targets nor a runtime give one")
+	}
```
craciunoiuc, line 102:
```suggestion
	if buildOpts.Runtime == "" && len(buildOpts.Platform) == 0 {
		return fmt.Errorf("no target architecture: use --arch when neither declared targets nor a runtime have one")
	}
```

skills/unikraft-cli/SKILL.md
```diff
@@ -49,6 +52,7 @@ unikraft build [flags] [<path>]
 **Common Flags:**
 
 - `-o, --output <dest>`: Output destination (registry tag or local OCI archive path).
+- `--arch <arch>`: Architectures to build (`x86_64`, `arm64`). Required when neither declared Kraftfile targets nor a runtime give an architecture, such as for Dockerfile-only builds.
```
craciunoiuc, line 55:
```suggestion
- `--arch <arch>`: Architectures to build (`x86_64`, `arm64`).
```

README.md
```diff
@@ -346,6 +346,12 @@ unikraft volumes clone my-volume --set name=my-volume-backup
 unikraft build . --output my-org/my-app:latest
 ```
 
+**Build and publish an image from a Dockerfile, without a Kraftfile:**
+
+```sh
+unikraft build ./Dockerfile --arch x86_64 --output my-org/my-app:latest
+```
+
```
craciunoiuc, line 354: Up to debate if we should have this in the readme

internal/builder/input.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"golang.org/x/mod/semver"
+
+	"unikraft.com/x/kraftfile"
+	"unikraft.com/x/log"
+)
+
+// DefaultDockerfileName is the file that a project directory without a
+// Kraftfile must contain to be built.
+const DefaultDockerfileName = "Dockerfile"
```
craciunoiuc, line 24: Maybe this can be found in some of the libraries we already use? 🤔

craciunoiuc (changes requested):
First set of comments

General note: Go people hate abstraction (for the sake of abstraction)

internal/builder/input.go
danielvallance: do you mind if i keep this as its own function? it is used twice so i think having it as its own function keeps it simpler

internal/builder/input.go
danielvallance: i had a brief look - there is dockerui.DefaultDockerfileName, but that is not currently imported so i don't personally think it's worth dragging in more packages just for "Dockerfile"

internal/builder/input.go
danielvallance: sure - I can add that afterwards

internal/builder/input.go
craciunoiuc: yeah not worth

craciunoiuc (commented):
I think you have like a crazy commit message rule.

You can ask it for under 50 lines for titles and under 70 lines for commit bodies.

Other than that it looks good to me I think? @jedevc you're free to do a pass

internal/builder/input.go
jedevc: It's already there 🤔

internal/builder/input.go
jedevc, line 1: All of this logic is *very* similar to `ParseDirectory`.

Instead of doing all that, can we just try that first, then if we detect an error (probably need to make an exported "no kraftfile found in directory" error type), try and load from a dockerfile?

Then we don't need to import *all* this logic again.

internal/builder/input.go
craciunoiuc: well if you ask me so nicely how can I say no

internal/builder/input.go
craciunoiuc: :thinking:

jedevc (approved):
```
```

PR #505, opened 2026-09-17
feat(shell): Add a prompt if trying to run a sandbox command

dragosgheorghioiu:
image

also includes:
- rename from `--plugin` to `--plugin-name`
- `--plugin-image`


internal/cmd/sandbox.go
```diff
@@ -348,9 +527,10 @@ type CopySandboxInstanceCmd struct {
 	Source      string `arg:"" name:"source" help:"File to copy from, either a local path or <instance>:<path>."`
 	Destination string `arg:"" name:"destination" help:"Where to copy it to, either a local path or <instance>:<path>."`
 
-	Plugin  string `name:"plugin" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
-	Force   bool   `name:"force" help:"Overwrite the local file if it already exists."`
-	Parents bool   `name:"parents" short:"p" help:"Create missing parent directories on the destination path."`
+	Plugin      string `name:"plugin-name" default:"${sandbox_plugin}" help:"Name of the sandbox plugin to use." placeholder:"name"`
+	PluginImage string `name:"plugin-image" default:"${sandbox_plugin_image}" help:"Image of the sandbox plugin to attach when the instance has none." placeholder:"ref"`
```
jedevc: Let's split common options like this to a separate struct?

internal/cmd/sandbox.go
```diff
@@ -125,52 +151,105 @@ func parseEnv(records []string) (map[string]string, error) {
 	return env, nil
 }
 
-func resolveSandboxTarget(ctx context.Context, partition *resource.Partition, target, plugin string) (sandbox.Target, error) {
+func resolveSandboxTarget(ctx context.Context, stdio config.Stdio, partition *resource.Partition, target, plugin, image string) (sandbox.Target, error) {
```
jedevc: Then we can pass the whole pluginname and pluginimage here if they're a separate struct.

internal/tui/selector/single.go
```diff
@@ -52,7 +54,35 @@ func singleSelect[T ~string](question string, defaultValue string, options []T)
...
-	p := tea.NewProgram(&singleSelectModel{
+	selected, err := runSingleSelect(newSingleSelectModel(question, items, cursor))
+	if err != nil {
+		return zero, err
+	}
+	return mapped[items[selected].text], nil
+}
+
+// Choose is the index of the option picked, the cursor starting on the last one; giving up is [ErrNoOptionSelected].
+func Choose(ctx context.Context, in io.Reader, out io.Writer, question string, options ...string) (int, error) {
```
jedevc: Why is `Single` separate from from `Choose`? Aren't they the same?

internal/cmd/sandbox.go
```diff
@@ -7,37 +7,63 @@ package cmd
...
 	"unikraft.com/cli/internal/resource"
+	"unikraft.com/cli/internal/resource/cmd"
 	"unikraft.com/cli/internal/resource/value"
+	"unikraft.com/cli/internal/tui/selector"
 	"unikraft.com/cloud/sdk/plugins/sandbox"
 	xio "unikraft.com/x/io"
 	"unikraft.com/x/shell"
 )
 
 const copyPathSeparator = ":"
```
khatibomar: since you added this const the next const should join it under same brackets

so it becomes

```go
const (
     copyPathSeparator = ":"

    defaultPluginImage = "plugins/sandbox:latest"

	startHint  = "%[1]s it with `unikraft instance %[1]s %[2]s`"
	attachHint = "attach it with `unikraft instance edit %s --plugin name=%s,image=%s` while the instance is stopped"
.....
)
```

internal/cmd/sandbox.go
```diff
@@ -125,52 +151,105 @@ func parseEnv(records []string) (map[string]string, error) {
...
+		if err := missingPlugin(instance, plugin); err != nil {
+			return sandbox.Target{}, fmt.Errorf("after %s the plugin: %w", did, err)
+		}
+	}
+
+	if !instance.State.IsRunning() {
+		return sandbox.Target{}, fmt.Errorf("instance %q is not running (state: %s)", instance.Name, string(instance.State))
+	}
+
+	return sandboxTargetFor(ctx, instance, plugin)
```
jedevc: I would move all of this logic to a function called `attachPlugin`, it returns `error`.

Maybe also move to a separate file (maybe even in a separate package), to avoid this file containing all the commands and also all the helpers for it.

internal/tui/selector/single.go
dragosgheorghioiu: they are choose was able to print on stderr too but that wasn't really needed

internal/cmd/sandbox.go
dragosgheorghioiu: i moved them in a separate file

PR #508, opened 2026-09-21
refactor(sandbox): Export the sandbox command package

jesi-rgb:
The console needs to run commands on an instance through the sandbox plugin. It had copied `internal/sandbox`, because `internal/` blocks imports from other modules (see unikraft-cloud/console#1145, review by @jedevc).

This moves the package to `pkg/sandbox` so both binaries share it, and folds in what the console's copy had improved:

- a command that ignores SIGINT is killed once its `WaitDelay` is out, and its record dropped
- an output writer that panics ends only that command, not the process
- a node that cannot be reached is reported without its API URL
- log polls ask for 256 KiB, so a base64 WebSocket frame stays small

The shell transport moves in as `sandbox.Transport`, replacing `internal/cmd/shell_sandbox.go`, so the console does not copy that either.

`task lint` and `task test` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/sandbox/cmd.go
```diff
@@ -335,11 +366,15 @@ func (c *Cmd) Run() error {
 	return c.Wait()
 }
 
-// Forget drops the instance's record of the command
+// Forget drops the instance's record of the command. The plugin keeps one for
+// every command it has run, so a caller that has read all it wants of a
+// finished command says so. It is not an error worth reporting: the record is
+// the instance's, and it outliving the command harms nothing here.
```
dragosgheorghioiu: i would keep this comment shorter

pkg/sandbox/logs.go
```diff
@@ -12,20 +12,24 @@ import (
...
+	// logChunkSize is the most bytes one poll asks for.  Base64 and the
+	// WebSocket frame make each chunk larger again.
+	logChunkSize    = 256 << 10
 	logFetchTimeout = 10 * time.Second
-	logDrainBudget  = 10 * time.Second
+
+	// logDrainBudget is how long the last read of a command's output rides out
+	// a gateway that is not answering yet.
+	logDrainBudget = 10 * time.Second
 )
```
dragosgheorghioiu, line 25: i would keep the units here and remove the comments

pkg/sandbox/stdin.go
```diff
@@ -15,13 +15,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/docker/go-units"
-
 	plugin "unikraft.com/cloud/plugins/sandbox"
 )
 
 const (
```
dragosgheorghioiu, line 23: keep the units

dragosgheorghioiu (commented):
Looks good, just some really small nits.

But I would just move all of this out of the cli. Instead of doing the imports directly from the cli.

@jedevc maybe the sdk is better for that than x wdyt? - https://github.com/unikraft-cloud/go-sdk/pull/269

jedevc:
It would be nice to put in the go-sdk. But the current problem with that, is there's nothing else in the go-sdk that interacts with plugins at all - so it's kind of a dummy package... maybe that's okay for now?

I would put it in `platform/sandbox`, but otherwise... seems good.

pkg/sandbox/cmd.go
jesi-rgb: Done.

pkg/sandbox/logs.go
jesi-rgb: Done.

pkg/sandbox/stdin.go
jesi-rgb: Done.

dragosgheorghioiu:
> It would be nice to put in the go-sdk. But the current problem with that, is there's nothing else in the go-sdk that interacts with plugins at all - so it's kind of a dummy package... maybe that's okay for now?
> 
> I would put it in `platform/sandbox`, but otherwise... seems good.

what about `platform/plugins/sandbox` or just `plugins/sandbox` i can see at least some netshield integration in there too

jedevc:
I like the idea of `plugins/sandbox`, sure.

dragosgheorghioiu (approved):
```
```

dragosgheorghioiu:
will merge this and have the go-sdk pr after this one

PR #511, opened 2026-09-21
fix: Avoid timeouts when pushing to a registry

dragosgheorghioiu:
`unikraft image build --output <ref>` fails when pushing a large rootfs: a ~2 GB blob PUT dies 30s in. 

This adds a dedicated HTTP client for container-registry traffic with no response-header timeout, so a registry may take as long as it needs to commit a large blob.


internal/httpclient/httpclient.go
```diff
@@ -34,3 +41,35 @@ var InsecureHTTPClient = sdkhttpclient.NewHTTPClient(
 	sdkhttpclient.WithUserAgent(version.UserAgent()),
 	sdkhttpclient.WithInsecure(),
 )
+
+// RegistryHTTPClient is the HTTP client used for container registry traffic.
+// Unlike DefaultHTTPClient it sets no response header timeout, because a
+// registry committing a multi-gigabyte blob can take minutes to answer the
+// closing PUT. Pushes remain bounded by the request context.
```
craciunoiuc:
```suggestion
// RegistryHTTPClient is the HTTP client used for container registry traffic.
// Unlike DefaultHTTPClient it sets no response header timeout.
// Pushes remain bounded by the request context.
```

internal/httpclient/httpclient.go
```diff
@@ -34,3 +41,35 @@ var InsecureHTTPClient = sdkhttpclient.NewHTTPClient(
...
+
+func newRegistryTransport() *http.Transport {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DialContext = (&net.Dialer{
+		Timeout:   10 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}).DialContext
+	transport.MaxIdleConns = 500
+	transport.MaxIdleConnsPerHost = 100
+	transport.ResponseHeaderTimeout = 0
```
craciunoiuc: Were these values taken from somewhere?

internal/httpclient/httpclient.go
```diff
@@ -6,12 +6,19 @@
...
+	"os"
+	"time"
 
 	sdkhttpclient "unikraft.com/cloud/sdk/pkg/httpclient"
 	"unikraft.com/x/version"
 )
 
+// RegistryResponseHeaderTimeoutEnv sets a response header timeout on the
+// registry clients, which have none by default.
+const RegistryResponseHeaderTimeoutEnv = "UNIKRAFT_REGISTRY_RESPONSE_HEADER_TIMEOUT"
```
jedevc: :eyes:

I don't think this needs configurating.

internal/httpclient/httpclient_test.go
```diff
@@ -0,0 +1,81 @@
...
+			assert.Equal(t, test.expected, newRegistryTransport().ResponseHeaderTimeout)
+		})
+	}
+}
+
+// TestRegistryTransportReachesClient pins the SDK behaviour the registry
+// clients rely on: a transport passed with WithTransport is the one that
+// serves requests, and is not shadowed by the User-Agent wrapper the SDK puts
+// in front of it. Were it ignored, the SDK's own 30s default would apply and
+// the slow reply below would be tolerated.
```
craciunoiuc:
```suggestion
// TestRegistryTransportReachesClient pins the SDK behaviour the registry
// clients rely on: a transport passed with WithTransport is the one that
// serves requests.
```

jedevc:
:thinking: what kind of ref?

This does feel odd to me, `ResponseHeaderTimeout` is *after* the request is fully written. It takes 30s to write *after* all the data is there?

See https://github.com/containerd/containerd/blob/574a60700ee870807a052e2a0d0c7eed08b7ce8f/core/remotes/docker/registry.go#L254-L269, for the containerd default transport... which defaults to 30s.

craciunoiuc (commented):
Just one more important comment

dragosgheorghioiu:
> 🤔 what kind of ref?
> 
> This does feel odd to me, `ResponseHeaderTimeout` is _after_ the request is fully written. It takes 30s to write _after_ all the data is there?
> 
> See https://github.com/containerd/containerd/blob/574a60700ee870807a052e2a0d0c7eed08b7ce8f/core/remotes/docker/registry.go#L254-L269, for the containerd default transport... which defaults to 30s.

For a 2.5G image pushed to harbor:
```
FROM --platform=linux/x86_64 alpine:3.20 AS build

RUN dd if=/dev/urandom of=/big.bin bs=1M count=2560

FROM scratch

COPY --from=build /big.bin /big.bin
COPY --from=build /bin/busybox /bin/sh
``` 
You get this `http2: timeout awaiting response headers` so it seems it does take more than 30 seconds to respond after the data is there. It basically fails when committing the layers.
It may be the case that is just harbor being slow. Pushing to local registries doesn't have this problem.

